PR #5, opened 2019-02-06
Switched reference, added info about referent arrays.

CloneTrooper1019:
I switched out the reference to Rbx2Source with Roblox-File-Format, added a section about referent arrays, and tweaked a few things.


LPGhatguy:
Awesome, thanks!

PR #15, opened 2019-03-22
Int64 and Float64

Kampfkarren:
Side note, there's a LOT of repetitive code in this codebase. A proc macro would do this wonders.


LPGhatguy:
A proc macro is _probably_ too much, but I'm definitely interested in expanding the use of traits and regular macros!

I've been trying to be pretty cautious since I don't want to introduce fragile abstractions that we'll have to knock down as our understanding of the formats progress!

LPGhatguy:
Looks good! I might go through and update the macro to use something like paste instead of a module, or maybe I'll refactor all of the types to be contained in modules and stop importing them.

Kampfkarren:
I wasn't entirely sure how to make the macro, just that it needed one, that's why the syntax for it is so weird.

Issue #18, opened 2019-03-23
Smoke name mapping


Kampfkarren:
Opacity -> opacity_xml
Size -> size_xml
RiseVelocity -> riseVelocity_xml

Issue #16, opened 2019-03-22
Drive serialization from reflection database


LPGhatguy:
I'm pretty confident that the reflection database generated in `rbx_reflection` is pretty solid, and in the cases it isn't we can apply temporary fixups.

In order to make rbx-dom more compatible with Roblox, we should be able to refactor both `rbx_xml` and `rbx_binary`  to use the reflection database instead of blindly deserializing and serializing every property.

I think I'll start with `rbx_binary` since it needs attention anyways.

Issue #14, opened 2019-03-22
rbx_xml name mapping for Fire


LPGhatguy:
For `Fire`, `Size` serializes as `size_xml`, and another property I can't remember has a weird name too. Need to add an XML<->Canonical name mapping for `Fire`.

Kampfkarren:
That means this needs to be changed to be class specific instead of broad property names:

https://github.com/LPGhatguy/rbx-dom/blob/9700f1417e2de7f119fc1701138e6a370e84f744/rbx_xml/src/reflection.rs#L13

LPGhatguy:
I'll probably move it into the `rbx_reflection` crate eventually, too!

LPGhatguy:
Foundation for this is in now that #28 is around. I'm currently experimenting with porting `rbx_xml` over to use `rbx_reflection` under the hood, which will be known as reflection-guided serialization.

PR #28, opened 2019-04-06
Start tracking 'canonical properties'

LPGhatguy:
This PR is a robust solution for #18, #14, and is a step towards #16.

This change introduces the idea of a *canonical property* more concretely. This idea helps us map from serialization-specific names onto common names that projects can agree on for referring to the same idea.

The concept for property canonicalization came out of Rojo, and seeks to handle these cases:

1. Some properties have deprecated aliases, or aliases used only for serialization, like `Fire.Size`
2. Some properties are scripted through special getters/setters, like `CollectionService` tags
3. Some properties are not scriptable, but specified in the disk formats, like `Lighting.Technology`

A new file was added to generate_reflection called `canonical-properties.toml` that specifies these relationships. Previously, these ideas were present in rbx_xml's ad hoc reflection database, but they're needed for rbx_binary as well.

There are a couple of assumptions that this approach makes:

1. Most properties in the API dump are scriptable and have the same name and type when serialized
2. Most properties that don't follow (1) are just aliases of other properties

While I'm sure about (1), there are non-trivial cases where (2) doesn't hold up. The most notable example is part color, which used to be `BrickColor`, but is now `Color3uint8`, which requires a name **and type** conversion to `Color3` before it can be reflected to Lua.


LPGhatguy:
It's mostly ready, I think!

PR #29, opened 2019-04-15
BrickColor support

Kampfkarren:
BrickColors are serialized as normal numbers, so this is the best way to do it I feel.

Makes use of TryFrom<T>, stabilized in Rust 1.34.0.

### DIFF
--- a/rbx_dom_weak/src/brick_color.rs
+++ b/rbx_dom_weak/src/brick_color.rs
@@ -10,7 +10,7 @@ macro_rules! make_brick_color {
 
 		impl BrickColor {
 			pub fn from_palette(value: u8) -> Option<BrickColor> {
-				match value.into() {
+				match value {
 					$(
 						$value => Some(BrickColor::$enum),
 					)+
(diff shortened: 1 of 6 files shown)


rbx_dom_weak/src/brick_color.rs
```diff
@@ -0,0 +1,176 @@
...
+			fn try_from(value: u8) -> Result<Self, Self::Error> {
+				match value {
+					$(
+						$value => Ok(BrickColor::$enum),
+					)+
+
+					_ => Err(()),
+				}
+			}
+		}
```
LPGhatguy: I'm not super comfortable bumping the minimum Rust version for `TryFrom`, can we just use an inherent impl with a descriptive name?

Kampfkarren: Done.

rbx_dom_weak/src/brick_color.rs
```diff
@@ -0,0 +1,163 @@
...
+macro_rules! make_brick_color {
+	({$([$enum:ident, $name:tt, $value:tt],)+}) => {
+		#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+		pub enum BrickColor {
+			$($enum = $value,)+
+		}
+
+		impl BrickColor {
+			pub fn from_palette(value: u8) -> Option<BrickColor> {
+				match value.into() {
```
LPGhatguy:
```suggestion
				match value {
```

LPGhatguy:
Thanks a bunch!

PR #30, opened 2019-04-16
BinaryType for bool and str

Kampfkarren:


rbx_binary/src/core.rs
```diff
@@ -1,3 +1,42 @@
...
+    fn write_array<W: Write>(
+        output: &mut W,
+        values: &[&T]
+    ) -> io::Result<()> {
+        for value in values {
+            Self::write_binary(output, value)?;
+        }
+
+        Ok(())
+    }
```
LPGhatguy: Can we collapse this trait to just the `read_array` and `write_array` methods? I'd rather that we have to consider the loop for every type instead of having to deal with the boilerplate of writing both versions of the method in cases where they don't overlap.

We can provide the `read_single` and `write_single` methods with provided implementations that use the `read_array` and `write_array` forms instead and specialize those if we need to!

Additionally, when we write `read_array` in each case, we should use [`Vec::with_capacity`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity), which will save us a ton of allocations since we know our size in advance!

I'm also wondering if the write interface should accept an iterator instead of a slice, which lets us avoid intermediate allocations when we're using derived values!

rbx_binary/src/deserializer.rs
```diff
@@ -300,9 +301,9 @@ fn decode_prop_chunk<R: Read>(
 
     match data_type {
         0x01 => {
-            let values = decode_string_array(&mut source, instance_type.referents.len())?;
+            let values = StringType::read_array(&mut source, instance_type.referents.len())?;
 
-            for (index, value) in values.iter().enumerate() {
+            for (index, value) in values.into_iter().enumerate() {
```
LPGhatguy: This is a good change that I've been picking up on more code of mine recently!

rbx_binary/src/types.rs
```diff
@@ -4,54 +4,39 @@ use std::{
...
 
-pub fn encode_bool_array<W: Write>(output: &mut W, values: &[bool]) -> io::Result<()> {
-    for &value in values {
-        encode_bool(output, value)?;
+    fn write_binary<W: Write>(output: &mut W, value: &bool) -> io::Result<()> {
+        if *value {
+            output.write_u8(1)
+        } else {
+            output.write_u8(0)
+        }
```
LPGhatguy: Apparently you can just cast `bool` into a `u8` to get this effect!

```rust
fn main() {
    println!("{}", true as u8); // 1
}
```

LPGhatguy:
This is a good step forward, I'm going to merge and tinker with it in `master`.

Thanks!

PR #32, opened 2019-04-20
Reflection guided serialization for rbx_xml

### LINKED ISSUE #14: rbx_xml name mapping for Fire

### LINKED ISSUE #16: Drive serialization from reflection database

### LINKED ISSUE #18: Smoke name mapping

LPGhatguy:
Fixes #18, fixes #16, fixes #14.

Converts rbx_xml to depend on rbx_reflection, using its database to drive serialization. This is probably a breaking change, so it might be a good idea to increment the public API before/after this is merged and bundle some more breaking changes in.

Follow up changes:
- (#35) Make rbx_reflection API less public (getters instead of `pub` fields)
    - Hide `Cow` types that are leaking right now
    - Change names to include words like `Reflection` and `Descriptor`
- (#34) Serializer/deserializer options like turning off reflection guidance (raw mode?)
- (#36) Nail down `RbxValue::try_convert/try_convert_ref` API, document it, since it changes rbx_dom_weak
- (#37) Expand test suite where reflection-guided serialization might fall down


LPGhatguy:
Going to push off TODOs into future changes and push to merge this!

PR #46, opened 2019-05-08
BrickColor to Color3

### LINKED ISSUE #45: Conversions from BrickColor to Color3
This should be pretty straightforward in the context of `RbxValue::try_convert_ref`. I'm not sure if we want to bother trying to go the other direction since it's lossy.

Kampfkarren:
Closes #45. The `brick_color.rs` file is getting a bit messy IMO, is it worth setting up a toml file and a proc macro to clean it up?


rbx_dom_weak/src/brick_color.rs
```diff
@@ -1,8 +1,9 @@
+#![allow(clippy::unreadable_literal)]
```
Kampfkarren: Suggests 0.123456 be turned into 0.123_456, which in our case is unhelpful.

LPGhatguy:
Can we define the colors with numbers from 0-255 instead of floats from 0-1? The conversion back should be easy, and all of the brick colors should have really nice and round numbers.

I don't think there's any reason to use TOML + a proc macro or build script, we can already make a DSL with a regular macro and the extra complexity and code to avoid writing code would be silly! :)

Kampfkarren:
@LPGhatguy Did the int thing a few days ago.

rbx_dom_weak/src/brick_color.rs
```diff
@@ -18,6 +18,26 @@ macro_rules! make_brick_color {
...
+				match self {
+					$(
+						BrickColor::$enum => [
+							($color3_r as f32) / 255.0,
+							($color3_g as f32) / 255.0,
+							($color3_b as f32) / 255.0,
+						],
+					)+
+				}
+			}
```
LPGhatguy: Can we leave these APIs `pub(crate)` for now? I'm not sure on naming, like which one should be default, and whether it should be `as_color3` and `as_color3uint8` or whatever.

rbx_dom_weak/src/value.rs
```diff
@@ -187,6 +187,8 @@ impl RbxValue {
 
             (RbxValue::String { value }, RbxValueType::Content) => Converted(RbxValue::Content { value: value.clone() }),
 
+            (RbxValue::BrickColor { value }, RbxValueType::Color3) => Converted(RbxValue::Color3 { value: value.as_rgb_f32() }),
```
LPGhatguy: How about a conversion from `BrickColor` to `Color3uint8`? :D

LPGhatguy:
Fixed the merge conflict I caused by fixing the whitespace from tabs to spaces, LGTM!

PR #39, opened 2019-04-25
Add support for INF and NAN

Kampfkarren:
Closes #33

PR #62, opened 2019-06-03
INF and NAN support everywhere

Kampfkarren:
Closes #39 
I believe I got everything.

### DIFF
--- a/rbx_xml/src/serializer_core.rs
+++ b/rbx_xml/src/serializer_core.rs
@@ -63,11 +63,36 @@ impl<W: Write> XmlEventWriter<W> {
         Ok(())
     }
 
+    /// Writes a `f32` value, but using Roblox's INF, -INF, and NAN values
+    #[inline]
+    pub fn write_characters_f32(&mut self, value: f32) -> Result<(), NewEncodeError> {
+        self.write_characters_f64(value as f64)
+    }
+
+    /// Writes a `f64` value, but using Roblox's INF, -INF, and NAN values
+    pub fn write_characters_f64(&mut self, value: f64) -> Result<(), NewEncodeError> {
+        if value == std::f64::INFINITY {
+            self.write_characters("INF")
+        } else if value == std::f64::NEG_INFINITY {
+            self.write_characters("-INF")
+        } else if value.is_nan() {
+            self.write_characters("NAN")
+        } else {
+            self.write_characters(value)
+        }
+    }
+
     /// The same as `write_characters`, but wraps the characters in a tag with
     /// the given name and no attributes.
     pub fn write_tag_characters<T: std::fmt::Display>(&mut self, tag: &str, value: T) -> Result<(), NewEncodeError> {
         self.write(XmlWriteEvent::start_element(tag))?;
         self.write_characters(value)?;
+        self.write(XmlWriteEvent::end_element())
+    }
+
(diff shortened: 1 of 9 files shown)


rbx_xml/src/serializer_core.rs
```diff
@@ -63,6 +63,25 @@ impl<W: Write> XmlEventWriter<W> {
         Ok(())
     }
 
+    /// Writes a `f32` value, but using Roblox's INF, -INF, and NAN values
+    #[inline]
+    pub fn write_characters_f32(&mut self, value: f32) -> Result<(), NewEncodeError> {
+        self.write_characters_f64(value as f64)
+    }
```
LPGhatguy: I'm kind of nervous about casting NaN between `f32` and `f64`. Can we duplicate the logic using the associated consts for `f32`?

rbx_xml/src/serializer_core.rs
```diff
@@ -63,6 +63,25 @@ impl<W: Write> XmlEventWriter<W> {
...
+    pub fn write_characters_f64(&mut self, value: f64) -> Result<(), NewEncodeError> {
+        if value == std::f64::INFINITY {
+            self.write_characters("INF")
+        } else if value == std::f64::NEG_INFINITY {
+            self.write_characters("-INF")
+        } else if value.is_nan() {
+            self.write_characters("NAN")
+        } else {
+            self.write_characters(value)
+        }
```
LPGhatguy: This might be cleaner as a `match` statement. For the NaN case, we might be able to use a match guard like `x if x.is_nan() => { ... }`

rbx_xml/test-files/inf-and-nan.rbxmx
```diff
@@ -7,25 +7,125 @@
 			<string name="Name">Folder</string>
 			<BinaryString name="Tags"></BinaryString>
 		</Properties>
-		<Item class="NumberValue" referent="RBX3B3D9D3DB43D4E6793B190B081E0A886">
+		<Item class="Folder" referent="RBX84B85A2C500048E79846CD63A6127BA3">
```
LPGhatguy: Can we avoid introducing a grouping for just the NumberValue instances and keep everything flat?

LPGhatguy:
Travis CI build failure is my fault, I'll get it fixed in master.

rbx_xml/src/serializer_core.rs
Kampfkarren: Casting f32 to f64 is lossless, so nothing should go wrong.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8775b4db9e81f01f6315154bd72f6e17

rbx_xml/src/serializer_core.rs
Kampfkarren: Didn't know that was a feature :eyes:

Kampfkarren:
```
warning: floating-point types cannot be used in patterns
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #41620 <https://github.com/rust-lang/rust/issues/41620>
```

Kampfkarren:
@LPGhatguy Flattened the rbxmx file and responded to your other comments.

LPGhatguy:
Build failure doesn't look like your fault, I'll fix it post-merge.

PR #63, opened 2019-06-05
Implement SharedString for rbx_dom_weak and rbx_xml

### LINKED ISSUE #60: SharedString in rbx_dom_weak
We should be able to create a shared string dictionary per tree. When moving instances between trees, it should be possible to add entries from the shared string table.

I'm not super sure how the entries should be cleaned up, especially since it's possible for consumers of rbx_dom_weak to mutate properties arbitrarily. Maybe that'll be left up to consumers? Maybe there could be some ref-counting nonsense?

### LINKED ISSUE #61: SharedString in rbx_xml
Depends on #60.

LPGhatguy:
Closes #60, closes #61.

This is a pretty early implementation of SharedString. It implements SharedString with a process-wide cache akin to `boost::flyweight`, using MD5 as a hash key like how it seems Roblox does.

I tried another approach in the `shared-string-xml`, where I planned to transform all `SharedString` values into `BinaryString`, but I think it was too lossy and that this approach is better.

Things to do before merge:
* [x] Sweep through and Improve code-level documentation
* [x] Improve tests for `shared_string` module, since it's pretty sketchy right now
* [x] Figure out story for `Serialize`/`Deserialize` impls for SharedString
* [x] Changelog entry


rbx_dom_weak/src/shared_string.rs
```diff
@@ -0,0 +1,208 @@
...
+        panic!("ahh");
+    }
+}
+
+impl Drop for SharedString {
+    fn drop(&mut self) {
+        // If the reference we're about to drop is the very last reference to
+        // the buffer, we'll be able to unwrap it and remove it from the
+        // SharedString cache.
+        match Arc::try_unwrap(self.data.take().unwrap()) {
```
Kampfkarren: Use `if let Ok(_) = ...`. You should use clippy!

rbx_xml/src/deserializer.rs
```diff
@@ -315,6 +324,72 @@ fn deserialize_metadata<R: Read>(reader: &mut XmlEventReader<R>, state: &mut Par
...
+            XmlReadEvent::StartElement { name, .. } => {
+                if name.local_name == "SharedString" {
+                    deserialize_shared_string(reader, state)?;
+                } else {
+                    let event = reader.expect_next().unwrap();
+                    return Err(reader.error(DecodeErrorKind::UnexpectedXmlEvent(event)));
+                }
+            }
+            XmlReadEvent::EndElement { name } => {
+                if name.local_name == "SharedStrings" {
```
Kampfkarren: Typo?

rbx_xml/src/deserializer.rs
```diff
@@ -315,6 +324,72 @@ fn deserialize_metadata<R: Read>(reader: &mut XmlEventReader<R>, state: &mut Par
...
+                }
+            }
+            _ => {
+                let event = reader.expect_next().unwrap();
+                return Err(reader.error(DecodeErrorKind::UnexpectedXmlEvent(event)));
+            }
+        }
+    }
+
+    reader.expect_end_with_name("SharedStrings")?;
```
Kampfkarren: Here too?

rbx_xml/src/deserializer.rs
LPGhatguy: Actually not a typo, that's the name of the `SharedString` dictionary in the file, `SharedStrings`!

PR #65, opened 2019-07-11
Fix BrickColor

LPGhatguy:
BrickColor is pretty much totally broken.

This tries to... fix it.


LPGhatguy:
I'm compiling locally with 1.31.1 and the test suite is passing. I wonder what difference between 1.31.1 on my machine and 1.31.0 on the CI machine is that's causing this!

LPGhatguy:
It was a Linux-only dependency (getrandom), and it supports 1.32 and newer. Bumping the minimum compiler from 1.31.1 to 1.32.0 now.

rbx_dom_weak/src/brick_color.rs
```diff
@@ -344,6 +367,13 @@ make_brick_color!({
 });
 
 impl BrickColor {
+    /// Convert from a BrickColor palette number to a BrickColor.
+    ///
+    /// This function should not be used for new code, it only exists because of
+    /// a misunderstanding of the range of valid BrickColor values.
+    ///
+    /// This function operates like `BrickColor.palette` within Roblox.
+    #[deprecated]
```
Kampfkarren:
```suggestion
    #[deprecated(
        since = "VERSION",
        note = "Please use BrickColor::from_number instead."
    )]
```

Kampfkarren: My note doesn't mention that they aren't equivalent, but documentation should.

LPGhatguy: I guess the next release should be 1.9.0

PR #70, opened 2019-08-23
Refactor reflection class database into functions

jeparlefrancais:
By refactoring the RbxClassDescriptor in different functions, there won't be any compile-time stack overflow when building with `opt-level = 0` (the default level).

This will also remove the need to add `opt-level = 1` to every project that depends rbx_reflection.

Yay!


generate_reflection/src/emitter_rust.rs
```diff
@@ -56,10 +56,35 @@ pub fn emit_version<W: Write>(output: &mut W, database: &ReflectionDatabase) ->
...
+        let function_name_token = get_generated_function_name(class_name);
+        let class_name_literal = class_name.as_rust();
+
+        quote!(map.insert(#class_name_literal, #function_name_token()))
+    });
+
+    let len_literal = Literal::usize_unsuffixed(classes.len());
 
     quote! {
+        #![allow(non_snake_case)]
```
LPGhatguy: We already depend on the "heck" crate in generate_reflection, so we might as well use it and generate good function names that Rust likes!

jeparlefrancais: Cool! I did not know something like this was is in there!

LPGhatguy:
Pulled down the class database diff in my editor now that I'm back from RustConf, since the GitHub UI can't deal with such a large file...

There's one last bit, we need to sort both the list of generated functions as well as the body of the `generate_classes` method. Otherwise, we'll get HashMap-driven nondeterminism and get lots of nasty big diffs.

jeparlefrancais:
Yep I thought about it but I supposed it would hurt the performance so I let it go. But I think it's a good idea, because anyway the classes are generated only once.

Will work on it as soon as I get home (and that I get internet)!

LPGhatguy:
My Roblox Studio installation is broken on my machine. It moved to Program Files (which means we don't have write access to `BuiltInPlugins` at regular security), and also my uninstaller fails to find that new location and so I can't fix it right now!

The generator should work now, if I were able to run it!

PR #71, opened 2019-08-24
Fix casing issues with EncodedValue

Kampfkarren:


Kampfkarren:
Ok, that fixed NumberSequence, but ColorSequence has a *separate* issue with it that I'm going to fix.

PR #76, opened 2019-10-09
Fixes #67

### LINKED ISSUE #67: rbx_xml empty BinaryString values should not include CDATA
Right now they serialize like this:

```xml
<BinaryString name="Tags"><![CDATA[]]></BinaryString>
```

but they should serialize like this instead:

```xml
<BinaryString name="Tags"></BinaryString>
```

This is safe because it lines up with how Roblox serializes these values.

Kampfkarren:
Fixes #67.

### DIFF
--- a/rbx_xml/src/types/binary_string.rs
+++ b/rbx_xml/src/types/binary_string.rs
@@ -20,7 +20,9 @@ impl XmlType<[u8]> for BinaryStringType {
         value: &[u8],
     ) -> Result<(), EncodeError> {
         writer.write(XmlWriteEvent::start_element(Self::XML_TAG_NAME).attr("name", name))?;
-        writer.write(XmlWriteEvent::characters(&base64::encode(value)))?;
+        if value.len() > 0 {
+            writer.write(XmlWriteEvent::cdata(&base64::encode(value)))?;
+        }
         writer.end_element()?;
 
         Ok(())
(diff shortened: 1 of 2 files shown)


LPGhatguy:
Will this still produce CDATA for non-empty binary strings?

LPGhatguy:
The round trip tests just make sure that `x == deserialize(serialize(x))`. Since I bet we'll treat them the same when _deserializing_, there isn't really any verification that the serialization stays consistent...

I opened #77 to track that fix in the future.

Kampfkarren:
Fair enough. I'll write a test for the binary thing later today.

Kampfkarren:
Oh shoot, I completely misread the actual issue this was fixing 😅 
I'll fix it up in a minute.

rbx_xml/src/types/binary_string.rs
```diff
@@ -20,7 +20,9 @@ impl XmlType<[u8]> for BinaryStringType {
         value: &[u8],
     ) -> Result<(), EncodeError> {
         writer.write(XmlWriteEvent::start_element(Self::XML_TAG_NAME).attr("name", name))?;
-        writer.write(XmlWriteEvent::cdata(&base64::encode(value)))?;
+        if value.len() > 0 {
```
LPGhatguy:
```suggestion
        if !value.is_empty() {
```

I think there's a Clippy lint that talks about this, but alas, they're not as warnings on this repo yet, heh.

LPGhatguy:
Can you add an entry to the changelog for `rbx_xml` and link to this PR?

LPGhatguy:
Travis build hung, I'll restart it...

PR #79, opened 2019-11-05
Add a workflow for GitHub Actions

LPGhatguy:
Part of ongoing work to start stripping away Travis-CI in favor of GitHub Actions!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,37 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - master
+
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+    - master
+
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
MisterUncloaked: should we be running CI on more operating systems or is this unlikely to fail on Mac & Windows?

LPGhatguy: There shouldn't be any platform-specific code that would fail in this package. It doesn't interact with terminals, filesystems, networks, etc, so we should be okay.

If that turns out not to be true in the future, we can revisit this and expand what platforms we check on!

PR #81, opened 2019-11-06
Rewrite rbx_binary, take two

### LINKED ISSUE #11: Generated rbxl files have duplicate services when loaded
This manifests itself mostly in Rojo -- services aren't tagged with the right piece of metadata in the binary format I don't think, which causes this fun:

image 11

### LINKED ISSUE #77: Add snapshot tests with insta
I think the tests we've got for the serialization libraries are insufficient!

Rojo adopted the 'insta' crate for this and it's worked very well. I think it would help us be more confident with changes like #76.

LPGhatguy:
This PR rewrites the binary serializer and deserializer from scratch. I'm hoping to improve its performance, correctness, and API in mostly one sweep.

We should be able to land these changes without breaking the API surface (much) and then come back in a second pass with an improved API similar to what we did with rbx_xml.

## Notable Changes
- Much more grokkable serializer. The serializer's state is now packed into a `BinarySerializer` struct that incrementally discovers information and writes everything all at once.
- Brand new deserializer. Changes in philosophy are similar to that of the serializer.
- Fixes #77. Introduces (baseline) snapshot testing for rbx_binary that can be replicated for rbx_xml using the rbx_dom_test crate.
- Fixes #11. This was pretty trivial as part of adopting reflection information.
- Bumps all crates to have a minimum Rust version of 1.36.0, up from 1.32.0.

## Performance
I built a small performance test suite exercising the limited capabilities of the two implementations. My not-very-scientific results look promising:

image

## TODO
- [x] Deserializer feature parity
- [x] Serializer using reflection information
- [ ] Use reflection-supplied default values in serializer
- [ ] Tests for serializer reflection usage
- [ ] Tests for deserializer reflection usage
- [x] Benchmark performance delta:
    - [x] Serializer
    - [x] Deserializer


rbx_binary/src/types.rs
```diff
@@ -1,159 +1,159 @@
-use std::{
```
moradiaz2685-boop, line 1: Enterado

moradiaz2685-boop (approved):

PR #82, opened 2019-12-10
Add EncodeOutput type to rbx_xml

LPGhatguy:
Experimental PR that introduces a new type, `EncodeOutput`, and changes the public rbx_xml serialization functions to return it.

Currently, it contains the referent map that's built up during serialization. This is part of an experiment to implement sourcemaps for Roblox. Rojo will consume the result of this struct and write a JSON file whose keys are referents and whose values are file paths!


LPGhatguy:
It looks like referents are less available than I expected in other serializers, and so this feature probably can't be based on referents. Looking into using a version of the output from `Instance:GetFullName()`.

PR #86, opened 2020-01-08
Test strategy to get default values via place file saved from Studio

LPGhatguy:
Measuring default values via scripts is finnicky and limited.

A better approach would be to trick Studio into writing all of the default values for every instance into a single file that we can read!


LPGhatguy:
I'm going to rework this to be on top of #86 so that we only have one outstanding PR that rewrites this crate...

LPGhatguy:
This is merged into the other work now.

PR #93, opened 2020-01-22
Add Faces and Axes types to rbx_dom_weak

LPGhatguy:
I ported these types out of #91. If we go down the road of publishing that crate and creating an rbx_dom_weak 2.0, we'll merge the implementations again!

## TODO
- [ ] Docs
- [ ] Conversions to/from integers
- [ ] Presence in `RbxValue`/`RbxValueType`


LPGhatguy:
I think we can hold off on the conversion to rbx_types, which will resolve this!

PR #96, opened 2020-02-06
Break apart rbx_reflection

### LINKED ISSUE #95: Move away from codegen in favor of serialization for reflection
The fact that rbx_reflection is a generated Rust module is a neat trick, but ultimately problematic for bugs and implementation maintenance. We also have to maintain a parallel generator for rbx_dom_lua, and sometimes feature parity is messed up.

Additionally, it becomes difficult to update the reflection information out-of-band. It'd be nice to distribute new reflection information automatically that applications could update to, but baking it into the executable makes that more difficult than it should be.

I think there are a reasonable set of steps to solve this problem:
- Change generate_reflection to output serialized versions of the reflection database, like MessagePack, JSON, or bincode.
- Split rbx_reflection's bundled reflection database into a separate crate that just exposes an embedded serialized database.
- Add configuration for rbx_xml and rbx_binary to let them accept a reference to a reflection database

There may be a small performance cost to switching from generated code to serialized modules, but if we choose the right format, it should be minimal.

LPGhatguy:
Closes #95.

This PR will remove the generated reflection database from rbx_reflection and move it to another crate, rbx_reflection_database. The core types will remain and will become depended on by generate_reflection.

This PR also takes the opportunity to port types over to rbx_types, which is more consistent and well-defined.

## TODO
- Finish rbx_reflection API
    - [x] Docs
    - [x] Stop exposing all bitflags members?
    - [x] ~Find a home for new value resolution code~
        - Deferred
- [x] ~Update rbx_dom_lua to use JSON-format database~
    - This can take advantage of the new JSON file support in Rojo 6.
    - Deferred
- [x] Integrate #86 to measure default values
- [x] ~Fix Infinity serialization in JSON database?~
    - Not sure there's much we can do here right now.
- [x] Fix default values to only contain canonical properties
- [x] Rework definition of "serializable"
    - Is a non-canonical property that has a serializable and canonical variant serializable?
    - Is a canonical property that has a non-canonical but canonically-serializable variant serializable?
    - Can we use an enum to represent serializing as-is, serializing as a different name, or not serializing?
- [x] Change rbx_reflection human-readable tag serialization to be a list of strings


LPGhatguy:
I guess rmp-serde doesn't support some feature that we use in the database. I went through and tried to replace it with another format, but none of them deserialize correctly despite serializing correctly.

All of the crates I tried:

* `rmp-serde`: Deserialization error, something about an enum (`PropertySerialization`) and an invalid length of 1. 🤷 
    * This crate looks like it's stuck in a bad place. Lots of issues like this that can't be fixed because of a pre-1.0 backwards compatibility guarantee.
* `serde_cbor`: Deserialization error, something about a sequence vs a slice. 🤷 
    * This crate is being rewritten from scratch under a new maintainer, so a fix is unlikely.
* `bincode`: Deserialization error, Doesn't support Serde's `deserialize_identifier`. I don't know what that is.
* `ron`: Deserialization panic (!!) with `not implemented: IdDeserializer may only be used for identifiers`
* `serde_json`: Serializes non-finite floats as `null`, which results in corrupt data that can't be deserialized again.
* BSON: Doesn't support unsigned integers; a non-starter.

PR #97, opened 2020-02-07
Port rbx_xml to rbx_dom_weak 2.0

LPGhatguy:
This is a VERY early work in progress. The hope by starting in early is that we can have a functioning format implementation using rbx_dom_weak 2.0, and continue to use rbx_dom_weak 1.0-based implementations if we need to write binary files until it's ported.

This commit is based on #96, which is required to make rbx_reflection have types based on rbx_dom_weak 2.0/rbx_types.


LPGhatguy:
this was not merged.

PR #105, opened 2020-06-18
Remove crazy HTML tables from FORMAT.md

LPGhatguy:
This makes the document much easier to read and easier to edit!

### DIFF
--- a/rbx_binary/FORMAT.md
+++ b/rbx_binary/FORMAT.md
@@ -7,6 +7,7 @@ This document is based on:
 - Observing `rbxm` and `rbxl` output from Roblox Studio
 
 ## Contents
+- [Document Conventions](#document-conventions)
 - [File Structure](#file-structure)
 - [File Header](#file-header)
 - [Chunks](#chunks)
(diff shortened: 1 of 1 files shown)


rbx_binary/FORMAT.md
```diff
@@ -53,71 +53,19 @@ This document is based on:
...
-		<td colspan="6">Signature (<code>89 ff 0d 0a 1a 0a</code>)</td>
-		<td colspan="2">Version (<code>0</code>)</td>
-	</tr>
-	<tr>
-		<td colspan="4">Number of instance types (<code>u32</code>)</td>
-		<td colspan="4">Number of instances (<code>u32</code>)</td>
-		<td colspan="8">Reserved bytes (always 0)</td>
-	</tr>
-</table>
+Every file starts with a 16 byte header.
```
dekkonot, line 56: `Number of instance types`, `Number of Instances` and `Reserved bytes` appear to have disappeared!

LPGhatguy: oops

dekkonot:
I'm not sure if totally removing the tables is a good idea. It definitely makes it easier to edit, but it hurts readability quite a bit. Pure markdown tables might be a nice in-between. You lose a bit of the formatting, but it definitely looks nicer than a bulleted list:

|       Magic Number      |          Signature         |        Version       |
|:-----------------------:|:--------------------------:|:--------------------:|
|   8 bytes (`<roblox!`)  | 6 bytes (`89 ff 0a 1a 0a`) |          u16         |
| **Instance Type Count** |     **Instance Count**     |  **Reserved Space**  |
|           i32           |             i32            | 8 bytes (always `0`) |

PR #106, opened 2020-06-20
Add SSTR and SharedString to binary format spec

dekkonot:


rbx_binary/FORMAT.md
```diff
@@ -116,6 +120,26 @@ Observed metadata entries and their values:
...
+|:--------------------|:----------------------|:---------------------------------------------|
+| Version             | u32                   | The version of the `SSTR` chunk (always `0`) |
+| Shared String Count | u32                   | The number of SharedStrings in the chunk     |
+| Strings             | Array(Shared Strings) | The actual shared string entries             |
+
+A shared string entry looks like this:
+
+| Field Name    | Format   | Value                                                                   |
+|:--------------|:---------|:------------------------------------------------------------------------|
+| MD5 Hash      | 16 bytes | An [MD5](https://en.wikipedia.org/wiki/MD5) hash of the `Shared String` |
```
LPGhatguy, line 136: Do you know if this hash is used by anything that loads these files? I remember some discussion internally at Roblox about how MD5 can't be trusted, but instead of picking a better hash function, the decision was just to ignore it sometimes.

I do also know that there is work being done on replacing MD5 in shared strings. I don't know if this will be a new version of the `SSTR` chunk (most hashes are bigger than 16 bytes).

dekkonot: I didn't check everything, though there's not many places where it could be used. I can check tomorrow.

LPGhatguy: I think what you've got here is what's _supposed_ to be in that field, so no big deal for now. 😄

PR #109, opened 2020-06-23
Implement Int32

kennethloeffler:


LPGhatguy:
Is this depending on a branch of rbx-test-files still? If possible, we should ensure it's pointing to the latest master commit.

PR #113, opened 2020-06-25
Implement UDim and update test-files

kennethloeffler:
Note https://github.com/Roblox/rbx-dom/commit/89a31273fcb81fef2e9097481eae160b884525be - is this ok?

### DIFF
--- a/rbx_binary/src/tests/serializer.rs
+++ b/rbx_binary/src/tests/serializer.rs
@@ -1,5 +1,5 @@
 use rbx_dom_weak::{
-    types::{Ref, UDim},
+    types::{Ref, Region3, Vector3},
     InstanceBuilder, WeakDom,
 };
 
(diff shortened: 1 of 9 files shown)


rbx_binary/src/tests/serializer.rs
```diff
@@ -56,9 +56,10 @@ fn unknown_property() {
 /// This test will need to be updated once we implement the type used here.
 #[test]
 fn unimplemented_type_known_property() {
-    let tree = WeakDom::new(
-        InstanceBuilder::new("UIListLayout").with_property("Padding", UDim::new(1.0, -30)),
-    );
+    let tree = WeakDom::new(InstanceBuilder::new("UIListLayout").with_property(
+        "Padding",
+        Region3::new(Vector3::new(0.0, 0.0, 50.0), Vector3::new(0.0, 0.0, 50.0)),
+    ));
```
LPGhatguy, line 62: This is kind of funny, can we pick a property that definitely doesn't exist here instead of one that does, but with the wrong type?

PR #114, opened 2020-06-25
Fix UDim example in binary.md

dekkonot:
I did an oopsie and the bytes were in the wrong order.


format/binary.md
```diff
@@ -358,6 +358,8 @@ The **correct** interpretation of this data, with accumulation, is:
 ### Color3uint8
 **Type ID 0x1A**
 
+The `Color3uint8` type is stored as a sequence of bytes that
```
LPGhatguy, line 361: Unfinished statement here. Can we push this off into another PR?

Issue #89, opened 2020-01-15
rbx_xml Axes support


LPGhatguy:
The `Axes` type is only used in the `ArcHandles` instance.

PR #119, opened 2020-06-29
rbx_binary Implement Axes

kennethloeffler:


LPGhatguy:
Can you split apart the XML and binary format implementations into two PRs? Because they don't share code, reviewing one should not block the other!

If this closes an issue like #89, you can use [GitHub closing keywords](https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords) to make GitHub automatically close the related issue when this PR is merged.

format/binary.md
```diff
@@ -439,7 +443,7 @@ When interleaved, the same array would instead look like this:
 |:-----:|:--------:|:--------:|:--------:|:--------:|
 | Row 1 | `A0`     | `B0`     | `C0`     | `D0`     |
 | Row 2 | `A1`     | `B1`     | `C1`     | `D1`     |
-| Row 3 | `A2`     | `B2`     | `C2`     | `D2`     |
+p| Row 3 | `A2`     | `B2`     | `C2`     | `D2`     |
```
dekkonot, line 446: That `p` shouldn't be there

kennethloeffler: I had to fix this in the other pr too oof x_x

I’ll get on it shortly :)

rbx_binary/src/text_deserializer.rs
```diff
@@ -10,6 +10,8 @@ use byteorder::{LittleEndian, ReadBytesExt};
 use rbx_dom_weak::types::{Color3, UDim, UDim2, Vector2};
 use serde::{Deserialize, Serialize};
 
+use rbx_dom_weak::types::Axes;
+
```
LPGhatguy, line 14: We can merge these imports.

```suggestion
use rbx_dom_weak::types::{Axes, Color3, UDim, UDim2, Vector2};
use serde::{Deserialize, Serialize};

```

rbx_binary/src/deserializer.rs
```diff
@@ -474,7 +474,27 @@ impl<R: Read> BinaryDeserializer<R> {
...
             Type::Ray => {}
             Type::Faces => {}
-            Type::Axes => {}
+            Type::Axes => match canonical_type {
+                VariantType::Axes => {
+                    for referent in &type_info.referents {
+                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
+                        let value = chunk.read_u8()?;
+                        let rbx_value =
+                            Variant::Axes(Axes::from_bits(value).unwrap_or(Axes::empty()));
```
LPGhatguy, line 483: If there are unnecessary bits set here, this will silently assign an empty set of axes. We should raise an error here instead!

If we do want to keep this fallback to default axes, you can use `.unwrap_or_else(Axes::empty)`, which would defer the `Axes::empty()` call I believe.

kennethloeffler: Should I add a new variant of `InnerError` for an error here, e.g. `InnerError::InvalidPropData`?

LPGhatguy: Yeah, that would work for me. Similar to the other errors we have right now, making them hold strings describing which instance, which property, what we expected, and what we got would be great.

rbx_binary/src/deserializer.rs
```diff
@@ -474,7 +482,35 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                        let rbx_value = Variant::Axes(if let Some(axes) = Axes::from_bits(value) {
+                            axes
+                        } else {
+                            return Err(InnerError::InvalidPropData {
+                                type_name: type_info.type_name.clone(),
+                                prop_name,
+                                valid_value: "less than 7",
+                                actual_value: format!("{}", value),
+                            });
+                        });
```
LPGhatguy, line 499: Can you write this block using [`Option::ok_or_else`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or_else) and the question mark operator instead? It'll turn out a little nicer!

rbx_binary/src/deserializer.rs
```diff
@@ -474,7 +482,35 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
+                        let value = chunk.read_u8()?;
+                        let rbx_value = Variant::Axes(if let Some(axes) = Axes::from_bits(value) {
+                            axes
+                        } else {
+                            return Err(InnerError::InvalidPropData {
+                                type_name: type_info.type_name.clone(),
+                                prop_name,
+                                valid_value: "less than 7",
+                                actual_value: format!("{}", value),
```
Kampfkarren, line 497: value.to_string()

PR #128, opened 2020-10-26
Implement CFrame (mostly) in rbx_binary

Dekkonot:
Implementation reads and writes CFrames, but doesn't write the CFrame special cases any differently than normal CFrames. Instead, they're just written as raw matrices.


rbx_binary/src/deserializer.rs
```diff
@@ -77,6 +80,164 @@ pub(crate) enum InnerError {
...
+        prop_name: String,
+        id: u8,
+    },
+}
+
+// TODO potentially move this to a different file if/when we do the inverse for serializing
+pub(crate) fn special_case_to_rotation(id: u8) -> Option<Matrix3> {
+    match id {
+        0x02 => Some(Matrix3::new(
+            //02 -> (1, 0, 0) (0, 1, 0) (0, 0, 1)
```
LPGhatguy, line 96: Do you think we need these comments? They look like they might be left over from when you were writing out the code from the cases.

rbx_binary/src/deserializer.rs
```diff
@@ -591,7 +752,73 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                    chunk.read_interleaved_f32_array(&mut y)?;
+                    chunk.read_interleaved_f32_array(&mut z)?;
+
+                    for i in 0..referents.len() {
+                        let instance = self.instances_by_ref.get_mut(&referents[i]).unwrap();
+                        instance.properties.push((
+                            canonical_name.clone(),
+                            Variant::CFrame(CFrame::new(
+                                Vector3::new(x[i], y[i], z[i]),
+                                rotations[i],
```
LPGhatguy, line 808: Can we do the iterator zipping dance like in the other implementations instead of using array indexing here? Each array access forces the compiler to check if our access is in bounds of the array, while using iterator combinators can help the compiler elide them. Because of this, numeric for loops in Rust loop are actually sort of rare!

In this particular case, I'd probably zip `x`, `y`, and `z` into a iterator that yields `Vector3`, then zip it with `referents` and `rotations`.

rbx_binary/src/serializer.rs
```diff
@@ -623,6 +626,44 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+                                x.push(value.position.x);
+                                y.push(value.position.y);
+                                z.push(value.position.z);
+                            } else {
+                                return type_mismatch(i, &rbx_value, "CFrame");
+                            }
+                        }
+
+                        for matrix in rotations {
+                            //TODO handle special cases
```
LPGhatguy, line 647: Can you file an issue and link it here? This comment is vague if the reader is unaware of the special forms that CFrames have. Maybe we should indicate that writing `0` here indicates the default CFrame representation.

rbx_binary/src/text_deserializer.rs
```diff
@@ -286,6 +290,51 @@ impl DecodedValues {
...
+                reader.read_interleaved_f32_array(&mut x).unwrap();
+                reader.read_interleaved_f32_array(&mut y).unwrap();
+                reader.read_interleaved_f32_array(&mut z).unwrap();
+
+                let mut values =
+                    vec![CFrame::new(Vector3::new(0.0, 0.0, 0.0), Matrix3::identity()); prop_count];
+
+                for i in 0..prop_count {
+                    values[i] = CFrame::new(Vector3::new(x[i], y[i], z[i]), rotations[i])
+                }
```
LPGhatguy, line 334: Same sort of deal with iterator zipping here as above. This code is quite a bit shorter and more readable than the iterator zipping code, I wonder if we can discern whether there's a perf loss from the assembly...

rbx_types/src/basic_types.rs
```diff
@@ -142,6 +142,9 @@ impl Matrix3 {
             z: Vector3::new(0.0, 0.0, 1.0),
         }
     }
+    pub fn new(x: Vector3, y: Vector3, z: Vector3) -> Self {
+        Self { x, y, z }
+    }
```
LPGhatguy, line 147: Can we move `new` above `identity` and throw a blank line between them? This is sort of the default constructor, which I like to have first conventionally.

LPGhatguy (changes requested):

rbx_binary/src/serializer.rs
Dekkonot: One might argue that people shouldn't be touching that code that writes CFrames if they're not familiar with the format! 😛

Linking it to an issue is a good idea regardless though, so I'll do that.

rbx_binary/src/text_deserializer.rs
Dekkonot: The zipping code turns out to not be that much less readable, so I think I'm okay with switching to it; the performance benefit of not indexing 5 vectors is probably worth it.

Dekkonot:
I'm not sure why those tests passed considering Vector3 snapshots don't exist on this branch, but I'll take it.

rbx_binary/src/serializer.rs
```diff
@@ -644,6 +645,46 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+                                x.push(value.position.x);
+                                y.push(value.position.y);
+                                z.push(value.position.z);
+                            } else {
+                                return type_mismatch(i, &rbx_value, "CFrame");
+                            }
+                        }
+
+                        for matrix in rotations {
+                            //TODO write special cases; see: https://github.com/rojo-rbx/rbx-dom/issues/129
```
LPGhatguy, line 666:
```suggestion
                            // TODO write special cases; see: https://github.com/rojo-rbx/rbx-dom/issues/129
```
Small space fix

LPGhatguy (commented):
Looking great! Posted a suggestion for a one space insertion, and then it looks perfect.

LPGhatguy (approved):

PR #130, opened 2020-10-28
Update rbx_reflection patch format, add version back to rbx_reflection_database

LPGhatguy:
This was work I started awhile ago, reviving it.


Kampfkarren:
Is this the one that can be dynamically updated? :eyes:

LPGhatguy:
The database can actually already be dynamically updated on the master branch, this PR just inches us a bit closer to being able to actually ship that database!

PR #131, opened 2020-10-29
Remove byte_order as a dependency of rbx_binary

Dekkonot:
Removes byte_order as a dependency of rbx_binary.

Doesn't implement new methods for niche data types like f64 and u16; these aren't used in many places, so it seems unnecessary.

Instead, those functions are replaced with inlines such as:
```rust
{
    let mut bytes = [0; 8];
    chunk.read_exact(&mut bytes)?;
    f64::from_le_bytes(bytes)
}
```

I also resisted the urge to clean up any of the other functions since I wanted this to be a pure commit, but expect something like that in the future. Notably, I would like to (eventually) figure out a way to swap the `read_interleaved_TYPE_array` functions to also using `to_le_bytes` and `from_le_bytes`.

### DIFF
--- a/rbx_binary/Cargo.toml
+++ b/rbx_binary/Cargo.toml
@@ -14,7 +14,6 @@ edition = "2018"
 unstable_text_format = ["base64", "serde"]
 
 [dependencies]
-byteorder = "1.2.7"
 log = "0.4.6"
 lz4 = "1.23.1"
 rbx_dom_weak = { path = "../rbx_dom_weak" }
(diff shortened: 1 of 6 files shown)


LPGhatguy (changes requested):
Wow, that conversion was pretty straightforward!

Can we implement the methods for `f64` and `u16` that we would use? Though I understand that they're not used much now, I think the code is easier to read when it gets to just write `read_le_f64` instead of creating an array and filling it.

LPGhatguy (approved):
approving le pull request

PR #132, opened 2020-11-05
Implement EnumValue in rbx_binary

LPGhatguy:
This PR implements `Variant::EnumValue` for rbx_binary. I introduced a new `read_interleaved_u32_array` method that operates very similarly to its i32 and f32 brethren, but uses the new `from_be_bytes` method instead of manual bit shifting. From there, most of the code is the same as the `Int32` implementation.

Most of the diffs in this PR are from snapshots that had enums that were previously ignored. This provided a pretty good scattershot sample of enums working, actually.

## TODO
* [x] Serialization


rbx_binary/src/core.rs
```diff
@@ -91,6 +91,24 @@ pub trait RbxReadExt: Read {
         Ok(())
     }
 
+    fn read_interleaved_u32_array(&mut self, output: &mut [u32]) -> io::Result<()> {
+        let mut buffer = vec![0; output.len() * mem::size_of::<u32>()];
```
Dekkonot, line 95: `u32` has an explicitly defined size in Rust (4 bytes). Do we need to use `mem::size_of`?

Dekkonot (approved):

rbx_binary/src/core.rs
Kampfkarren: It makes it more explicit what the number is for, and it compiles down anyway.

Dekkonot: Forgot that it's determined at compilation. Oops! Nevermind!

PR #133, opened 2020-11-05
rbx_binary implement Color3uint8

kennethloeffler:
As part of this PR I have made `serializer.rs` [use canonical property names when fetching property values to write, rather than the serialized names](https://github.com/rojo-rbx/rbx-dom/blob/1b2cfa7efef6556504fc9a732f13cbcad3615523/rbx_binary/src/serializer.rs#L446). This became an issue because `deserializer.rs` uses canonical names as keys for property maps. This caused the serializer to erroneously write default prop values since `instance.properties.get("Color3uint8")` results in a `None`. I alternatively could have made `deserializer.rs` use `"Color3uint8"` as the key, but that seems smelly to me.

I have also updated the test files submodule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside 
 | BrickColor         | `Part.BrickColor`               | ✔ | ✔ | ✔ | ❌ |
 | CFrame             | `Camera.CFrame`                 | ✔ | ✔ | ✔ | ➖ |
 | Color3             | `Lighting.Ambient`              | ✔ | ✔ | ✔ | ✔ |
-| Color3uint8        | `N/A`                           | ✔ | ✔ | ✔ | ❌ |
+| Color3uint8        | `N/A`                           | ✔ | ✔ | ✔ | ✔ |
 | ColorSequence      | `Beam.Color`                    | ✔ | ✔ | ✔ | ❌ |
 | Content            | `Decal.Texture`                 | ✔ | ✔ | ✔ | ✔ |
 | Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ❌ |
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside 
 | Color3uint8        | `N/A`                           | ✔ | ✔ | ✔ | ✔ |
 | ColorSequence      | `Beam.Color`                    | ✔ | ✔ | ✔ | ❌ |
 | Content            | `Decal.Texture`                 | ✔ | ✔ | ✔ | ✔ |
-| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ❌ |
+| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ✔ |
 | Faces              | `Handles.Faces`                 | ✔ | ❌ | ✔ | ✔ |
 | Float32            | `Players.RespawnTime`           | ✔ | ✔ | ✔ | ✔ |
 | Float64            | `Sound.PlaybackLoudness`        | ✔ | ✔ | ✔ | ✔ |
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside 
 | Color3uint8        | `N/A`                           | ✔ | ✔ | ✔ | ✔ |
 | ColorSequence      | `Beam.Color`                    | ✔ | ✔ | ✔ | ❌ |
 | Content            | `Decal.Texture`                 | ✔ | ✔ | ✔ | ✔ |
-| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ✔ |
+| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ❌ |
 | Faces              | `Handles.Faces`                 | ✔ | ❌ | ✔ | ✔ |
 | Float32            | `Players.RespawnTime`           | ✔ | ✔ | ✔ | ✔ |
 | Float64            | `Sound.PlaybackLoudness`        | ✔ | ✔ | ✔ | ✔ |
--- a/rbx_binary/src/deserializer.rs
+++ b/rbx_binary/src/deserializer.rs
@@ -859,21 +859,13 @@ impl<R: Read> BinaryDeserializer<R> {
             Type::Color3uint8 => match canonical_type {
                 VariantType::Color3 => {
                     let len = type_info.referents.len();
-                    let mut r = Vec::with_capacity(len);
-                    let mut g = Vec::with_capacity(len);
-                    let mut b = Vec::with_capacity(len);
+                    let mut r = vec![0; len];
+                    let mut g = vec![0; len];
+                    let mut b = vec![0; len];
 
-                    for _ in 0..len {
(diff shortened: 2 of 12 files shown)


rbx_binary/src/deserializer.rs
```diff
@@ -856,7 +856,45 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                    for _ in 0..len {
+                        r.push(chunk.read_u8().unwrap());
+                    }
+
+                    for _ in 0..len {
+                        g.push(chunk.read_u8().unwrap());
+                    }
+
+                    for _ in 0..len {
+                        b.push(chunk.read_u8().unwrap());
```
Kampfkarren, line 875: Are you sure you want to be using `unwrap()` here instead of `?`?

rbx_binary/src/text_deserializer.rs
```diff
@@ -401,6 +402,32 @@ impl DecodedValues {
...
+                for _ in 0..prop_count {
+                    r.push(reader.read_u8().unwrap());
+                }
+
+                for _ in 0..prop_count {
+                    g.push(reader.read_u8().unwrap());
+                }
+
+                for _ in 0..prop_count {
+                    b.push(reader.read_u8().unwrap());
```
Kampfkarren, line 419: here too

Dekkonot: `read_u8` returns a Result whereas the parent function here returns an Option. If you look at the file, all of the other read functions use `unwrap` too.

rbx_binary/src/deserializer.rs
```diff
@@ -856,7 +856,45 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                    }
```
LPGhatguy, line 876: Instead of writing these loops by hand, we can use Rust's [Read::read_exact](https://doc.rust-lang.org/stable/std/io/trait.Read.html#method.read_exact).

```rust
let mut r = vec![0; len];
chunk.read_exact(r.as_mut_slice())?;
```

...for each color.

rbx_binary/src/serializer.rs
```diff
@@ -696,6 +704,33 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+                        let mut r = Vec::with_capacity(values.len());
+                        let mut g = Vec::with_capacity(values.len());
+                        let mut b = Vec::with_capacity(values.len());
+
+                        for (i, rbx_value) in values {
+                            if let Variant::Color3uint8(value) = rbx_value.as_ref() {
+                                r.push(value.r);
+                                g.push(value.g);
+                                b.push(value.b);
+                            } else {
```
LPGhatguy, line 717: We should probably also support `Color3` values here, since they're often trivially converted. If the `Color3`'s range does not fit within a `Color3uint8`, it should be clamped.

rbx_binary/src/serializer.rs
```diff
@@ -696,6 +704,33 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+                            chunk.write_u8(component)?;
+                        }
+
+                        for component in g {
+                            chunk.write_u8(component)?;
+                        }
+
+                        for component in b {
+                            chunk.write_u8(component)?;
+                        }
```
LPGhatguy, line 732: Similar to the write examples, we have a method for this from the Rust standard library, [`Write::write_all`](https://doc.rust-lang.org/stable/std/io/trait.Write.html#method.write_all).

We should write something like...

```rust
chunk.write_all(r.as_slice())?;
```

...for each color.

rbx_binary/src/serializer.rs
```diff
@@ -443,7 +443,15 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
                                 .properties
-                                .get(prop_name)
+                                .get(
+                                    if let Some(descriptor) =
+                                        find_canonical_property_descriptor(type_name, prop_name)
+                                    {
+                                        &*descriptor.name
+                                    } else {
+                                        prop_name
+                                    },
```
LPGhatguy, line 453: This is a much too complicated expression to slap inline here. I need to think over the implications of this as well.

README.md
```diff
@@ -66,10 +66,10 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside
 | BrickColor         | `Part.BrickColor`               | ✔ | ✔ | ✔ | ❌ |
 | CFrame             | `Camera.CFrame`                 | ✔ | ✔ | ✔ | ➖ |
 | Color3             | `Lighting.Ambient`              | ✔ | ✔ | ✔ | ✔ |
-| Color3uint8        | `N/A`                           | ✔ | ✔ | ✔ | ❌ |
+| Color3uint8        | `N/A`                           | ✔ | ✔ | ✔ | ✔ |
 | ColorSequence      | `Beam.Color`                    | ✔ | ✔ | ✔ | ❌ |
 | Content            | `Decal.Texture`                 | ✔ | ✔ | ✔ | ✔ |
-| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ❌ |
+| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ✔ |
```
LPGhatguy, line 72: This is an unrelated change that I've already landed in master. I'd prefer if changes like this were not included in feature PRs.

LPGhatguy (changes requested):

rbx_binary/src/serializer.rs
LPGhatguy: This is also a problem because we're redundantly looking up the same property descriptor in the database for every instance. This process is expensive and can take a long time.

LPGhatguy: I am working on what I think is the correct solution to this problem. We should avoid introducing a fix here, especially as part of a regular property feature implementation.

kennethloeffler: Should I have `deserializer.rs` use `Color3uint8` as the key for the property map instead of the canonical name (as described in the original PR comment)? Or perhaps check explicitly in `serializer.rs` (similar to `Name`) and do `instance.properties.get("Color")`? I'm grepping through the lua database and it looks like this type is exclusively used by `BasePart` and friends, so either way seems like it would work. Your call.

kennethloeffler:
Leaving that problem alone, I've made the other changes.

rbx_binary/src/serializer.rs
LPGhatguy: We should do neither here, the change I'm working on should solve this problem with no hard-coded special cases. `Color` is not quite the same level of specialness as `Name` 😄

rbx_binary/src/tests/serializer.rs
```diff
@@ -124,3 +124,30 @@ fn logical_properties_basepart_size() {
...
+fn color3_to_color3uint8() {
+    let tree = WeakDom::new(
+        InstanceBuilder::new("Folder")
+            .with_child(
+                InstanceBuilder::new("Part")
+                    .with_property("Color3uint8", Color3::new(-0.25, 0.5, 1.2)),
+            )
+            .with_child(
+                InstanceBuilder::new("Part")
+                    .with_property("Color", Color3::new(0.1111, 0.3333, 0.9999)),
```
LPGhatguy, line 141: Can we change this property to be `Color3uint8` with a `Color3uint8` value? As is, this is redundant with the third instance. Testing all of the valid combinations here is a good idea I think.

LPGhatguy (approved):

PR #134, opened 2020-11-06
Fix rbx_binary's relationship to canonical and logical properties

LPGhatguy:
This is a pretty involved change; I hope I've added enough documentation inline in the code to help guide people through the new design decisions here.

## Logical Properties
This is a concept that has existed in rbx-dom informally for awhile now, but I think I've finally found a good name: logical properties.

A logical property describes an attribute of an instance, and may have multiple property names. We already have special terms to describe two kinds of property names: canonical names and serialized names. The idea is that all names for a logical property have storage that aliases. One of them must be canonical and zero or one of them must be marked for serialization.

Examples of groups of property names that describe logical properties:
* `BasePart.Size` (canonical), `BasePart.size` (serialized)
* `BasePart.Color` (canonical), `BasePart.Color3uint8` (serialized), `BasePart.BrickColor` (other), `BasePart.brickColor` (other)
* `SpawnLocation.AllowTeamChangeOnTouch` (canonical, serialized)

## PropInfo's New Semantics
Previously, rbx_binary's serializer would create a new `PropInfo` struct for each unique property name found for each class. This meant that the same logical property could be represented multiple times, and for properties whose canonical and serialized names are different, resulted in issues. The need for this fix was brought up by #133.

In this PR, exactly one `PropInfo` is created for each logical property. `TypeInfo` now explicitly contains a map from _canonical_ property names to `PropInfo` structs, and the serialized name is stored inline in `PropInfo`.

Additionally, I've added a set of aliases for each `PropInfo`. This set will be empty for most instances, but helps consolidate and deduplicate values with different names that form the same logical property. This would come up when mixing an older model with parts that have `BrickColor` alongside a newer model with parts that have `Color3uint8`. The `BrickColor` parts will be automatically upgraded to have `Color3uint8` instead, which should match the behavior of Roblox Studio.

One nice side effect of this change is that we're now able to borrow property name data in a few more places, as most names will come from static data located in the reflection database.

### DIFF
--- a/rbx_binary/src/core.rs
+++ b/rbx_binary/src/core.rs
@@ -3,7 +3,7 @@ use std::{
     mem,
 };
 
-use rbx_reflection::{PropertyDescriptor, PropertyKind, PropertySerialization};
+use rbx_reflection::{ClassDescriptor, PropertyDescriptor, PropertyKind, PropertySerialization};
 
 pub static FILE_MAGIC_HEADER: &[u8] = b"<roblox!";
 pub static FILE_SIGNATURE: &[u8] = b"\x89\xff\x0d\x0a\x1a\x0a";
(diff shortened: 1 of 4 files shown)


rbx_binary/src/core.rs
```diff
@@ -314,94 +314,101 @@ pub fn find_canonical_property_descriptor(
...
-                                    .get(serialized_name.as_ref())
-                                    .unwrap();
-
-                                return Some((canonical_descriptor, serialized_descriptor));
-                            }
-                            _ => unimplemented!(),
-                        },
-                        _ => return None,
+                    match &canonical.kind {
+                        PropertyKind::Canonical { serialization } => {
```
Kampfkarren, line 374: `if let PropertyKind::Canonical { serialization } = &canonical.kind {`

Dekkonot (approved):
I'm not running clippy over this right now (we should do that for the entire workspace eventually...) so there's a chance I missed something but after looking through this, this seems to make sense.

LPGhatguy:
Let's do it then!

PR #136, opened 2020-11-07
Implement BrickColor in rbx_binary

Dekkonot:
Nothing unusual in here. This implementation returns an Err (in deserializer) or panics (in text_deserializer) when it encounters any unknown BrickColor codes, which matches the behavior of other types when they encounter a bad piece of data.

### DIFF
--- a/rbx_binary/src/deserializer.rs
+++ b/rbx_binary/src/deserializer.rs
@@ -684,13 +684,11 @@ impl<R: Read> BinaryDeserializer<R> {
                         let color = value
                             .try_into()
                             .map(|value| BrickColor::from_number(value).unwrap())
-                            .map_err(|_| {
-                                InnerError::InvalidPropData {
-                                    type_name: type_info.type_name.clone(),
-                                    prop_name: prop_name.clone(),
-                                    valid_value: "a valid BrickColor",
-                                    actual_value: value.to_string(),
-                                }
+                            .map_err(|_| InnerError::InvalidPropData {
+                                type_name: type_info.type_name.clone(),
+                                prop_name: prop_name.clone(),
+                                valid_value: "a valid BrickColor",
+                                actual_value: value.to_string(),
                             })?;
                         instance
                             .properties
--- a/rbx_binary/src/deserializer.rs
+++ b/rbx_binary/src/deserializer.rs
@@ -7,8 +7,8 @@ use std::{
 
 use rbx_dom_weak::{
     types::{
-        Axes, CFrame, Color3, Color3uint8, EnumValue, Faces, Matrix3, Ref, UDim, UDim2, Variant,
-        VariantType, Vector2, Vector3,
+        Axes, BrickColor, CFrame, Color3, Color3uint8, EnumValue, Faces, Matrix3, Ref, UDim, UDim2,
+        Variant, VariantType, Vector2, Vector3,
     },
     InstanceBuilder, WeakDom,
 };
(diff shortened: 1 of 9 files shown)


rbx_binary/src/deserializer.rs
```diff
@@ -668,7 +668,36 @@ impl<R: Read> BinaryDeserializer<R> {
...
             },
-            Type::BrickColor => {}
+            Type::BrickColor => match canonical_type {
+                VariantType::BrickColor => {
+                    let mut values = vec![0; type_info.referents.len()];
+                    chunk.read_interleaved_u32_array(&mut values)?;
+
+                    for (value, referent) in values.into_iter().zip(&type_info.referents) {
+                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
+                        let color = BrickColor::from_number(value as u16).ok_or_else(|| {
```
LPGhatguy, line 678: This cast won't do what we want for values that don't fit in a `u16`. I see two ways we can fix this:

1. Use `value.try_into()` (from the `std::convert::TryInto` trait), which will let us handle out-of-range values.
2. Switch `BrickColor` to be backed by a `u32`. It's not clear to me if `BrickColor` is 32 bits in the file format because it was convenient for Roblox, or if that's the intended size of a `BrickColor`.

rbx_binary/src/tests/models.rs
```diff
@@ -42,4 +42,5 @@ binary_tests! {
     cframe_special_cases,
     cframe_case_mixture,
     three_unique_parts,
+    three_brickcolorvalues,
```
LPGhatguy, line 45: Huh. I didn't know this was a type.

rbx_binary/src/deserializer.rs
Dekkonot: My guess is because it's convenient more than it being 'correct'. I don't think there's any reason for us to explicitly mirror the size in rbx-dom though.

I suppose we can also use "try_into" though. The end-goal is the same; we'll be raising an identical error.

LPGhatguy: To be clear, using `as u16` will never raise an error. It will silently wrap and put a smaller value into `BrickColor::new`. See this playground for an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2a8058e6e4ac5b83fdbafd0258272e7e

rbx_binary/src/tests/models.rs
Dekkonot: I didn't either until I started implementing this! It ended up being a happy coincidence.

rbx_binary/src/deserializer.rs
```diff
@@ -674,7 +674,36 @@ impl<R: Read> BinaryDeserializer<R> {
...
+            Type::BrickColor => match canonical_type {
+                VariantType::BrickColor => {
+                    let mut values = vec![0; type_info.referents.len()];
+                    chunk.read_interleaved_u32_array(&mut values)?;
+
+                    for (value, referent) in values.into_iter().zip(&type_info.referents) {
+                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
+                        let color = value
+                            .try_into()
+                            .map(|value| BrickColor::from_number(value).unwrap())
```
LPGhatguy, line 686: We should not use `unwrap` in the main serializer/deserializer codebase: we don't want to panic if a file has an invalid BrickColor value: we should instead return an error.

I would write this chain with something like:

```rust
let color = value
    .try_into()
    .ok() // turn `Result<T, E>` into `Option<T>`, since we don't use the error anyways
    .and_then(BrickColor::from_number)
    .ok_or_else(|| InnerError::InvalidPropData {
        type_name: type_info.type_name.clone(),
        prop_name: prop_name.clone(),
        valid_value: "a valid BrickColor",
        actual_value: value.to_string(),
    })?;
```

Dekkonot: Ah darn it, I knew I forgot something. That's on me, I know better than to leave `unwrap` in production code.

I also forgot that `ok` was a thing. That's a lot more elegant. I'll update it tomorrow (or if you want to suggest a change, I'll accept it).

LPGhatguy: Let's just wait until tomorrow!

Nezuo:
BrickColor needs to be marked as implemented in the ``README.md``.

LPGhatguy (approved):

PR #139, opened 2020-11-10
Tidy up the rbx_binary deserializer

LPGhatguy:
This PR changes the strategy that rbx_binary's deserializer uses when constructing instances. Instead of pushing properties into its `Instance` type and constructing `InstanceBuilder` objects when actually populating the `WeakDom`, we now incrementally build `InstanceBuilder` objects. This has the advantage of assigning referents much earlier in the process, which makes implementing the `Ref` type much easier.

As part of this change, I moved the portion of the code where we special case the `Name` property. I think the new home is a better spot.

Incidentally, this change let us drop a lot of boilerplate around constructing `Variant` objects because the `InstanceBuilder` methods are generic. This might increase compile times a small amount, but it is an ergonomic win that we should keep unless the drop is large. These monomorphizations are also likely to be instantiated by other code, so I'm not too worried.

### DIFF
--- a/rbx_binary/src/deserializer.rs
+++ b/rbx_binary/src/deserializer.rs
@@ -7,8 +7,9 @@ use std::{
 
 use rbx_dom_weak::{
     types::{
-        Axes, BrickColor, CFrame, Color3, Color3uint8, CustomPhysicalProperties, EnumValue, Faces,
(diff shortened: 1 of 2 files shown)


LPGhatguy:
Well, clippy passed, so this code must be perfect. Let's go!

PR #140, opened 2020-11-10
rbx_binary implement Ref

kennethloeffler:
I changed the visibility of `InstanceBuilder.referent` to public because I couldn't see another way to get to the `Ref` in the binary deserializer - maybe this should be different?


rbx_dom_weak/src/instance.rs
```diff
@@ -33,11 +33,13 @@ let dom = WeakDom::new(data_model);
...
 #[derive(Debug)]
 pub struct InstanceBuilder {
-    pub(crate) referent: Ref,
     pub(crate) name: String,
     pub(crate) class: String,
     pub(crate) properties: HashMap<String, Variant>,
     pub(crate) children: Vec<InstanceBuilder>,
+
+    /// The referent of the instance that this `InstanceBuilder` refers to.
+    pub referent: Ref,
```
LPGhatguy, line 42: I don't think we can just make this field public, it would not be allowed to mutate it to any value as this would violate a lot of assumptions. We can add a getter instead, named `referent()`.

rbx_binary/src/tests/snapshots/rbx_binary__tests__util__axes__decoded.snap
```diff
@@ -6,6 +6,7 @@ expression: decoded_viewed
   name: ""
   class: ArcHandles
   properties:
+    Adornee: "[unknown ID]"
```
LPGhatguy, line 9: Ah! I'm going to land a change so that these will end up as zeroes instead of being reported unknown ID.

LPGhatguy: Ok! If you merge the master branch in, these values should be replaced with `"null"` explicitly, instead of an unknown value.

kennethloeffler: Lookin good!

rbx_binary/src/deserializer.rs
```diff
@@ -861,7 +861,36 @@ impl<R: Read> BinaryDeserializer<R> {
                     });
                 }
             },
-            Type::Ref => {}
+            Type::Ref => match canonical_type {
+                VariantType::Ref => {
+                    let mut differences = vec![0; type_info.referents.len()];
+                    chunk.read_interleaved_i32_array(&mut differences)?;
```
Dekkonot, line 867: We have a function for reading referents here that we should probably use instead
https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_binary/src/core.rs#L127

rbx_binary/src/serializer.rs
```diff
@@ -803,6 +803,30 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+                                }
+
+                                buf.push(curr_ref - prev_ref);
+                                prev_ref = curr_ref;
+                            } else {
+                                return type_mismatch(i, &rbx_value, "Ref");
+                            }
+                        }
+
+                        chunk.write_interleaved_i32_array(buf.into_iter())?;
```
Dekkonot, line 828: Same with a function to write referents: https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_binary/src/core.rs#L261

rbx_binary/src/text_deserializer.rs
```diff
@@ -362,6 +363,20 @@ impl DecodedValues {
 
                 Some(DecodedValues::Enum(values))
             }
+            Type::Ref => {
+                let mut differences = vec![0; prop_count];
+                reader.read_interleaved_i32_array(&mut differences).unwrap();
```
Dekkonot, line 368: Can use `read_referent` here too

Dekkonot (changes requested):

Dekkonot (approved):

rbx_binary/src/serializer.rs
```diff
@@ -803,6 +803,28 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+                                    return Err(InnerError::InvalidInstanceId { referent: *value });
+                                }
+
+                                buf.push(curr_ref);
+                            } else {
+                                return type_mismatch(i, &rbx_value, "Ref");
+                            }
+                        }
+
+                        chunk.write_referent_array(buf.into_iter())?;
```
LPGhatguy, line 826: It doesn't look like we need to track `curr_ref` here anymore. Instead of creating a value, we should be able to just push directly into the buffer in both branches here.

Additionally, does referring to an invalid instance ID necessitate an error? If Roblox Studio saves a model that has a reference to an instance not included in the model, does a null referent get written, or does a referent get allocated for that value that is unused?

kennethloeffler: Looks like Studio does write out the null referent in that case, so yeah, the third branch here can never be hit. I've removed it.

LPGhatguy (approved):

PR #142, opened 2020-11-11
Implement SharedString in rbx-binary

Dekkonot:
This PR seeks to implement support for the `SSTR` chunk and the `SharedString` datatype in rbx-binary.

Because this isn't as trivial as the other datatypes, I thought having a PR to actually keep track of it would be good, so I'm opening this as a draft ahead of time.

As of me making this PR, we need to:

- [x] Implement SharedString in text_deserializer.rs
- [x] Implement SharedString in deserializer.rs
- [x] Implement SharedString in serializer.rs
- [x] Add actual tests


rbx_binary/src/deserializer.rs
```diff
@@ -263,6 +264,8 @@ struct BinaryDeserializer<R> {
     /// are interpreted by Roblox.
     metadata: HashMap<String, String>,
 
+    shared_strings: HashMap<u32, SharedString>,
```
LPGhatguy, line 267: This does not need to be a `HashMap`, we should be able to get away with this being a `Vec` instead. In general, any time we have a map of contiguous integer keys, we can use an array-like data structure instead of a map-like one.

Can we also add a comment like the other members of this struct describing what's in here? The key term here is probably "shared string dictionary".

rbx_binary/src/deserializer.rs
```diff
@@ -344,6 +348,29 @@ impl<R: Read> BinaryDeserializer<R> {
         Ok(())
     }
 
+    fn decode_sstr_chunk(&mut self, mut chunk: &[u8]) -> Result<(), InnerError> {
+        let version = chunk.read_le_u32()?;
+
+        if version != 0 {
+            return Err(InnerError::UnknownChunkVersion {
+                chunk_name: "PRNT",
```
LPGhatguy, line 356:
```suggestion
                chunk_name: "SSTR",
```

rbx_binary/src/deserializer.rs
```diff
@@ -344,6 +348,29 @@ impl<R: Read> BinaryDeserializer<R> {
...
+        }
+
+        let num_entries = chunk.read_le_u32()?;
+
+        for i in 0..num_entries {
+            chunk.read_exact(&mut [0; 16])?; // We don't do anything with the hash.
+
+            let data = chunk.read_binary_string()?;
+
+            self.shared_strings.insert(i, SharedString::new(data));
```
LPGhatguy, line 368: We should be able to get rid of the blank lines between statements in this loop. It's all pretty straight-line code.

rbx_binary/src/deserializer.rs
```diff
@@ -947,6 +974,37 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                        let shared_string = self.shared_strings.get(&value).ok_or_else(|| {
+                            InnerError::InvalidPropData {
+                                type_name: type_info.type_name.clone(),
+                                prop_name: prop_name.clone(),
+                                valid_value: "a valid SharedString",
+                                actual_value: format!("{:?}", value),
+                            }
+                        })?;
+
+                        let value: BinaryString = shared_string.data().into();
```
LPGhatguy, line 992: Why convert the value into a `BinaryString` here? We have support for `SharedString` in rbx_dom_weak, so we should use that here.

rbx_binary/src/text_deserializer.rs
```diff
@@ -84,6 +85,30 @@ fn decode_meta_chunk<R: Read>(mut reader: R) -> DecodedChunk {
...
+    let num_entries = reader.read_le_u32().unwrap();
+    let mut entries = Vec::with_capacity(num_entries as usize);
+
+    for _ in 0..num_entries {
+        let mut hash = [0; 16];
+        reader.read_exact(&mut hash).unwrap();
+
+        let data = reader.read_binary_string().unwrap();
+
+        entries.push(SharedString::new(data));
```
LPGhatguy, line 99: Similar to other code, we can compress these statements a bit since they aren't very complicated.
```suggestion
        let data = reader.read_binary_string().unwrap();
        entries.push(SharedString::new(data));
```

LPGhatguy (changes requested):

Dekkonot:
Accidentally includes the test file in that last PR, expect the CI to start failing from here on out. 🥴

rbx_binary/src/deserializer.rs
```diff
@@ -947,6 +974,39 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                                    valid_value: "a valid SharedString",
+                                    actual_value: format!("{:?}", value),
+                                }
+                            })?;
+
+                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
+
+                        instance.builder.add_property(
+                            &canonical_name,
+                            Variant::SharedString(shared_string.clone()),
```
LPGhatguy, line 997:
```suggestion
                            shared_string.clone(),
```

Should be able to avoid naming `Variant` here explicitly.

Dekkonot:
I'm not too sure about the format we're going with for serializing SharedString, but it's usable for testing and isn't noisy visually, so I think it's fine. Interested if anyone has opinions on it.

Keep in mind we only get one implementation of `Serializer` for SharedString, so it'll end up being global.

LPGhatguy:
I see that you cherrypicked the SharedString fix commit from master, but I think that this is just delaying your merge conflicts as those changes are now showing up as part of this PR too!

I would recommend backing up to b26b76b and then doing a normal merge (`git merge master`), which should let you resolve the conflicts fairly easily and commit them.

Dekkonot:
That was a good call. Was going to wait until the end (since a bunch of PRs were coming in at once) but I suppose we're probably mostly done for now.

Dekkonot:
Alright so as it turns out, serializing `SharedString` isn't a trivial task.

We have to write the `SSTR` chunk prior to writing `PROP` chunks, but as a result we don't actually know what `SharedString`s exist in the DOM up until then. The only solutions I can think of at the moment are to either change how we're deserializing so that `SSTR` can come after `PROP` chunks or to keep track of `SharedString`s somehow when they're added to the DOM.

Neither one is great, since the first one involves a second pass over the file (`rbx_xml` does this but XML trees and binary chunks are different stories) and the second involves either some overhead or an invasive change to `WeakDom` to keep track of `SharedString`s.

I think the second one is more practical long-term (SharedStrings being first-class citizens solves the problem for `rbx_xml` too) but it has some implications for the future of the project so I don't want to just *do it* without consulting someone first and discussing the pros and cons (it's also late and I don't want to do it tonight 😛)

LPGhatguy:
fwiw, `rbx_xml` doesn't take a second pass over the file while deserializing. It keeps [a list of rewrites to apply](https://github.com/rojo-rbx/rbx-dom/blob/91e543ca047ada31aced674b52318b33580e44ac/rbx_xml/src/deserializer.rs#L119-L131), which is used for both referents and shared strings. The second pass then just drains those arrays and applies the properties to the tree, using information that's now complete as the entire file has been traversed.

We should be able to apply the same technique here.

Dekkonot:
Right, sorry, I wasn't very thorough in investigating `rbx_xml`'s deserialization since it wasn't *super* important to this case. I just noticed that it looked to be doing that, so I thought it was worth mentioning above.

This is about serializing and not deserializing though, so it may have been misleading to bring up. That's my bad.

This is different than `rbx_xml` since we can't write the `SSTR` chunk after the fact; it has to be before `PROP` chunks that use it. We can probably get away with just keeping track of the various information and then writing any `PROP` chunks that use `SharedStrings` after the `SSTR` chunk, though I'm not sure if Studio will load a file that does that. We can give it a go though.

LPGhatguy:
I think I understand the problem now. I'll restate it to confirm. 😅 

We need to write the `SSTR` chunk before the shared strings contained within can be referenced by `PROP` chunks. However, we don't know _which shared strings we need to write_ until we've traversed all the properties.

I think we already have a good place to do this. It happens in a single step, traverses all instance properties, and happens before we emit any part of the file. We can extend [`BinarySerializer::collect_type_infos`](https://github.com/rojo-rbx/rbx-dom/blob/91e543ca047ada31aced674b52318b33580e44ac/rbx_binary/src/serializer.rs#L229-L347) to also collect any new `SharedString` values and assign them an ID.

Does that solution sound workable? I prefer solutions that generate files that are closer to what Roblox Studio outputs whenever we can, and this should let us be nearly identical.

Dekkonot:
Hah. I knew that deriving Hash would be a bad idea. I'll push a fix for that in a bit.

As for the build failure... It turns out that UnionOperations have both `formFactorRaw` and `FormFactor` serialized. Parts don't. That's... neat. I'm not sure if that's what's causing it to fail (I hope not) but it could be an issue. Feel like taking a look @LPGhatguy?

Dekkonot:
Alright upon further inspection it looks like it is actually that. We use different error text in `serialize_properties`, which is convenient, since it let me actually identify that it *isn't* that.

The only other place we throw `UnsupportedPropType` is in [`collect_type_info`](https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_binary/src/serializer.rs#L305).

Conveniently not related to this PR which is nice (I didn't touch that part of the code!), but it does need to be resolved before we can merge this, and it's really only through luck that it didn't come up until now.

But hey, at least we learned something about `UnionOperation` today!

rbx_binary/src/serializer.rs
```diff
@@ -413,6 +427,18 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
                 }
             }
+
+            if let Variant::SharedString(shared_string) = prop_value {
+                if !self.shared_string_ids.contains_key(shared_string) {
+                    let id = self
+                        .shared_strings
+                        .len()
+                        .try_into()
+                        .map_err(|_| InnerError::TooManySharedStrings)?;
```
LPGhatguy, line 437: I don't think this will ever get anywhere near overflowing in practice. It should be okay to just use `as u32` here, or even `try_into().unwrap()`.

rbx_binary/src/serializer.rs
```diff
@@ -452,6 +478,31 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
+        }
+
+        let mut chunk = ChunkBuilder::new(b"SSTR", ChunkCompression::Compressed);
+
+        chunk.write_le_u32(0)?; // SSTR version number
+        chunk.write_le_u32(self.shared_strings.len() as u32)?;
+
+        for shared_string in &self.shared_strings {
+            // Better to write nothing than write half a hash
+            chunk.write_all(b"\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0")?;
```
LPGhatguy, line 497:
```suggestion
            chunk.write_all(&[0; 16])?;
```

rbx_binary/src/text_deserializer.rs
```diff
@@ -649,6 +690,25 @@ pub enum DecodedChunk {
...
+        let hash = value.hash();
+        let mut hash_hex = String::with_capacity(hash.as_bytes().len() * 2);
+
+        for byte in hash.as_bytes() {
+            write!(hash_hex, "{:02x}", byte).unwrap();
+        }
+        processed.push(format!("[{} bytes] {}", value.data().len(), hash_hex))
+    }
+
+    serializer.collect_seq(processed)
```
LPGhatguy, line 709: Using `collect_seq` here when we aren't generating an iterator is pretty weird. You should use [`serialize_seq`](https://docs.rs/serde/1.0.117/serde/trait.Serializer.html#tymethod.serialize_seq) instead, then call the methods on its result. This lets us avoid needing to allocate a `Vec` here needlessly.

We can also reuse the `String` between loop iterations. Allocate it at the beginning of this function, `clear()` it at the beginning of the loop, and then write into it and use `serialize_element` on it.

LPGhatguy: Another question. Can we serialize these values the same way that we serialize them for `DomViewer`? We should be able to make up a little struct that derives `Serialize` to use here:

```rust
#[derive(Serialize)]
struct SerializedSharedString<'a> {
    len: usize,
    hash: &'a str,
}
```

Then we can create one, serialize it into the sequence, and happy re-use the `String` it referred to for the next value.

rbx_binary/src/serializer.rs
```diff
@@ -977,6 +1028,23 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
...
                     }
+                    Type::SharedString => {
+                        let mut entries = Vec::with_capacity(values.len());
+
+                        for (i, rbx_value) in values {
+                            if let Variant::SharedString(value) = rbx_value.as_ref() {
+                                let id = self
+                                    .shared_string_ids
+                                    .get(value)
+                                    .ok_or_else(|| InnerError::TooManySharedStrings {})?;
```
LPGhatguy, line 1039: Same comment as other spot, we shouldn't be worried about overflowing a u32 here.

LPGhatguy (changes requested):
Looking pretty good! Mostly comments about Rust-isms.

LPGhatguy (approved):

rbx_binary/src/serializer.rs
```diff
@@ -977,6 +1021,20 @@ impl<'a, W: Write> BinarySerializer<'a, W> {
 
                         chunk.write_interleaved_i64_array(buf.into_iter())?;
                     }
+                    Type::SharedString => {
+                        let mut entries = Vec::with_capacity(values.len());
+
+                        for (i, rbx_value) in values {
+                            if let Variant::SharedString(value) = rbx_value.as_ref() {
+                                let id = &self.shared_string_ids.get[value];
```
LPGhatguy, line 1029:
```suggestion
                                let id = &self.shared_string_ids[value];
```

oops

PR #151, opened 2020-11-16
Break type_infos member into new struct, TypeInfos

LPGhatguy:
This PR collects together two members from `BinarySerializer` into a new struct. It's intended to help overcome some of the aliasing borrow issues encountered in #142.

This is a common technique in Rust: when a method borrows too much of a struct and it starts being prohibitive, we break out the borrowed parts into a new struct. Usually, this isn't so bad, like in this case. In some gnarlier cases, it can help to make the function in question a free function that takes many independent references.


Dekkonot (approved):
Seems to have solved the problem. Thanks!

PR #156, opened 2020-12-02
[rbx_dom_lua] Support Axes, Faces, and Ray, improve Rect support.

CloneTrooper1019:
This patch implements the remaining datatypes that are missing from rbx_dom_lua, and fixes problems with encoding the Rect type.

### DIFF
--- a/rbx_dom_lua/src/EncodedValue.lua
+++ b/rbx_dom_lua/src/EncodedValue.lua
@@ -20,6 +20,41 @@ local function serializeFloat(value)
 	return value
 end
 
+local function enumsEncoder(enum)
+	local items = enum:GetEnumItems()
+
+	return function(flags)
+		local mask = 0
+		
+		for i, item in ipairs(items) do
+			if flags[item.Name] then
+				mask += (2 ^ item.Value)
+			end
+		end
+		
+		return mask
+	end
+end
+
+local function enumsDecoder(constructor, enum)
+	local decode = unpackDecoder(constructor)
+	local items = enum:GetEnumItems()
+
+	return function(mask)
+		local set = {}
+
+		for _,item in pairs(items) do
+			local bit = (2 ^ item.Value)
+			
+			if bit32.btest(bit, mask) then
+				table.insert(set, item)
+			end
+		end
+		
+		return decode(set)
+	end
+end
+
 local encoders
 encoders = {
 	Bool = identity,
@@ -29,6 +64,9 @@ encoders = {
 	Int32 = identity,
 	Int64 = identity,
 	String = identity,
+
+	Axes = enumsEncoder(Enum.Axis),
+	Faces = enumsEncoder(Enum.NormalId),
 
 	BinaryString = base64.encode,
 	SharedString = base64.encode,
@@ -77,8 +115,8 @@ encoders = {
 	end,
 	Rect = function(value)
 		return {
-			Min = {value.Min.X, value.Min.Y},
-			Max = {value.Max.X, value.Max.Y},
+			Min = encoders.Vector2(value.Min),
+			Max = encoders.Vector2(value.Max),
 		}
 	end,
 	UDim = function(value)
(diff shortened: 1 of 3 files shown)


rbx_dom_lua/src/EncodedValue.spec.lua
```diff
@@ -47,6 +47,36 @@ return function()
...
+			Value = 0b_111,
+		}
+
+		local ok, decoded = EncodedValue.decode(input)
+		assert(ok, decoded)
+
+		for _,axis in pairs(Enum.Axis:GetEnumItems()) do
+			local set = decoded[axis.Name]
+			expect(set).to.equal(true)
+		end
```
LPGhatguy, line 77: Can we check the axes individually here? Using a loop here when there are three items feels kinda funny.

If possible, we should also expand the set of tests a little bit here. Try all the axes, none of the axes, and maybe one axis.

rbx_dom_lua/src/EncodedValue.spec.lua
```diff
@@ -47,6 +47,36 @@ return function()
 		expect(decoded).to.equal(output)
 	end)
 
+	it("should decode a 'Faces' bit-mask value", function()
+		local input = {
+			Type = "Faces",
+			Value = 0b_111111,
```
LPGhatguy, line 53:
```suggestion
			Value = 0b111111,
```

We don't need these optional spacer characters.

rbx_dom_lua/src/EncodedValue.spec.lua
```diff
@@ -47,6 +47,36 @@ return function()
...
+		for _,normalId in pairs(Enum.NormalId:GetEnumItems()) do
+			local set = decoded[normalId.Name]
+			expect(set).to.equal(true)
+		end
+	end)
+
+	it("should decode an 'Axes' bit-mask value", function()
+		local input = {
+			Type = "Axes",
+			Value = 0b_111,
```
LPGhatguy, line 68:
```suggestion
			Value = 0b111,
```

rbx_dom_lua/src/EncodedValue.lua
```diff
@@ -20,6 +20,41 @@ local function serializeFloat(value)
 	return value
 end
 
+local function enumsEncoder(enum)
+	local items = enum:GetEnumItems()
+
+	return function(flags)
+		local mask = 0
+		
+		for i, item in ipairs(items) do
```
LPGhatguy, line 29:
```suggestion
		for _, item in ipairs(items) do
```

rbx_dom_lua/src/EncodedValue.lua
```diff
@@ -20,6 +20,41 @@ local function serializeFloat(value)
...
+end
+
+local function enumsDecoder(constructor, enum)
+	local decode = unpackDecoder(constructor)
+	local items = enum:GetEnumItems()
+
+	return function(mask)
+		local set = {}
+
+		for _,item in pairs(items) do
```
LPGhatguy, line 46:
```suggestion
		for _, item in ipairs(items) do
```

rbx_dom_lua/src/EncodedValue.lua
```diff
@@ -20,6 +20,41 @@ local function serializeFloat(value)
 	return value
 end
 
+local function enumsEncoder(enum)
```
LPGhatguy, line 23: Can you write comments describing why these functions do what they do? Many readers won't be familiar with why bitwise operations would be involved in dealing with axes and faces.

LPGhatguy (changes requested):

rbx_dom_lua/src/EncodedValue.spec.lua
CloneTrooper1019: That’s fine. It was more of a personal readability choice and I wasn’t sure if there was any consensus there in the style guide.

rbx_dom_lua/src/EncodedValue.lua
CloneTrooper1019: Done! Let me know if it needs any adjustments.

rbx_dom_lua/src/EncodedValue.spec.lua
CloneTrooper1019: I adjusted the code for the original Axes test as requested. I also added an additional mixed bit-input test for both Axes and Faces to ensure the stability of this encoding technique.

CloneTrooper1019:
Didn’t realize I forgot to request another review, whoops.

LPGhatguy (approved):

PR #166, opened 2021-03-05
Getting and setting of attributes

Kampfkarren:
- [x] get_attributes
- [ ] attributes_from_map

### DIFF
--- a/rbx_types/Cargo.toml
+++ b/rbx_types/Cargo.toml
@@ -17,10 +17,12 @@ bitflags = "1.2.1"
 lazy_static = "1.4.0"
 rand = "0.7.3"
 serde = { version = "1.0.104", features = ["derive"], optional = true }
+thiserror = "1.0.16"
 rbx_dom_weak = { version = "1.10.1", optional = true }
 
 [dev-dependencies]
 bincode = "1.2.1"
+insta = "0.16.0"
 serde_json = "1.0.45"
 
 [features]
--- a/rbx_types/src/attributes.rs
+++ b/rbx_types/src/attributes.rs
@@ -1,0 +1,293 @@
+// Thanks to Anaminus! https://github.com/RobloxAPI/rbxattr/blob/master/spec.md
+
+use crate::{basic_types::*, brick_color::BrickColor, variant::Variant};
+use std::{
+    collections::HashMap,
+    convert::TryFrom,
+    io::{self, Read},
+};
+use thiserror::Error;
+
+macro_rules! create_attribute_type {
+    ({
+        $(
+            $key:ident = $number:tt,
+        )+
+    }) => {
+        #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+        enum AttributeType {
+            $(
+                $key = $number,
+            )+
+        }
+
+        impl TryFrom<u8> for AttributeType {
+            type Error = AttributeDeserializeError;
+
+            fn try_from(byte: u8) -> Result<Self, Self::Error> {
+                match byte {
+                    $(
+                        $number => Ok(Self::$key),
+                    )+
+
+                    other => Err(AttributeDeserializeError::InvalidValueType(other))
+                }
+            }
+        }
+    };
+}
+
+create_attribute_type!({
+    String = 0x02,
+    Bool = 0x03,
+    Float = 0x05,
+    Double = 0x06,
+    UDim = 0x09,
+    UDim2 = 0x0A,
+    BrickColor = 0x0E,
+    Color3 = 0x0F,
+    Vector2 = 0x10,
+    Vector3 = 0x11,
+    NumberSequence = 0x17,
+    ColorSequence = 0x19,
+    NumberRange = 0x1B,
+    Rect = 0x1C,
+});
+
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+pub enum AttributeDeserializeError {
+    #[error("invalid value type: {0}")]
+    InvalidValueType(u8),
+
+    #[error("invalid brick color: {0}")]
+    InvalidBrickColor(u32),
+
+    #[error("invalid entry key")]
+    InvalidEntryKey,
+
+    #[error("invalid name")]
+    InvalidName,
+
+    #[error("invalid size")]
+    InvalidSize,
+
+    #[error("no value type was found")]
+    NoValueType,
+
+    #[error("couldn't read bytes to deserialize {0}")]
+    Other(&'static str),
+}
+
+fn read_u8<R: Read>(mut reader: R) -> io::Result<u8> {
+    let mut bytes = [0u8; 1];
+    reader.read_exact(&mut bytes)?;
+    Ok(bytes[0])
+}
+
+fn read_i32<R: Read>(mut reader: R) -> io::Result<i32> {
+    let mut bytes = [0u8; 4];
+    reader.read_exact(&mut bytes)?;
+    Ok(i32::from_le_bytes(bytes))
+}
+
+fn read_u32<R: Read>(mut reader: R) -> io::Result<u32> {
+    let mut bytes = [0u8; 4];
+    reader.read_exact(&mut bytes)?;
+    Ok(u32::from_le_bytes(bytes))
+}
+
+fn read_f32<R: Read>(mut reader: R) -> io::Result<f32> {
+    let mut bytes = [0u8; 4];
+    reader.read_exact(&mut bytes)?;
+    Ok(f32::from_le_bytes(bytes))
+}
+
+fn read_string<R: Read>(mut reader: R) -> io::Result<String> {
+    let size = read_u32(&mut reader)? as usize;
+    let mut characters = vec![0u8; size];
+    reader.read_exact(&mut characters)?;
+    Ok(characters.into_iter().map(|byte| byte as char).collect())
+}
+
+fn read_color3<R: Read>(mut reader: R) -> io::Result<Color3> {
+    Ok(Color3::new(
+        read_f32(&mut reader)?,
+        read_f32(&mut reader)?,
+        read_f32(&mut reader)?,
+    ))
+}
+
+fn read_udim<R: Read>(mut reader: R) -> io::Result<UDim> {
+    Ok(UDim::new(read_f32(&mut reader)?, read_i32(&mut reader)?))
+}
+
+fn read_vector2<R: Read>(mut reader: R) -> io::Result<Vector2> {
+    Ok(Vector2::new(read_f32(&mut reader)?, read_f32(&mut reader)?))
+}
+
+/// Reads through an attribute property (AttributesSerialize) and returns a map of attribute names -> values.
+pub fn get_attributes<R: Read>(
+    mut value: R,
+) -> Result<HashMap<String, Variant>, AttributeDeserializeError> {
+    let size = read_u32(&mut value).map_err(|_| AttributeDeserializeError::InvalidSize)?;
+    let mut attributes = HashMap::with_capacity(size as usize);
+
+    for _ in 0..size {
+        let name = read_string(&mut value).map_err(|_| AttributeDeserializeError::InvalidName)?;
+
+        let attribute_key = AttributeType::try_from(
+            read_u8(&mut value).map_err(|_| AttributeDeserializeError::NoValueType)?,
+        )?;
+
+        let value = match attribute_key {
+            AttributeType::BrickColor => {
+                let color = read_u32(&mut value)
+                    .map_err(|_| AttributeDeserializeError::Other("BrickColor"))?;
+
+                Variant::BrickColor(
+                    BrickColor::from_number(color as u16)
+                        .ok_or_else(|| AttributeDeserializeError::InvalidBrickColor(color))?,
+                )
+            }
+
+            AttributeType::Bool => Variant::Bool(
+                read_u8(&mut value).map_err(|_| AttributeDeserializeError::Other("bool"))? != 0,
+            ),
+
+            AttributeType::Color3 => Variant::Color3(
+                read_color3(&mut value).map_err(|_| AttributeDeserializeError::Other("Color3"))?,
+            ),
+
+            AttributeType::ColorSequence => {
+                let size =
+                    read_u32(&mut value).map_err(|_| AttributeDeserializeError::InvalidSize)?;
+                let mut keypoints = Vec::with_capacity(size as usize);
+
+                for _ in 0..size {
+                    // `envelope` is not represented in rbx_types, apparently it's always zero.
+                    let _envelope = read_f32(&mut value).map_err(|_| {
+                        AttributeDeserializeError::Other("ColorSequenceKeypoint envelope")
+                    })?;
+
+                    let time = read_f32(&mut value).map_err(|_| {
+                        AttributeDeserializeError::Other("ColorSequenceKeypoint time")
+                    })?;
+
+                    let color = read_color3(&mut value).map_err(|_| {
+                        AttributeDeserializeError::Other("ColorSequenceKeypoint color")
+                    })?;
+
+                    keypoints.push(ColorSequenceKeypoint::new(time, color));
+                }
+
+                Variant::ColorSequence(ColorSequence { keypoints })
+            }
+
+            AttributeType::Double => {
+                let mut bytes = [0u8; 8];
+                value
+                    .read_exact(&mut bytes)
+                    .map_err(|_| AttributeDeserializeError::Other("double"))?;
+                Variant::Float64(f64::from_le_bytes(bytes))
+            }
+
+            AttributeType::Float => Variant::Float32(
+                read_f32(&mut value).map_err(|_| AttributeDeserializeError::Other("falot"))?,
+            ),
+
+            AttributeType::NumberRange => Variant::NumberRange(NumberRange::new(
+                read_f32(&mut value)
+                    .map_err(|_| AttributeDeserializeError::Other("NumberRange min"))?,
+                read_f32(&mut value)
+                    .map_err(|_| AttributeDeserializeError::Other("NumberRange max"))?,
+            )),
+
+            AttributeType::NumberSequence => {
+                let size =
+                    read_u32(&mut value).map_err(|_| AttributeDeserializeError::InvalidSize)?;
+
+                let mut keypoints = Vec::with_capacity(size as usize);
+
+                for _ in 0..size {
+                    let envelope = read_f32(&mut value)
+                        .map_err(|_| AttributeDeserializeError::Other("NumberSequence envelope"))?;
+
+                    let time = read_f32(&mut value)
+                        .map_err(|_| AttributeDeserializeError::Other("NumberSequence time"))?;
+
+                    let value = read_f32(&mut value)
+                        .map_err(|_| AttributeDeserializeError::Other("NumberSequence value"))?;
+
+                    keypoints.push(NumberSequenceKeypoint::new(time, value, envelope));
+                }
+
+                Variant::NumberSequence(NumberSequence { keypoints })
+            }
+
+            AttributeType::Rect => Variant::Rect(Rect::new(
+                read_vector2(&mut value)
+                    .map_err(|_| AttributeDeserializeError::Other("Rect min"))?,
+                read_vector2(&mut value)
+                    .map_err(|_| AttributeDeserializeError::Other("Rect max"))?,
+            )),
+
+            AttributeType::String => Variant::String(
+                read_string(&mut value).map_err(|_| AttributeDeserializeError::Other("string"))?,
+            ),
+
+            AttributeType::UDim => Variant::UDim(
+                read_udim(&mut value).map_err(|_| AttributeDeserializeError::Other("UDim"))?,
+            ),
+
+            AttributeType::UDim2 => Variant::UDim2(UDim2::new(
+                read_udim(&mut value).map_err(|_| AttributeDeserializeError::Other("UDim2 X"))?,
+                read_udim(&mut value).map_err(|_| AttributeDeserializeError::Other("UDim2 Y"))?,
+            )),
+
+            AttributeType::Vector2 => Variant::Vector2(Vector2::new(
+                read_f32(&mut value).map_err(|_| AttributeDeserializeError::Other("Vector2 X"))?,
+                read_f32(&mut value).map_err(|_| AttributeDeserializeError::Other("Vector2 Y"))?,
+            )),
+
+            AttributeType::Vector3 => Variant::Vector3(Vector3::new(
+                read_f32(&mut value).map_err(|_| AttributeDeserializeError::Other("Vector3 X"))?,
+                read_f32(&mut value).map_err(|_| AttributeDeserializeError::Other("Vector3 Y"))?,
+                read_f32(&mut value).map_err(|_| AttributeDeserializeError::Other("Vector3 Z"))?,
+            )),
+        };
+
+        attributes.insert(name, value);
+    }
+
+    Ok(attributes)
+}
+
+#[cfg(all(feature = "serde", test))]
+mod tests {
+    use super::*;
+
+    // This is taken from rbx-test-files/models/attributes/xml.rbxmx.
+    // This is pasted raw as to not create a circular dependency in test (rbx_types -> rbx_xml/rbx_binary -> rbx_types)
+    const ATTRIBUTES_BASE64: &'static str = r"
+    DwAAAAMAAABOYU4GAAAAAAAA+P8IAAAASW5maW5pdHkGAAAAAAAA8H8NAAAAQ29sb3JTZXF1
+    ZW5jZRkDAAAAAAAAAAAAAAAAAIA/AAAAAAAAAAAAAAAAAAAAPwAAAAAAAIA/AAAAAAAAAAAA
+    AIA/AAAAAAAAAAAAAIA/BwAAAFZlY3RvcjMRAACAPwAAAEAAAEBABwAAAFZlY3RvcjIQAAAg
+    QQAASEIOAAAATnVtYmVyU2VxdWVuY2UXAwAAAAAAAAAAAAAAAACAPwAAAAAAAAA/AAAAAAAA
+    AAAAAIA/AACAPwYAAABDb2xvcjMPo6IiPwAAAAAAAIA/CgAAAEJyaWNrQ29sb3IO7AMAAAQA
+    AABSZWN0HAAAgD8AAABAAABAQAAAgEAFAAAAVURpbTIKAAAAPwoAAAAzMzM/HgAAAAQAAABV
+    RGltCQAAAD9kAAAACwAAAE51bWJlclJhbmdlGwAAoEAAACBBBgAAAE51bWJlcgYAAAAAgBzI
+    QAcAAABCb29sZWFuAwEGAAAAU3RyaW5nAg0AAABIZWxsbywgd29ybGQh
+    ";
+
+    #[test]
+    fn test_attributes() {
+        let attributes_value =
+            base64::decode(&ATTRIBUTES_BASE64.split_whitespace().collect::<String>())
+                .expect("bad base64 for attributes");
+
+        let attributes =
+            get_attributes(&attributes_value[..]).expect("couldn't deserialize attributes");
+
+        insta::assert_yaml_snapshot!(attributes);
+    }
+}
--- a/rbx_types/src/lib.rs
+++ b/rbx_types/src/lib.rs
@@ -5,6 +5,7 @@ mod serde_util;
 #[cfg(feature = "legacy-compat")]
 mod legacy_compat;
 
+mod attributes;
 mod axes;
 mod basic_types;
 mod binary_string;
@@ -17,6 +18,7 @@ mod referent;
 mod shared_string;
 mod variant;
 
+pub use attributes::*;
 pub use axes::*;
 pub use basic_types::*;
 pub use binary_string::*;
--- a/rbx_types/src/snapshots/rbx_types__attributes__tests__attributes.snap
+++ b/rbx_types/src/snapshots/rbx_types__attributes__tests__attributes.snap
@@ -1,0 +1,95 @@
+---
+source: rbx_types/src/attributes.rs
+expression: attributes
+---
+NumberSequence:
+  Type: NumberSequence
+  Value:
+    Keypoints:
+      - Time: 0.0
+        Value: 1.0
+        Envelope: 0.0
+      - Time: 0.5
+        Value: 0.0
+        Envelope: 0.0
+      - Time: 1.0
+        Value: 1.0
+        Envelope: 0.0
+"NaN":
+  Type: Float64
+  Value: .nan
+ColorSequence:
+  Type: ColorSequence
+  Value:
+    Keypoints:
+      - Time: 0.0
+        Color:
+          - 1.0
+          - 0.0
+          - 0.0
+      - Time: 0.5
+        Color:
+          - 0.0
+          - 1.0
+          - 0.0
+      - Time: 1.0
+        Color:
+          - 0.0
+          - 0.0
+          - 1.0
+Color3:
+  Type: Color3
+  Value:
+    - 0.6352941393852234
+    - 0.0
+    - 1.0
+Rect:
+  Type: Rect
+  Value:
+    - - 1.0
+      - 2.0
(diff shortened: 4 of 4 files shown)


Kampfkarren:
I have no idea why the tests are failing.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,121 @@
...
+    write_f32(bytes, vector2.y);
+}
+
+/// Writes the attribute property (AttributesSerialize) from a map of attribute names -> values.
+pub fn attributes_from_map<
+    K: Into<String>,
+    V: Into<Variant>,
+    I: Iterator<Item = (K, V)> + ExactSizeIterator,
+    M: IntoIterator<IntoIter = I, Item = (K, V)>,
+>(
```
Kampfkarren, line 41: I'm iffy about this type signature, would like a second glance at it.

I want it to capture HashMaps, Vec<(K, V)>, and stuff like BTreeMap.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    #[error("couldn't read bytes to deserialize {0}")]
+    Other(&'static str),
+}
+
+#[cfg(all(feature = "serde", test))]
+mod tests {
+    use super::*;
+
+    // This is taken from rbx-test-files/models/attributes/xml.rbxmx.
+    // This is pasted raw as to not create a circular dependency in test (rbx_types -> rbx_xml/rbx_binary -> rbx_types)
```
Dekkonot, line 103: What do you mean by that? rbx-test-files is stored as a submodule in the root of the workspace, which is how rbx_binary uses it. Is there a reason you can't pull from there (other than the test file not being merged in yet)?

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let new_attribute_bytes = attributes_from_map(attributes_stable_order.to_owned())
+            .expect("couldn't get attributes from map");
+        let new_attributes = get_attributes(&new_attribute_bytes[..])
+            .expect("couldn't deserialize crate produced binary");
+        let mut new_attributes_stable_order = new_attributes.into_iter().collect::<Vec<_>>();
+        new_attributes_stable_order.sort_by_cached_key(|(key, _)| key.to_owned());
+
+        // They are not checked directly against each other because the data contains NaN.
+        assert_eq!(
```
Dekkonot, line 136: We should probably be using insta snapshots here instead of just manually asserting that these are equal.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,180 @@
...
+    Ok(u32::from_le_bytes(bytes))
+}
+
+fn read_f32<R: Read>(mut reader: R) -> io::Result<f32> {
+    let mut bytes = [0u8; 4];
+    reader.read_exact(&mut bytes)?;
+    Ok(f32::from_le_bytes(bytes))
+}
+
+fn read_string<R: Read>(mut reader: R) -> io::Result<String> {
```
Dekkonot, line 34: Given that strings in Lua don't necessarily have to be UTF-8, and Attributes support binary data (which won't be valid UTF-8), it seems like it might be pertinent to return the Vec instead of turning it into a String.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,180 @@
...
+                read_color3(&mut value).map_err(|_| AttributeError::Other("Color3"))?,
+            ),
+
+            AttributeType::ColorSequence => {
+                let size = read_u32(&mut value).map_err(|_| AttributeError::InvalidSize)?;
+                let mut keypoints = Vec::with_capacity(size as usize);
+
+                for _ in 0..size {
+                    // `envelope` is not represented in rbx_types, apparently it's always zero.
+                    let _envelope = read_f32(&mut value)
```
Dekkonot, line 93: Given that this is basically a no-op, we should probably make it read 4 bytes from `value` instead of reading it using `read_f32`.

Not sure if Rust will kill this during compilation but it makes me uncomfortable to have code that does nothing.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,180 @@
...
+                        .map_err(|_| AttributeError::Other("ColorSequenceKeypoint color"))?;
+
+                    keypoints.push(ColorSequenceKeypoint::new(time, color));
+                }
+
+                Variant::ColorSequence(ColorSequence { keypoints })
+            }
+
+            AttributeType::Float32 => {
+                Variant::Float32(read_f32(&mut value).map_err(|_| AttributeError::Other("falot"))?)
```
Dekkonot, line 109: `falot` -> `float`

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,121 @@
+use super::*;
+use crate::{basic_types::*, variant::Variant};
+
+fn write_f32(bytes: &mut Vec<u8>, n: f32) {
+    bytes.append(&mut n.to_le_bytes().to_vec());
```
Dekkonot, line 5: We should use `write_all` here instead like we do in rbx_binary. It replaces this with `bytes.write_all(&n.to_le_bytes()).unwrap()`. `write_all` shouldn't ever fail in this case (it will only fail if the Vec is too large afaik) so `unwrap` should be safe to use.

See `write_u32` as well.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,121 @@
...
+    V: Into<Variant>,
+    I: Iterator<Item = (K, V)> + ExactSizeIterator,
+    M: IntoIterator<IntoIter = I, Item = (K, V)>,
+>(
+    map: M,
+) -> Result<Vec<u8>, AttributeError> {
+    let map = map.into_iter();
+    let mut bytes = Vec::new();
+
+    bytes.extend((map.len() as u32).to_le_bytes().iter());
```
Dekkonot, line 47: We can probably use `bytes.write_all` as mentioned previously.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// Reads through an attribute property (AttributesSerialize) and returns a map of attribute names -> values.
+pub fn get_attributes<R: Read>(mut value: R) -> Result<HashMap<String, Variant>, AttributeError> {
+    let size = read_u32(&mut value).map_err(|_| AttributeError::InvalidSize)?;
+    let mut attributes = HashMap::with_capacity(size as usize);
+
+    for _ in 0..size {
+        let name = read_string(&mut value).map_err(|_| AttributeError::InvalidName)?;
+
+        let attribute_key =
```
Dekkonot, line 65: We can probably just cast directly into the Enum using `try_into` like we do in rbx_binary:
https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_binary/src/deserializer.rs#L419

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,121 @@
...
+    for (name, variant) in map {
+        let variant = variant.into();
+
+        write_string(&mut bytes, &name.into());
+
+        let attribute_type = AttributeType::try_from(variant.ty())?;
```
Dekkonot, line 53: We can try directly casting to the Enum here too

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,121 @@
...
+                    write_f32(&mut bytes, 0.0); // Envelope
+                    write_f32(&mut bytes, keypoint.time);
+                    write_color3(&mut bytes, keypoint.color);
+                }
+            }
+            (AttributeType::Float32, Variant::Float32(float)) => {
+                write_f32(&mut bytes, float);
+            }
+            (AttributeType::Float64, Variant::Float64(float)) => {
+                bytes.append(&mut float.to_le_bytes().to_vec());
```
Dekkonot, line 75: Another case where `write_all` might be more appropriate

Dekkonot (changes requested):
Generally, my thoughts are that we should do this more like rbx_binary so that there's consistency in the code base.

rbx_types/src/attributes/mod.rs
Kampfkarren: The dependency isn't with rbx-test-files, I can include that easily. The problem is with then parsing that data, actually getting the AttributesSerialize *out* of the file. This is what rbx_binary and rbx_xml are for, but then we create a circular dependency (which was my first try, but then things like Variant were mismatched because the ones rbx_binary was giving me were from its own crate).

I *could* potentially just write an XML parser just for this, but it seems a bit unnecessary.

rbx_types/src/attributes/mod.rs
Kampfkarren: This would be an identical snapshot with `attributes_stable_order` in a working case, which is already in its own snapshot. The `assert_eq!` is there because that way, it's impossible to flub it and have two separate snaps.

rbx_types/src/attributes/reader.rs
Kampfkarren: I think the explicit way this is written is good, but I can change it if thats standard.

rbx_types/src/attributes/mod.rs
Dekkonot: Right I didn't think about that. Sorry, carry on!

rbx_types/src/attributes/reader.rs
Kampfkarren: Turns out this leads into a weird issue--all the AttributeTypes are `AttributeType::T => Variant::T`, it has made a lot of the code surrounding it easier to work with. In this case, `AttributeType::String => Variant::String`. Would you recommend making this BinaryString or similar?

rbx_types/src/attributes/reader.rs
Kampfkarren: Enum as in the u8? That's what it looks like in your example.

Dekkonot: Upon a closer look at the implementation in rbx_binary, it seems as though I made an assumption I shouldn't have. In my defense, the code was written before I touched the codebase, but I have no one to blame but myself for not reading it.

I was mostly hoping to reduce some boilerplate but it seems to just be a hidden abstraction in rbx_binary. 🙁

rbx_types/src/attributes/reader.rs
Dekkonot: Up to preference I guess. It probably won't matter performance-wise, and if it does we can always remove it. This is a living project, after all.

Dekkonot: To clarify, it looks like we don't have a set style for dealing with always-null bytes in rbx_binary, so it doesn't matter for now. The implementation for ColorSequence reads a float32, so we should probably keep this as-is.

rbx_types/src/attributes/reader.rs
Dekkonot: That's an unfortunate problem. It isn't an ideal solution but I'm going to hesitantly say "yes" to making it BinaryString since supporting binary data should be a priority of ours. Worst-case we can always change it later.

rbx_types/src/attributes/mod.rs
Dekkonot: Seems reasonable.

Dekkonot (approved):
Hesitantly approving this though I'd like @LPGhatguy to give it a once over and perhaps give their thoughts on using Vec\<u8\> over String.

Kampfkarren:
For posterity--it was a decision of correctness vs usefulness, and correctness one. In reality, Roblox should not be allowing weird strings to go through, but it would be incorrect to assert through types that this is impossible.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+mod reader;
+mod writer;
+
+pub use reader::get_attributes;
+pub use writer::attributes_from_map;
+
+macro_rules! create_attribute_type {
+    ({
+        $(
+            $key:ident = $number:tt,
```
LPGhatguy, line 14:
```suggestion
            $key:ident = $number:literal,
```

We should prefer to use a more specific matcher than `tt` when possible. Here, `expr` or `literal` should fit well.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
+use crate::variant::VariantType;
+use std::convert::TryFrom;
+use thiserror::Error;
+
+mod reader;
+mod writer;
+
+pub use reader::get_attributes;
+pub use writer::attributes_from_map;
```
LPGhatguy, line 9:
```suggestion
use std::convert::TryFrom;

use thiserror::Error;

use crate::variant::VariantType;

mod reader;
mod writer;

pub use reader::get_attributes;
pub use writer::attributes_from_map;
```

We should prefer to keep our imports sorted into sections based on where they're from.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    UDim2 = 0x0A,
+    BrickColor = 0x0E,
+    Color3 = 0x0F,
+    Vector2 = 0x10,
+    Vector3 = 0x11,
+    NumberSequence = 0x17,
+    ColorSequence = 0x19,
+    NumberRange = 0x1B,
+    Rect = 0x1C,
+});
```
LPGhatguy, line 69: What's the source of the holes here? Are these types we haven't figured out or types that we don't support?

Documentation here in either case would be great.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+pub enum AttributeError {
```
LPGhatguy, line 72: I think that public error enums are usually a trap as it hurts our ability to refactor. Usually, consumers don't need to match on these. We should make this type `pub(crate)` at most and expose a top-level error type that's just an opaque error struct.

`thiserror` supports this pattern pretty well.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    AAAAAIA/AACAPwYAAABDb2xvcjMPo6IiPwAAAAAAAIA/CgAAAEJyaWNrQ29sb3IO7AMAAAQA
+    AABSZWN0HAAAgD8AAABAAABAQAAAgEAFAAAAVURpbTIKAAAAPwoAAAAzMzM/HgAAAAQAAABV
+    RGltCQAAAD9kAAAACwAAAE51bWJlclJhbmdlGwAAoEAAACBBBgAAAE51bWJlcgYAAAAAgBzI
+    QAcAAABCb29sZWFuAwEGAAAAU3RyaW5nAg0AAABIZWxsbywgd29ybGQh
+    ";
+
+    #[test]
+    fn test_attributes() {
+        let attributes_value =
+            base64::decode(&ATTRIBUTES_BASE64.split_whitespace().collect::<String>())
```
LPGhatguy, line 118: Can we change the literal to instead just not include whitespace? You can end a line with `\` to make Rust ignore all whitespace after it. See this playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e94130ddf01f1c03bec42c84fc49f9dd

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+                .expect("bad base64 for attributes");
+
+        let attributes =
+            get_attributes(&attributes_value[..]).expect("couldn't deserialize attributes");
+
+        let mut attributes_stable_order = attributes.into_iter().collect::<Vec<_>>();
+        attributes_stable_order.sort_by_cached_key(|(key, _)| key.to_owned());
+        insta::assert_yaml_snapshot!(attributes_stable_order);
```
LPGhatguy, line 126: We should collect into a `BTreeMap` instead of collecting into a `Vec` and then sorting; should result in a better snapshot too.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let new_attribute_bytes = attributes_from_map(attributes_stable_order.to_owned())
+            .expect("couldn't get attributes from map");
+        let new_attributes = get_attributes(&new_attribute_bytes[..])
+            .expect("couldn't deserialize crate produced binary");
+        let mut new_attributes_stable_order = new_attributes.into_iter().collect::<Vec<_>>();
+        new_attributes_stable_order.sort_by_cached_key(|(key, _)| key.to_owned());
```
LPGhatguy, line 133: Same here, should just use a `BTreeMap`.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        // They are not checked directly against each other because the data contains NaN.
+        assert_eq!(
+            format!("{:#?}", attributes_stable_order),
+            format!("{:#?}", new_attributes_stable_order)
+        );
```
LPGhatguy, line 139: Maybe it'd be a good idea to split the NaN off into its own test; relying on formatting machinery for equality in the general case is not something I'm comfortable with.

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let attributes =
+            get_attributes(&attributes_value[..]).expect("couldn't deserialize attributes");
+
+        let mut attributes_stable_order = attributes.into_iter().collect::<Vec<_>>();
+        attributes_stable_order.sort_by_cached_key(|(key, _)| key.to_owned());
+        insta::assert_yaml_snapshot!(attributes_stable_order);
+
+        let new_attribute_bytes = attributes_from_map(attributes_stable_order.to_owned())
+            .expect("couldn't get attributes from map");
```
LPGhatguy, line 129: We should write an assertion about these bytes, maybe a snapshot test of their hex representation. Even if we don't review it, pinning down the representation would be good.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Reads through an attribute property (AttributesSerialize) and returns a map of attribute names -> values.
+pub fn get_attributes<R: Read>(mut value: R) -> Result<HashMap<Vec<u8>, Variant>, AttributeError> {
+    let size = read_u32(&mut value).map_err(|_| AttributeError::InvalidSize)?;
+    let mut attributes = HashMap::with_capacity(size as usize);
+
+    for _ in 0..size {
+        let name = read_string(&mut value).map_err(|_| AttributeError::InvalidName)?;
+
+        let attribute_key =
+            AttributeType::try_from(read_u8(&mut value).map_err(|_| AttributeError::NoValueType)?)?;
```
LPGhatguy, line 66: This is not a key, this is a type.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        let value = match attribute_key {
+            AttributeType::BrickColor => {
+                let color =
+                    read_u32(&mut value).map_err(|_| AttributeError::Other("BrickColor"))?;
+
+                Variant::BrickColor(
+                    BrickColor::from_number(color as u16)
+                        .ok_or(AttributeError::InvalidBrickColor(color))?,
+                )
```
LPGhatguy, line 76: Instead of naming `Variant::BrickColor` or any other variant, just use `.into()` here. Every type in `rbx_types` can be converted into a `Variant` that way.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,182 @@
...
+            AttributeType::Color3 => Variant::Color3(
+                read_color3(&mut value).map_err(|_| AttributeError::Other("Color3"))?,
+            ),
+
+            AttributeType::ColorSequence => {
+                let size = read_u32(&mut value).map_err(|_| AttributeError::InvalidSize)?;
+                let mut keypoints = Vec::with_capacity(size as usize);
+
+                for _ in 0..size {
+                    // `envelope` is not represented in rbx_types, apparently it's always zero.
```
LPGhatguy, line 92: Because this is a comment *in rbx_types*, the phrasing here is probably not what we want to commit. Let's prefer something like

> `envelope` is always zero and can be ignored.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,182 @@
...
+
+            AttributeType::Float32 => {
+                Variant::Float32(read_f32(&mut value).map_err(|_| AttributeError::Other("float"))?)
+            }
+
+            AttributeType::Float64 => {
+                let mut bytes = [0u8; 8];
+                value
+                    .read_exact(&mut bytes)
+                    .map_err(|_| AttributeError::Other("double"))?;
```
LPGhatguy, line 116: Let's make a method for this like the other types, even if it's only used here. Additionally, instead of calling it a "double", let's stick with `float64` in the error, too.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,182 @@
...
+                        .map_err(|_| AttributeError::Other("ColorSequenceKeypoint color"))?;
+
+                    keypoints.push(ColorSequenceKeypoint::new(time, color));
+                }
+
+                Variant::ColorSequence(ColorSequence { keypoints })
+            }
+
+            AttributeType::Float32 => {
+                Variant::Float32(read_f32(&mut value).map_err(|_| AttributeError::Other("float"))?)
```
LPGhatguy, line 109:
```suggestion
                Variant::Float32(read_f32(&mut value).map_err(|_| AttributeError::Other("float32"))?)
```

To disambiguate, I think it's helpful to always name the number of bits in errors like this. C's names for these things might be fairly ubiquitous, but that doesn't mean they're good.

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,182 @@
+// Thanks to Anaminus! https://github.com/RobloxAPI/rbxattr/blob/master/spec.md
+
+use super::*;
+use crate::{basic_types::*, brick_color::BrickColor, variant::Variant};
```
LPGhatguy, line 4: Can we spell out these globs? It's hard to tell where a type comes from, especially when there are multiple globs.

LPGhatguy (changes requested):
Implementation details are looking very good! It's great to finally have an implementation on the table for us to discuss. :D

## Attributes Type
I think this API is not quite enough. We should create a new type, `Attributes`, that would be suitable for including in `Variant` in the future. That type should contain the actual data (`HashMap<Vec<u8>, Variant>`) and expose methods for getting and setting attributes by key, as well as enumerating them.

We should then put the methods for encoding/decoding attributes on that struct. So something like this:

```rust
#[derive(Debug, Default, Clone, PartialEq)]
struct Attributes {
    data: HashMap<Vec<u8>, Variant>,
}

impl Attributes {
    pub fn new() -> Self;
    
    pub fn from_reader<R: Read>(reader: R) -> Result<Self, AttributesError>;
    pub fn to_writer<W: Write>(&self, writer: W) -> Result<Self, AttributesError>;

    pub fn get<K: Borrow<str>>(&self, key: K) -> Option<&Variant>;
    pub fn insert(&self, key: Vec<u8>, value: Variant) -> Option<Variant>;
    pub fn iter(&self) -> impl Iterator<Item = &Variant>;
}

// Addtional impls: IntoIterator, FromIterator<(String, Variant)>
```

## Snapshots
When snapshots include raw bytes instead of strings, it's harder to verify at a glance that they're correct. I recommended in the PR to switch from a sorted `Vec` into a `BTreeMap`, which should be the first part of this.

The second part is that I think we should drop non-UTF-8 attribute keys. Unless we're seeing these pop up in the wild, supporting them isn't very helpful and it just creates bad debug representations and confusing conversions at the boundaries of this crate.

Combined with creating a top-level type to hold attributes, we should be able to expand the API later without any breaking changes. Again, though, non-UTF-8 keys in attributes, just like non-UTF-8 tags, are likely very rare and can safely be considered malformed data.

rbx_types/src/attributes/mod.rs
Kampfkarren: This is from rbxattr's specifications, which marked them as "???". I can throw a comment in there.

rbx_types/src/attributes/mod.rs
Kampfkarren: I did this, but it turned out to be a bad idea. Because the internal structure is a HashMap, the byte string returned isn't deterministic.

This feels like it'll cause problems in Rojo saving files and making new diffs, should this just be a BTreeMap internally?

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+        })
+    }
+
+    /// Writes the attributes as a serialized string to the writer.
+    pub fn to_writer<W: Write>(&self, mut writer: W) -> Result<(), Error> {
+        let bytes = attributes_from_map(self.data.iter())?;
+        writer
+            .write_all(&bytes)
+            .map_err(AttributeError::ToWriterFail)
+            .map_err(Into::into)
```
LPGhatguy, line 125: Can we refactor `attributes_from_map` to accept a writer instead of buffering here? It kind of defeats the point of streaming output if we aren't streaming it!

rbx_types/src/attributes/reader.rs
```diff
@@ -0,0 +1,203 @@
...
+fn read_udim<R: Read>(mut reader: R) -> io::Result<UDim> {
+    Ok(UDim::new(read_f32(&mut reader)?, read_i32(&mut reader)?))
+}
+
+fn read_vector2<R: Read>(mut reader: R) -> io::Result<Vector2> {
+    Ok(Vector2::new(read_f32(&mut reader)?, read_f32(&mut reader)?))
+}
+
+/// Reads through an attribute property (AttributesSerialize) and returns a map of attribute names -> values.
+pub(crate) fn get_attributes<R: Read>(
```
LPGhatguy, line 74: I know this method is internal, but `get_attributes`/`attributes_from_map` don't feel like inverses of each other. Can we name this method `read_attributes` and change the other one to `write_attributes`?

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
+use super::*;
+
+use crate::{basic_types::*, variant::Variant};
```
LPGhatguy, line 3: Let's get rid of these globs here, too, so help us with type tracing.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+
+use std::{borrow::Borrow, io::Write};
```
LPGhatguy, line 5: Please sort your imports. std, then other crates, then within our crate.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+
+fn write_f32<W: Write>(mut writer: W, n: f32) {
+    writer
+        .write_all(&n.to_le_bytes().to_vec())
```
LPGhatguy, line 9: Definitely do not allocate a `Vec` for every four bytes we need to write, that's very excessive!

Use `&n.to_le_bytes()[..]` instead.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+        .expect("couldn't write f32 to buffer");
```
LPGhatguy, line 10: We should not panic, we should propagate this error.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// Writes the attribute property (AttributesSerialize) from a map of attribute names -> values.
+pub(crate) fn attributes_from_map<
+    K: Borrow<String>,
+    V: Borrow<Variant>,
+    I: Iterator<Item = (K, V)> + ExactSizeIterator,
+    M: IntoIterator<IntoIter = I, Item = (K, V)>,
+>(
+    map: M,
+) -> Result<Vec<u8>, AttributeError> {
```
LPGhatguy, line 53: We should be okay to make this method just take the `HashMap<String, Variant>`. That's the actual type we store these things in.

I mentioned it elsewhere, but this method should take `<W: Write>`. Having it allocate the space instead of writing into a user-provided buffer is much less flexible.

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+                    write_color3(&mut bytes, keypoint.color);
+                }
+            }
+            (AttributeType::Float32, Variant::Float32(float)) => {
+                write_f32(&mut bytes, *float);
+            }
+            (AttributeType::Float64, Variant::Float64(float)) => {
+                bytes
+                    .write_all(&float.to_le_bytes().to_vec())
+                    .expect("couldn't write double to buffer");
```
LPGhatguy, line 90: Please factor out this into its own method and adopt the advice from the other versions of it (ie avoiding allocating an 8 byte `Vec`).

rbx_types/src/attributes/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+    bytes
+        .write_all(&(map.len() as u32).to_le_bytes())
+        .expect("couldn't write map length to buffer");
+
+    for (name, variant) in map {
+        let variant = variant.borrow();
+
+        write_string(&mut bytes, name.borrow());
+
+        let attribute_type = AttributeType::try_from(variant.ty())?;
```
LPGhatguy, line 66: Why do we need `AttributeType` here? Why not just match on `Variant` and then return an error for variants that we don't match?

rbx_types/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum InnerError {
+    #[error(transparent)]
+    AttributeError {
+        #[from]
+        source: crate::attributes::AttributeError,
```
LPGhatguy, line 33: We're already using this type in this file, aren't we?

rbx_types/src/attributes/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+    InvalidVariantType,
+
+    #[error("no value type was found")]
+    NoValueType,
+
+    #[error("malformed attribute key")]
+    MalformedEntryKey(FromUtf8Error),
+
+    #[error("couldn't write to writer: {0}")]
+    ToWriterFail(io::Error),
```
LPGhatguy, line 212: When tracking a source error in Rust, it's unidiomatic to include the source error in your error's message. Instead, thiserror has a `#[source]` attribute (which is automatically applied to any fields named `source`) that does the right thing for you!

LPGhatguy (changes requested):

rbx_types/src/attributes/writer.rs
Kampfkarren: It was a panic because it's using a vector buffer, but now that it's going to be a Write, propagating makes more sense.

rbx_types/src/attributes/writer.rs
Kampfkarren: Yeah, I was thinking about simplifying the types now that it's private.

rbx_types/src/error.rs
Kampfkarren: Oops, that was from an earlier attempt.

rbx_types/src/attributes/mod.rs
Kampfkarren: What *is* the right thing out of curiosity? What does `#[source]` do?

rbx_types/src/attributes/writer.rs
Kampfkarren: It's to then push its type as a byte.

rbx_types/src/attributes/mod.rs
LPGhatguy: The correct thing to do is to write an error message without including the inner error, like "couldn't write to writer" or something more specific.

Adding `#[source]` makes the generated `std::error::Error` implementation have a correct value for [`Error::source`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.source). When the error is printed with debug, the chain of source errors will be printed out!

eAi:
It'd be great to have this PR merged (and support for this in Rojo)

Kampfkarren:
I need to fix up some lints on this but I don't really want to start unless I know it'll be merged soon after.

PR #168, opened 2021-04-09
Make rbx_binary resilient to unknown value types

LPGhatguy:
This PR fixes two cases:

- PROP chunks with no data are now silently ignored in order to match Roblox behavior. This is important with the new partly-released `OptionalCoordinateFrame` type.
- Unknown value types are now ignored, but print one warning per unknown type ID. The warning contains the instance type and property name of the first property in the file that has this byte. I wanted to reduce the amount of warning spam that might happen when new files starting including a type.

We'll want to backport this to dom-v1 so that we're able to patch Rojo 6.x and potentially 0.5.x as well.

### DIFF
--- a/rbx_binary/src/deserializer.rs
+++ b/rbx_binary/src/deserializer.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     convert::TryInto,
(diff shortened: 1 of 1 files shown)


LPGhatguy:
(Moving ahead with this; we can revise further after it lands!)

badcc:
This is awesome, thanks guys!

PR #169, opened 2021-04-09
Implement OptionalCoordinateFrame in rbx_xml

LPGhatguy:
Depends on changes in rbx_types to add this type.


kennethloeffler (approved):

LPGhatguy:
Ugh, GitHub Actions is down. I ran everything locally (and we have an earlier run) that should ensure this is okay.

PR #171, opened 2021-04-11
Write CFrame basic rotation IDs

### LINKED ISSUE #129: Implement writing special CFrame cases in the binary format
In the binary format, there are several special CFrame cases where the rotation matrix isn't written (see: [the documentation on CFrame](https://github.com/rojo-rbx/rbx-dom/blob/master/format/binary.md#cframe)).

PR #128 implements CFrame in the binary format for the most part, but doesn't write the special cases any differently than normal CFrames. This doesn't impact the functionality of rbx_binary any, but it does impact the space taken up by CFrames, so writing these special cases should be implemented at some point.

kennethloeffler:
Closes #129.

### DIFF
--- a/rbx_binary/src/deserializer.rs
+++ b/rbx_binary/src/deserializer.rs
@@ -18,6 +18,7 @@ use rbx_reflection::DataType;
 use thiserror::Error;
 
 use crate::{
+    cframe::from_basic_rotation_id,
     chunk::Chunk,
     core::{
         find_canonical_property_descriptor, RbxReadExt, FILE_MAGIC_HEADER, FILE_SIGNATURE,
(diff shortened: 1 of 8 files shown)


kennethloeffler:
> Creates a module and moves CFrame-related logic to
it. Renames special_case_to_rotation to to_basic_rotation_id
Its counterpart, from_basic_rotation_id, goes in the other
direction from rotation matrices to ids.

Whoops - these methods should be listed in the opposite order here.

kennethloeffler:
This does bring up the question of if we should do anything about non-orthonormal matrices, whether in the new `cframe` module or further up in `rbx_types`. Roblox Studio orthonormalizes CFrames' rotation matrices when setting them, so this doesn't seem like much of a problem to me - especially considering that one would really have to try in order to have a file that actually contains CFrames like this. Thus I wonder whether the above check is at all necessary. 

Without the comparison, in cases where they *do* differ, the end result would be one of the basic rotation matrices instead of the later orthonormalization by Studio. With respect only to this section of code: do we test for a bad matrix, and then pretend we never saw it (the current behavior)? or do we not test, and as a result inadvertently alter the contents of the file? I like the first option - we should probably make reasonable efforts to reflect the contents in as close to their original state as possible.

LPGhatguy:
I think we should leave non-orthonormalized matrices alone.

One goal of rbx-dom (that isn't really written down) is that we should be able to handle any _structurally_ valid model file, even if it's not _semantically_ valid. Roblox's DOM implementation, on the other hand, also does strict _semantic_ validation, which might include forcing CFrames to be orthonormalized when used on parts.

Dekkonot:
> I think we should leave non-orthonormalized matrices alone.

Agreed. If we start enforcing semantic correctness on files, it can get messy really quick. A DOM shouldn't decide what is or isn't correct as long as it can read it.

kennethloeffler:
Worth noting that it's now possible to give `from_basic_rotation_id` an invalid id and get back a garbage matrix.  Is this a problem?

Dekkonot:
I'm going to say no because the only way that will happen is if a codec like ours does it. Studio never produces garbage ID's like that.

rbx_binary/src/deserializer.rs
```diff
@@ -909,7 +783,7 @@ impl<R: Read> BinaryDeserializer<R> {
                                 ),
                             ));
                         } else {
-                            let special_case = special_case_to_rotation(id).ok_or_else(|| {
+                            let special_case = from_basic_rotation_id(id).ok_or_else(|| {
```
kennethloeffler, line 786: This will error for `id > 0x24`, but not for something like `id == 0x04` or `id == 0x24`. Maybe a little funky?

kennethloeffler: I don't think I like this behavior, so we're going back to the match.

kennethloeffler: On the other hand, we could guard against these - but that's probably less clear than simply enumerating all the valid ids in the match expression.

LPGhatguy:
How does Studio handle being given a garbage basic rotation ID? I'm worried that if more special cases are ever added, we'll silently corrupt them instead of knowingly fail.

If possible, we should reject invalid rotation IDs.

kennethloeffler:
We do currently reject them with an error: the deserializer emits an `InnerError::InvalidOrientationId` when `from_basic_rotation_id` fails to match. I think we should stick with a match for this function, because guarding against the invalid ids in the other implementation makes the code look way too voodoo for my taste.

I'm not sure what Roblox Studio does with them (yet) - I think I want to make a utility that lets me easily screw with property data blocks :P

rbx_types/src/basic_types.rs
```diff
@@ -74,10 +76,45 @@ pub struct Vector3 {
...
 }
 
+fn approx_unit_or_zero(value: f32) -> Option<i32> {
+    if abs_diff_eq!(value.abs(), 1.0) {
+        if value.is_sign_positive() {
+            Some(1)
+        } else {
+            Some(-1)
+        }
+    } else if abs_diff_eq!(value.abs(), 0.0) {
```
Dekkonot, line 86: How does this differ from `abs_diff_eq!(value, 0.0)`?

rbx_types/src/basic_types.rs
```diff
@@ -74,10 +76,45 @@ pub struct Vector3 {
...
+        Some(0)
+    } else {
+        None
+    }
+}
+
+fn get_normal_id(position: u8, value: i32) -> Option<u8> {
```
Dekkonot, line 93: This feels rather cryptic - is there a chance we could get some documentation for it in-file?

Dekkonot:
@kennethloeffler I was more interested in getting an explanation of how it works rather than what it does. That's my fault, I wasn't very clear.

After giving it some thought though, I don't think that's necessary. It's pretty self-apparent *what* it's doing, and the how boils down to "this is how it works in Roblox".

LPGhatguy (changes requested):

rbx_types/src/basic_types.rs
```diff
@@ -77,27 +75,16 @@ pub struct Vector3 {
...
-            Some(1)
-        } else {
-            Some(-1)
-        }
-    } else if abs_diff_eq!(value, 0.0) {
+    if value.abs() <= f32::EPSILON {
         Some(0)
+    } else if value.abs() - 1.0 <= f32::EPSILON {
+        let one: f32 = 1.0;
+        Some(one.copysign(value) as i32)
```
LPGhatguy, line 82:
```suggestion
        Some(1.0f32.copysign(value) as i32)
```

or maybe

```suggestion
        Some(1 * value.signum() as i32)
```

When I made this suggestion, I don't think I realized that the return value was `i32`. Whichever of these makes the most sense to you should be good!

LPGhatguy (approved):

PR #176, opened 2021-04-24
Implement OptionalCoordinateFrame in rbx_binary

kennethloeffler:
There's quite a bit of code duplicated between `CFrame` and `OptionalCFrame`. It's not clear to me how we can easily avoid it with the current structure of the code, especially in the deserializer.

`yaml-rust` serializes a `None` as a `~` in our snapshots. Do we want to change this, or is it okay?

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,42 +58,43 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside 
 
 ## Property Type Coverage
 
-| Property Type      | Example Property                | rbx_types | rbx_dom_lua | rbx_xml | rbx_binary
-|:------------------ |:------------------------------- |:--:|:--:|:--:|:--:|
-| Axes               | `ArcHandles.Axes`               | ✔ | ✔ | ✔ | ✔ |
-| BinaryString       | `Terrain.MaterialColors`        | ✔ | ➖ | ✔ | ✔ |
-| Bool               | `Part.Anchored`                 | ✔ | ✔ | ✔ | ✔ |
-| BrickColor         | `Part.BrickColor`               | ✔ | ✔ | ✔ | ✔ |
-| CFrame             | `Camera.CFrame`                 | ✔ | ✔ | ✔ | ✔ |
-| Color3             | `Lighting.Ambient`              | ✔ | ✔ | ✔ | ✔ |
-| Color3uint8        | `Part.BrickColor`               | ✔ | ✔ | ✔ | ✔ |
-| ColorSequence      | `Beam.Color`                    | ✔ | ✔ | ✔ | ✔ |
-| Content            | `Decal.Texture`                 | ✔ | ✔ | ✔ | ✔ |
-| Enum               | `Part.Shape`                    | ✔ | ✔ | ✔ | ✔ |
-| Faces              | `Handles.Faces`                 | ✔ | ✔ | ✔ | ✔ |
-| Float32            | `Players.RespawnTime`           | ✔ | ✔ | ✔ | ✔ |
-| Float64            | `Sound.PlaybackLoudness`        | ✔ | ✔ | ✔ | ✔ |
-| Int32              | `Frame.ZIndex`                  | ✔ | ✔ | ✔ | ✔ |
-| Int64              | `Player.UserId`                 | ✔ | ✔ | ✔ | ✔ |
-| NumberRange        | `ParticleEmitter.Lifetime`      | ✔ | ✔ | ✔ | ✔ |
-| NumberSequence     | `Beam.Transparency`             | ✔ | ✔ | ✔ | ✔ |
-| PhysicalProperties | `Part.CustomPhysicalProperties` | ✔ | ✔ | ✔ | ✔ |
-| ProtectedString    | `ModuleScript.Source`           | ✔ | ✔ | ✔ | ✔ |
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -58,42 +58,43 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside
...
+| Content                 | `Decal.Texture`                 | ✔ | ✔ | ✔ | ✔ |
+| Enum                    | `Part.Shape`                    | ✔ | ✔ | ✔ | ✔ |
+| Faces                   | `Handles.Faces`                 | ✔ | ✔ | ✔ | ✔ |
+| Float32                 | `Players.RespawnTime`           | ✔ | ✔ | ✔ | ✔ |
+| Float64                 | `Sound.PlaybackLoudness`        | ✔ | ✔ | ✔ | ✔ |
+| Int32                   | `Frame.ZIndex`                  | ✔ | ✔ | ✔ | ✔ |
+| Int64                   | `Player.UserId`                 | ✔ | ✔ | ✔ | ✔ |
+| NumberRange             | `ParticleEmitter.Lifetime`      | ✔ | ✔ | ✔ | ✔ |
+| NumberSequence          | `Beam.Transparency`             | ✔ | ✔ | ✔ | ✔ |
+| OptionalCoordinateFrame | `Model.WorldPivotData`          | ✔ | ❌ | ❌ | ✔ |
```
LPGhatguy, line 80: Curse you, vertical alignment!!

format/binary.md
```diff
@@ -606,6 +607,15 @@ When an array of Int64s is present, the bytes of the integers are subject to [by
...
 SharedStrings are stored as an [Interleaved Array](#byte-interleaving) of `u32`s that represent indices in the [`SSTR`](#sstr-chunk) string array.
 
+### OptionalCoordinateFrame
+**Type ID 0x1E**
+
+`OptionalCoordinateFrame` is stored the same way as [CFrame](#cframe), but with a couple interesting differences:
+* Immediately following OptionalCoordinateFrame's type ID is CFrame's type ID (`10`);
+* At the end of the chunk there is an array of Bools (preceded by Bool's type ID, `02`) that indicates which OptionalCoordinateFrames have a value.
+
+An `OptionalCoordinateFrame` with value `CFrame.new(0, 0, 1, 0, -1, 0, 1, 0, 0, 0, 0, 1)` followed by an `OptionalCoordinateFrame` with no value looks like this: `10 0a 02 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 7f 00 00 00 00 00 00 00 02 01 00`. Note that the valueless `OptionalCoordinateFrame` is written as the identity `CFrame` with its corresponding boolean `00` codifying its valuelessness.
```
LPGhatguy, line 617: Funky.

rbx_binary/src/deserializer.rs
```diff
@@ -1123,6 +1120,86 @@ impl<R: Read> BinaryDeserializer<R> {
...
                 }
             },
+            Type::OptionalCFrame => match canonical_type {
+                VariantType::OptionalCFrame => {
+                    let referents = &type_info.referents;
+                    let mut rotations = Vec::with_capacity(referents.len());
+
+                    // Roblox writes a type marker for CFrame here that we don't
+                    // need to use.
+                    chunk.read_u8()?;
```
LPGhatguy, line 1130: Let's assert that this value is what we expect. It's weird that it exists, but I'm a little worried that this type ID might just mean "Optional <something>" and we're intended to ascribe meaning to this byte. It could be used in the future to have more optional value types.

rbx_binary/src/deserializer.rs
```diff
@@ -1123,6 +1120,86 @@ impl<R: Read> BinaryDeserializer<R> {
...
+                    let mut y = vec![0.0; referents.len()];
+                    let mut z = vec![0.0; referents.len()];
+
+                    chunk.read_interleaved_f32_array(&mut x)?;
+                    chunk.read_interleaved_f32_array(&mut y)?;
+                    chunk.read_interleaved_f32_array(&mut z)?;
+
+                    // Roblox writes a type marker for Bool here that we don't
+                    // need to use.
+                    chunk.read_u8()?;
```
LPGhatguy, line 1173: Okay, while the other type marker might make sense, this seems pretty weird. I think we should also assert on its value and return an error if it doesn't match `Bool`.

rbx_types/src/basic_types.rs
```diff
@@ -166,6 +166,8 @@ impl CFrame {
     }
 }
 
+pub type OptionalCFrame = Option<CFrame>;
```
LPGhatguy, line 169: I don't think we need this typedef because of how idiomatic and pervasive `Option` is.

LPGhatguy (changes requested):
Minor comments. This looks really good!

rbx_binary/src/deserializer.rs
```diff
@@ -1127,7 +1127,7 @@ impl<R: Read> BinaryDeserializer<R> {
 
                     // Roblox writes a type marker for CFrame here that we don't
                     // need to use.
-                    chunk.read_u8()?;
+                    assert!(chunk.read_u8()? == Type::CFrame as u8);
```
LPGhatguy: Oh, sorry, I meant to return an error here! We don't want to panic un malformed input if we can.

LPGhatguy (approved):

PR #177, opened 2021-05-07
Verify scriptability of properties in generate_reflection

### LINKED ISSUE #153: generate_reflection should validate scriptability of properties
Some properties, like `Instance.SourceAssetId`, are marked as read-write scriptable in the reflection dump from Roblox Studio. However, they're only read-write scriptable for core scripts, which people outside Roblox can't write. As such, we should try to read/write every property on every instance to assign a new scriptability value that's relevant for users of rbx-dom.

kennethloeffler:
This PR reintroduces the scriptability check from 5bca08f that we subsequently lost in 9866d06. Closes #153.


generate_reflection/plugin/main.lua
```diff
@@ -2,12 +2,154 @@ local HttpService = game:GetService("HttpService")
...
+
+				superclassChanges[propertyName] = getPropertyChange(instance, propertyDescriptor)
+			end
+
+			if next(superclassChanges) then
+				propertyChanges[superclassName] = superclassChanges
+			end
+
+			superclassName = superclassDescriptor.Superclass
+		end
```
kennethloeffler, line 144: Thought of a possible issue here: if we can't read/write superclass properties on a particular subclass, but *can* on other subclasses, *and* we happen measure the former kind of subclass first, then these properties' measured scriptability will be wrong! AFAIK this is only possible with deprecated classes like `Speaker` that subclass `Instance` - we currently ignore these properties across the board, thus avoiding the issue. Should we worry about this?

kennethloeffler:
This seems like a more reliable approach to me. Are there perhaps any problems I'm not seeing with relying on the API dump for this information?

LPGhatguy (changes requested):
This is a wonderful change. Minor comment on `if-let` vs `match` and then I think we're good to go.

LPGhatguy (approved):

PR #181, opened 2021-05-28
Make generate_reflection save the fixture place automatically

kennethloeffler:
enigo has some problems (at least on Windows)<sup>1</sup>, and to get this to work more-or-less reliably we need to do things with the WinAPI that no crate I'm aware of currently does... so we'll do it live! 

This PR adds the struct `StudioKeyboard` that can send keystrokes to Roblox Studio, or indeed any process' top-level visible window. It does so by attaching Roblox Studio's input processing mechanism to the current foreground thread's via `AttachThreadInput` and sending keyboard events with `SendInput`. AutoHotKey also uses this technique. 

The failure cases feel kind of bad as an end user who is probably using cmd or powershell, but it definitely makes it easier for us to use this tool and makes it possible to run it on a CI machine. 

<sup>1</sup>For example, it uses a `transmute_copy` that invokes UB  to create keyboard events. It also doesn't properly use `SendInput` - the entire point is that it can insert multiple events serially into the input stream (which is why it takes an array), but enigo only passes one at a time!

### DIFF
--- a/generate_reflection/Cargo.toml
+++ b/generate_reflection/Cargo.toml
@@ -26,3 +26,4 @@ structopt = "0.3.9"
 tempfile = "3.1.0"
 tiny_http = "0.7.0"
 toml = "0.5.0"
+winapi = { version = "0.3.9", features = ["winuser", "impl-default"] }
--- a/generate_reflection/src/defaults_place.rs
+++ b/generate_reflection/src/defaults_place.rs
@@ -21,7 +21,10 @@ use rbx_reflection::{PropertyDescriptor, PropertyKind, PropertySerialization, Re
 use roblox_install::RobloxStudio;
 use tempfile::tempdir;
 
-use crate::plugin_injector::{PluginInjector, StudioInfo};
+use crate::{
+    plugin_injector::{PluginInjector, StudioInfo},
+    studio_keyboard::{Key, StudioKeyboard},
+};
 
 /// Use Roblox Studio to populate the reflection database with default values
 /// for as many properties as possible.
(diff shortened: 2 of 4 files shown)


kennethloeffler:
I've just sprinkled in some cfgs to get this passing, but we'll need to add Windows as a toolchain / target in the workflow if we want this to actually be checked.

Kampfkarren:
Are you guys *sure* this isn't a better solution than setting auto-recovery interval to a minute and waiting until an auto-save is put into a folder?

kennethloeffler:
This does work mechanically, but has a problem: Roblox Studio's autosaves are `.rbxl` files! generate_reflection currently uses rbx_xml to read the saved place file, using an option that completely bypasses the reflection database. rbx_binary does not have such an option. We'd need to think about this carefully and possibly implement a similar mechanism in rbx_binary. I've generated the JSON database as part of this commit for inspection: as-is, `BasePart.Color` turns into a `BrickColor` (from `Int32`).

There are a couple other problems that are not addressed by this commit. First, the autosave directory will not necessarily be `Documents/ROBLOX/AutoSaves` (and we cannot change this from Lua). Second, the existing property patches were created with `.rbxlx` files in mind and do not apply to `.rbxl` files.

Kampfkarren:
The autosave directory is a setting, I thought.

kennethloeffler:
It's a `QDir`!

kennethloeffler:
The scope of this work has increased somewhat, so this is the last change I'll make here before splitting off tomorrow into rbx_binary and roblox-install to get the rest done. roblox-install will need to read and write Studio settings. I still need to look at what exactly needs to be done to start using rbx_binary here.

LPGhatguy:
This is starting to look like something I'd be down to ship!

kennethloeffler:
I've removed all the earlier junk and replaced it with a compiled AutoHotKey script. Notably, this obviates the need for both #183 and #182.

kennethloeffler:
Seems we've pretty well exhausted the possible solutions. We cannot rely on Roblox Studio's auto-recovery feature because it outputs `.rbxl` files - we need an `.rbxlx` to generate our database, otherwise it becomes extremely difficult to generate correct XML. Requiring an AutoHotKey installation feels bad. Just slapping in a compiled script feels bad. There are no existing crates that can ensure Roblox Studio receives our keystrokes. 

The right solution seems to be AutoHotKey Except It's a Rust Library. I am starting work on a minimal crate that will provide functionality similar to 0269d7d so we can finally finish up this feature.

kennethloeffler:
...and there we are. AutoHotkey does a very elaborate song and dance to handle the problem of modifiers being down during the send, so I chose to just ignore that for now.

PR #182, opened 2021-05-31
Implement QDir in rbx_xml

kennethloeffler:
`generate_reflection` needs to be able to access Roblox Studio's auto-save path (see #181). It is stored in what is essentially a `.rbxmx` file in Roblox's installation root. Here's a piece of trivia for you: Roblox Studio will correctly read a path delimited with any number of (possibly interleaved) forward and back slashes, but always writes it back delimited with one forward slash.


rbx_types/src/variant.rs
```diff
@@ -126,6 +126,7 @@ make_variant! {
     Vector3(Vector3),
     Vector3int16(Vector3int16),
     OptionalCFrame(Option<CFrame>),
+    QDir(std::path::PathBuf),
```
LPGhatguy, line 129: Should we newtype this? I'm a little afraid of the format eventually growing a not-weird-Qt-type path. We can sort of follow the lead from `Content` here.

LPGhatguy (changes requested):
Easy implementation, question about future compatibility.

Can you also update the relevant CHANGELOG files?

kennethloeffler:
I just scattershot implemented various traits - let me know if you think we need more!

rbx_types/src/qdir.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl FromStr for QDir {
+    type Err = Infallible;
+
+    fn from_str(path: &str) -> Result<Self, Infallible> {
+        Ok(Self {
+            path_buf: PathBuf::from(path),
+        })
+    }
+}
```
LPGhatguy, line 106: Let's narrow these down to just `From<&Path>` and `From<PathBuf>`. Every impl we have here is another thing we have to uphold for forward compatibility and I think keeping a super niche type like this slim is probably okay.

rbx_types/src/qdir.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl Deref for QDir {
+    type Target = Path;
+
+    fn deref(&self) -> &Path {
+        self.path_buf.as_path()
+    }
+}
```
LPGhatguy, line 56: I don't think a deref here is all that important. We can `AsRef` into `Path`, we're not making a total drop-in replacement.

rbx_types/src/qdir.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn into_path_buf(self) -> PathBuf {
+        self.path_buf
+    }
+}
+
+impl AsRef<OsStr> for QDir {
+    fn as_ref(&self) -> &OsStr {
+        self.path_buf.as_os_str()
+    }
+}
```
LPGhatguy, line 36: I don't think we need this impl

rbx_xml/src/types/qdir.rs
```diff
@@ -0,0 +1,42 @@
+use std::io::{Read, Write};
+
+use crate::{
+    core::XmlType,
+    deserializer_core::XmlEventReader,
+    error::{DecodeError, EncodeError},
+    serializer_core::XmlEventWriter,
+};
+
+use rbx_dom_weak::types::QDir;
```
LPGhatguy, line 10:
```suggestion
use rbx_dom_weak::types::QDir;

use crate::{
    core::XmlType,
    deserializer_core::XmlEventReader,
    error::{DecodeError, EncodeError},
    serializer_core::XmlEventWriter,
};
```

We usually order external crates above internal imports.

LPGhatguy (changes requested):
Too many impls! I think we can cut down on these substantially because of how niche of a type this is.

kennethloeffler:
Further work on #181 shows this to be unnecessary.

PR #183, opened 2021-06-02
Add an option to disable reflection in `rbx_binary`

kennethloeffler:
`rbx_binary` needs to bypass the reflection database for #181. This PR thus adds the enum `DecodePropertyBehavior` to the deserializer. It has two variants: `Default` and `NoReflection`. `Default` corresponds to the original behavior. `NoReflection` causes the deserializer to read property names and types verbatim without consulting the reflection database.

`rbx_xml` has a similar mechanism: the struct `DecodeOptions`. I'm not 100% on whether we should also use such a struct here, or if an enum is enough (since we only need the one option). I'm also not 100% on the ramifications of totally bypassing reflection here - there might be something spooky that I've missed!

### DIFF
--- a/rbx_binary/CHANGELOG.md
+++ b/rbx_binary/CHANGELOG.md
@@ -1,6 +1,7 @@
 # rbx_binary Changelog
 
 ## Unreleased
+* Added an option to disable property reflection in the deserializer.
 
 ## 0.6.0-alpha.5 (2021-05-14)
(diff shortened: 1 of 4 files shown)


kennethloeffler:
One point of breakage I've found so far (besides the string types): `Color3uint8` is canonically a `Color3`, so the deserializer believed the type was incorrect.

kennethloeffler:
Further work on #181 shows this to be unnecessary.

PR #184, opened 2021-06-03
Fix null refs not being written for invalid instance IDs

kennethloeffler:
The serializer does not handle the case where an instance ID refers to an unserialized instance, resulting in nothing being written. This PR ensures that the null `Ref` is written for all IDs that do not refer to an instance, whether their Refs are `None` or are simply invalid.

### DIFF
--- a/rbx_binary/CHANGELOG.md
+++ b/rbx_binary/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 * Fixed null Refs not being written for unserialized instances. ([#184][pr-184])
 
-[pr-184]: https://github.com/rojo-rbx/pull/184
+[pr-184]: https://github.com/rojo-rbx/rbx-dom/pull/184
 
 ## 0.6.0-alpha.5 (2021-05-14)
 * Added `OptionalCoordinateFrame` support. ([#176][pr-176])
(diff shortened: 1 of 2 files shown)


kennethloeffler:
Maybe we should explicitly check for an invalid ID and spit out a warning. Referring to an unserialized instance seems like a mistake - Roblox Studio silently overwrites these types of refs with null, and in the case of descended Welds with an invalid `Part0` / `Part1` just totally nukes the instance [!!!], but this may very well not be The Right Thing to Do!

LPGhatguy:
Because our IDs are pretty harmless to leave dangling, I'm okay with this solution. Can you add a changelog entry?

rbx_binary/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # rbx_binary Changelog
 
 ## Unreleased
-* Fixed null Refs not being written for unserialized instances. [#184][pr-184]
+* Fixed null Refs not being written for unserialized instances.
```
LPGhatguy, line 4: Having a link here would be good. You need to add a `[pr-184]: https://github.com/rojo-rbx/rbx-dom/pull/184` note at the bottom of the section just like we do for the other PRs.

LPGhatguy (approved):

PR #188, opened 2021-06-16
Refresh the binary format specification

kennethloeffler:
This PR cleans up inconsistent formatting across the document and renames various chunk fields for (hopefully) increased clarity.

### DIFF
--- a/format/binary.md
+++ b/format/binary.md
@@ -453,8 +453,8 @@ The `Vector3int16` type is stored as three little-endian i16s:
 | Field Name | Format | Value                                |
 |:-----------|:-------|:-------------------------------------|
 | X          | `i16`    | The `X` component of the Vector3in16 |
-| X          | `i16`    | The `Y` component of the Vector3in16 |
-| X          | `i16`    | The `Z` component of the Vector3in16 |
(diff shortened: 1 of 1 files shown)


format/binary.md
```diff
@@ -51,12 +51,14 @@ This document is based on:
 	- [Roblox Float Format](#roblox-float-format)
 
 ## Document Conventions
-This document assumes a basic understanding of Rust's conventions for numeric types. For example:
+We assume a basic understanding of Rust's conventions for numeric types. For example:
```
LPGhatguy, line 54: I think that a document like this should avoid having pronouns of any kind, kinda like Wikipedia's writing style. We should keep this how it is or reword this to be passive somehow.

LPGhatguy (commented):
This looks great. One last nitpick.

I think the pluralization of types in the doc looks weird:

> blah blah `String`s blah blah

This might be a little bit cleaner:

> blah blah `String` values blah blah

LPGhatguy (approved):

LPGhatguy:
Slam dunk! Thanks for all your work on this.

PR #194, opened 2021-07-11
Specify CSK value as color components

Anaminus:
Instead of treating the Color field of ColorSequenceKeypoint as an exception, CSK is specified as having fields for each color component. This is similar to how Ray is specified as six floats instead of two Vector3s.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -514,13 +514,15 @@ The `ColorSequence` type is stored as a `u32` indicating how many `ColorSequence
 
 `ColorSequenceKeypoint` is a struct composed of the following fields:
 
-| Field Name        | Format              | Value                                          |
-|:------------------|:--------------------|:-----------------------------------------------|
-| Time              | `f32`               | The time value of the `ColorSequenceKeypoint`  |
-| Color             | [`Color3`](#Color3) | The color value of the `ColorSequenceKeypoint` |
-| Envelope (unused) | `f32`               | n/a; serialized, but not used                  |
+| Field Name        | Format              | Value                                              |
+|:------------------|:--------------------|:---------------------------------------------------|
+| Time              | `f32`               | The time value of the `ColorSequenceKeypoint`      |
+| ColorR            | `f32`               | The red component of the keypoint's color value.   |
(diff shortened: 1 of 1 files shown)


docs/binary.md
```diff
@@ -514,13 +514,15 @@ The `ColorSequence` type is stored as a `u32` indicating how many `ColorSequence
...
-|:------------------|:--------------------|:-----------------------------------------------|
-| Time              | `f32`               | The time value of the `ColorSequenceKeypoint`  |
-| Color             | [`Color3`](#Color3) | The color value of the `ColorSequenceKeypoint` |
-| Envelope (unused) | `f32`               | n/a; serialized, but not used                  |
-
-Note that `Color` is **not** subject to interleaving like a normal `Color3`. When multiple `ColorSequence` values are present, they are stored in sequence with no transformation or interleaving. Two `ColorSequence` values
+| Field Name        | Format              | Value                                              |
+|:------------------|:--------------------|:---------------------------------------------------|
+| Time              | `f32`               | The time value of the `ColorSequenceKeypoint`      |
+| ColorR            | `f32`               | The red component of the keypoint's color value.   |
```
Dekkonot, line 520: It's probably fine to just use `R`/`G`/`B` instead of `ColorR` etc.; it's awkward to read with the current proposed names.

Anaminus: I went with ColorR since it matched UDim2.ScaleX *that I now realize isn't specified that way*. Okay then. I guess it's also similar to the Ray fields. Could be `R`, `Color R`, or even `Red`. Any preference? I'm not really sure what the spec is going for when listing field names. They seem to be halfway between short descriptions and identifiers.

Dekkonot: There's a lack of standardization for this spec so realistically anything would be fine, but I would probably go with `R` since that's what we have Color3's components listed as.

We'll probably have to go through and change a few other types to match this format but I think it's a better way to document things, so I'll put in the work (assuming you don't).

Dekkonot (approved):

LPGhatguy (approved):
Very nice change!

Issue #193, opened 2021-07-01
Improve definition of OptionalCoordinateFrame in binary spec


Anaminus:
- Make "Optional" a generic type, rather than the specific "OptionalCoordinateFrame".
- Combine the Type ID and Value fields of the PROP chunk into a single "Values" field.
- A Values structure begins with an ID that identifies the type, which determines the shape of the remaining structure (usually an array of elements).
- **The Values field of a PROP chunk is optional**. That is, if the chunk ends before the Values field can be read, this can indicate that 1) no values are assigned, or 2) the type is Optional, where each property has no value.
    - The correct variation can be determined by constructing an empty PROP chunk for a non-Optional property type. If it succeeds for all non-Optional property types, then the first variation is correct. If it only succeeds for Optional types, then the second variation is correct.
- A Values with an ID of `0x1e` indicates the type "Optional". The structure of Optional is two Values. The first has some ID (such as CFrame), and the second must(?) be the type Bool (0x02). Each element in the second array indicates whether the corresponding value in the first array has a value.

Dekkonot:
> A Values with an ID of 0x1e indicates the type "Optional".

That seems like a pretty big assumption based on one property. I don't have any qualms with it but I would be a rather significant departure from the rest of the format in terms of design, so I'd want to be certain it's the truth. How did you come to this conclusion?

Anaminus:
The evidence for a Values field is the structure of the type. The presence of the first additional type ID suggests the reuse of a code path that parses value arrays. The presence of the second additional ID *strongly* suggests this, given that it is otherwise entirely unnecessary information (assuming that the second sub-array is always a Bool, which seems reasonable). At the very least, the specification should be updated to reflect this.

It is also reasonable to believe that type 0x1e is a generic type, not only because of the additional type IDs that are included in the structure, but because of nature of the type itself; an option type always involves some other type. Even the XML version suggests it by wrapping content of the value in a distinct inner tag that indicates the type. If OptionalCoordinateFrame was really a stand-alone type, then this tag would be redundant.

[...]

PR #195, opened 2021-07-13
Rename OptionalCoordinateFrame to Optional and update documentation

Dekkonot:
As per #193:
 - Rename `OptionalCoordinateFrame` to `Optional`
 - Revamp the `PROP` chunk description to make `Values` its own field
 - Redefine `Optional` to be a struct composed of two `Values`

Notably, I did not indicate that the `Values` field of `PROP` is optional, despite it being proposed in #193. This is because the one time it has been not present, it was indicated to be a mistake by the engineer responsible so I don't anticipate it happening again. It also complicates implementations somewhat to assume it's optional, so I'd prefer we not.

I am more than open to feedback on the phrasing or layout of the revisions because I'm not certain of them, but this should be a good starting point if nothing else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,43 +58,43 @@ Roblox Lua implementation of DOM APIs, allowing Instance reflection from inside 
 
 ## Property Type Coverage
 
-| Property Type           | Example Property                | rbx_types | rbx_dom_lua | rbx_xml | rbx_binary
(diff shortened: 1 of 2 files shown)


Dekkonot:
Okay well I didn't touch any Rust so that's not my fault but we should probably fix that

PR #198, opened 2021-07-14
Add conversions between Color3 and Color3uint8

kennethloeffler:
`BasePart.Color` was being incorrectly serialized as a `Color3` instead of a `Color3uint8` in the XML format, as per rojo-rbx/rojo#443. This PR adds conversions between these two types mainly to address this issue, but also because they will likely be useful in the future.


LPGhatguy (approved):
LGTM!

LPGhatguy:
(Workflow issue is known; I'll be updating our clippy version to fix this)

PR #199, opened 2021-07-16
Implement Tags in rbx_types, rbx_binary, and rbx_xml

kennethloeffler:
There is currently no special representation for `CollectionService` tags in rbx-dom. To address this, this PR adds the new type `Tags` to rbx_types. It is a wrapper over `Vec<String>` that provides conversions between byte arrays and `Tags`. It changes the type of `Instance.Tags` from `BinaryString` to `Tags` and makes rbx_binary and rbx_xml write to and read from this property using `Tags`. This makes it possible for consumers like Rojo to expose a more friendly interface for `CollectionService` tags rather than requiring users to provide a base64 string.


rbx_xml/src/types/tags.rs
```diff
@@ -0,0 +1,60 @@
...
+        if !contents.is_empty() {
+            writer.write(XmlWriteEvent::cdata(&base64::encode(&contents)))?;
+        }
+
+        Ok(())
+    }
+
+    fn read_xml<R: Read>(reader: &mut XmlEventReader<R>) -> Result<Tags, DecodeError> {
+        let value = reader.read_base64_characters()?;
+        Ok(value.try_into().unwrap_or_default())
```
kennethloeffler, line 28: I wanted to use `map_err` and the question mark operator here but couldn't get it to compile :/

rbx_types/src/tags.rs
```diff
@@ -0,0 +1,108 @@
...
+impl Tags {
+    pub fn new() -> Tags {
+        Self {
+            tag_list: Vec::new(),
+        }
+    }
+
+    pub fn as_vec(&self) -> &Vec<String> {
+        &self.tag_list
+    }
```
LPGhatguy, line 24: You never want to return `&Vec<T>` in Rust. How about `fn as_slice(&self) -> &[String]`?

rbx_types/src/tags.rs
```diff
@@ -0,0 +1,108 @@
...
+impl From<Vec<&str>> for Tags {
+    fn from(tag_list: Vec<&str>) -> Tags {
+        Self {
+            tag_list: tag_list
+                .iter()
+                .map(|tag_name| String::from_str(tag_name).unwrap())
+                .collect(),
+        }
+    }
+}
```
LPGhatguy, line 46: Not sure we need this; users can convert to `Vec<String>` first themselves:

```rust
tags.into_iter().map(String::from).collect()
```

If we wanted this, we would be better off implementing `FromIterator` to get a bunch of containers for free.

rbx_types/src/tags.rs
```diff
@@ -0,0 +1,108 @@
...
+        let mut buf: Vec<u8> = Vec::with_capacity(tag_list.capacity() + tag_list.len());
+
+        for tag_name in tag_list.as_slice() {
+            buf.extend_from_slice(tag_name.as_bytes());
+            buf.extend_from_slice(b"\0")
+        }
+
+        buf
+    }
+}
```
LPGhatguy, line 60: This feels pretty weird to have as part of `From`. If we want this method, we should write an explicit method for it.

This can also be a one-liner, which makes me think we should consider writing that inline when we need it in rbx_xml and rbx_binary:

```rust
fn encode_tags(tags: &[&str]) -> Vec<u8> {
    tags.join("\u{0000}").into_bytes()
}
```

rbx_types/src/tags.rs
```diff
@@ -0,0 +1,108 @@
...
+                    *element != 0
+                })
+                .collect();
+
+            tag_list.push(String::from_utf8(tag_name)?);
+        }
+
+        Ok(Self { tag_list })
+    }
+}
```
LPGhatguy, line 86: You can use [`slice::split`](https://doc.rust-lang.org/std/primitive.slice.html#method.split) and save a lot of effort here.

Similar to the `From` implementation, this should probably be its own method instead of a generic trait implementation.

rbx_types/src/tags.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::string::FromUtf8Error;
+
+#[derive(Default, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(transparent)
+)]
+pub struct Tags {
+    tag_list: Vec<String>,
```
LPGhatguy, line 12: This could have a general name like `members`.

kennethloeffler:
rbx_xml has no great way to discern if a `BinaryString` in the file is in fact a `Tags` or not. It should start dealing with property descriptors in a way that's closer to rbx_binary (which handles this just fine). This work is blocked until then.

kennethloeffler:
Never mind on that - we can add a conversion from `BinaryString` to `Tags`, which makes the `Tags` XML reader/writer totally unnecessary. However, rbx_xml could probably still use a refactor at some point, especially with respect to its tests: in particular, I think it would be good to start using the files from the test-files submodule.

kennethloeffler:
Simply adding the conversion wasn't enough for rbx_xml to write `Tags` as `BinaryString`. This change fixes that, but it's ugly :(

PR #204, opened 2021-08-05
Patch WeldConstraint properties in reflection database

kennethloeffler:
Adds aliases for `WeldConstraint.Part0Internal` and `WeldConstraint.Part1Internal` to address rojo-rbx/rojo#458. This PR also fixes a small casing issue in rbx_dom_lua's `findCanonicalPropertyDescriptor`.


kennethloeffler:
The build is failing because the criterion crate's MSRV was recently bumped to 1.46.0 (and `matches!` was stabilized in 1.42.0, which is the specific reason for the failure), while ours is still 1.40.0. Should we increment the MSRV?

LPGhatguy:
If you rebase onto master right now, things should be good.

LPGhatguy (approved):

PR #205, opened 2021-08-08
Add custom reader/writer for CollisionGroups in rbx_dom_lua

kennethloeffler:
This PR adds a custom reader and writer for `Workspace.CollisionGroups` to rbx_dom_lua. This property causes the Rojo plugin to error during live-sync for some users who use Remodel to copy `Workspace` into a Rojo project.

Similar to `Tags`, all the parsing that this implementation does can and should be removed once we have a better representation for `CollisionGroups` in rbx_types.


kennethloeffler:
Decided to go ahead and implement CollisionGroups in rbx_types. I'll circle back around to this PR once that's in shape..

PR #206, opened 2021-08-11
Implement CollisionGroups in rbx_types

kennethloeffler:
This PR implements `CollisionGroups` in rbx_types. An entry in `CollisionGroups` consists of a name, an ID, and a mask that indicates the collidablilty of the entry with other groups. The serde representation of this type uses an optional list of names of non-collidable groups instead of the mask, as follows:
```json
"CollisionGroups": [
  {
    "name": "AGroup",
    "id": 1
  },
  {
    "name": "AnotherGroup",
    "id": 2,
    "notCollidableWith": [ "AGroup", "Default" ]
  }
]
```
A group without a `notCollidableWith` field is collidable with all other groups (assuming it is not named in any other group's `notCollidableWith`), similar to the behavior in Roblox where a newly created collision group is by default colllidable with all other groups.

I decided to hard code the default collision group into the serde serializer/deserializer because in Roblox it always exists, must be named `Default`, and must have an ID of `0`. An entry for the default collision group is thus not required and `Default` can always be used in `notCollidableWith`.


kennethloeffler:
Thinking about this again due to https://github.com/rojo-rbx/rojo/issues/476...

I'm on the fence with this implementation. I'm not happy with its behavior in Rojo's `fmt-project`. Formatting the above example results in this:
```json
"CollisionGroups": [
  {
    "name": "Default",
    "id": 0,
    "notCollidableWith": [ "AnotherGroup" ]
  },
  {
    "name": "AGroup",
    "id": 1,
    "notCollidableWith": [ "AnotherGroup" ]
  },
  {
    "name": "AnotherGroup",
    "id": 2,
    "notCollidableWith": [ "AGroup", "Default" ]
  }
]
```
There's some redundancy in the `notCollidableWith` field and the omitted `Default` collision group appears. I'm considering removing the hard-coded `Default` group, but omitting this group or giving it an ID other than 0 results complicates rbx-dom-lua's `CollisionGroups` implementation and also breaks several Roblox Studio builtin plugins when built into a place file. This seems to be a pretty big footgun that we ought to protect users from. Is there a better way to represent `CollisionGroups` that avoids these problems?

GlowingUmbreon:
> Is there a better way to represent `CollisionGroups` that avoids these problems?

Maybe you could just do this?
```json
[
  { // ID: 1
    "name": "AGroup",
    "notCollidableWith": [ "AnotherGroup" ]
  },
  { // ID: 2
    "name": "AnotherGroup",
    "notCollidableWith": [ "AGroup", "Default" ]
  }
]
```
In this example the id's are generated based on the index of the array (most people don't need to worry about the ID of collision groups), if a game is using a model with a specific CollisionGroupId this may not be ideal but as long as they are not using ID 1000 or something it should be fine. 

If a person wants to change the default group they could do.
```json
[
  { // ID: 0
    "name": "Default",
    "notCollidableWith": [ "AnotherGroup" ]
  },
  { // ID: 1
    "name": "AGroup",
    "notCollidableWith": [ "AnotherGroup" ]
  },
  { // ID: 2
    "name": "AnotherGroup",
    "notCollidableWith": [ "AGroup", "Default" ]
  }
]
```

I'm not entirely sure if this is a good solution as I do not know about the internals of rojo or the workflows of other users, but this would work in my situation.

LPGhatguy:
Why not keep the default collision group in the representation, but omit it from the serialized representation completely?

kennethloeffler:
Yeah, I think that works. The redundancies in `notCollidableWith` after a `rojo fmt-project` don't seem worth working around - I guess we can live with it for now.

kennethloeffler:
I have NO idea what's going on with the snapshots on CI. I'll come back to this later, I'm doing stuff in Rojo today

kennethloeffler:
Issue stemmed from a recent change in serde_yaml - my machine was still using the old version!

PR #210, opened 2021-09-17
Add dataType member to rbx_dom_lua's PropertyDescriptor

kennethloeffler:
For Rojo's two-way sync, the plugin needs to be able to get a property's data type from a `PropertyDescriptor` so that it can use `EncodedValue.encode` to encode values destined for the server. Rojo's reconciler used to have a method that was hard coded to do this only for the `String` type, but it was lost after https://github.com/rojo-rbx/rojo/commit/f66860bdfe1b6a5f0174465c0a6c964b56e304cb.

### DIFF
--- a/rbx_dom_lua/src/PropertyDescriptor.lua
+++ b/rbx_dom_lua/src/PropertyDescriptor.lua
@@ -20,10 +20,21 @@ local function set(container, key, value)
 end
 
 function PropertyDescriptor.fromRaw(data, className, propertyName)
-	local dataType, ty = next(data.DataType)
+	local key, value = next(data.DataType)
 
 	return setmetatable({
-		dataType = dataType == "Enum" and dataType or ty,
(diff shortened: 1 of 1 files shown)


rbx_dom_lua/src/PropertyDescriptor.lua
```diff
@@ -20,7 +20,10 @@ local function set(container, key, value)
 end
 
 function PropertyDescriptor.fromRaw(data, className, propertyName)
+	local dataType, ty = next(data.DataType)
```
LPGhatguy, line 23: I think we should rename these variables and leave a comment as this is unintuitive.

The first value is only _actually_ `dataType` if it's equal to `Enum`, so we should just call it `key`, and the second value `value`. We should also mention in a comment about the two cases we're expecting:
- `{ Enum = "<name of enum>" }`
- `{ Value = "<data type>" }`

LPGhatguy (approved):

PR #214, opened 2022-03-07
Add documentation for ProtectedString (PROP 29)

metatablecat:
Added documentation for the ProtectedString prop type.

This property was exposed to us when Roblox released a collection of RBXMs with bytecode in them (then again, the prop type had been accesible since Roblox started shipping the Client scripts with the same PROP type)

image

Generally, there's no real usecase to actually exposing this into the tool itself, but I've added it here for reference anyway. As far I'm aware, the only place where this is usable is with plugins.


metatablecat:
Closed due to weird wording, will probably update down the line

PR #215, opened 2022-03-15
Fix PackageLink binary serialization

st0nerhat:
When serializing a DOM containing a PackageLink to binary via remodel or rojo, it results in this error: `Unsupported property type: PackageLink.PackageId is of type Content`.

It appears the fix is to just add a fallback for the Content type.


LPGhatguy (approved):
Thank you, this is a great and simple PR!

PR #216, opened 2022-04-18
chore: Update reflection database and create exemptions for new unimplemented DataTypes

unix-system:


.gitignore
```diff
@@ -1,11 +1,8 @@
 # Cargo stuff
-/target
+**/target
```
unix-system, line 2: This change was created because if you built any of the subordinate crates, such as gen_reflection, it would be included by Git.

.gitignore
```diff
@@ -1,11 +1,8 @@
...
 **/*.rs.bk
 Cargo.lock
 
-# Criterion puts benchmark files into a target folder in each individual crate.
-/rbx_binary/target
```
unix-system, line 7: Due to change on L2, no longer necessary

unix-system:
@LPGhatguy Sorry about the delay here - no longer marked as a draft!

LPGhatguy:
Hey, thanks for the PR!

I ended up doing a couple of these changes as separate commits so that they were a little easier to reason about.

How did `target` folders end up getting created in places other than the root folder for you? Even when I cd into `rbx_binary` and run `cargo test` or `cargo build`, it uses the target folder from the root of the project because it's in a workspace.

unix-system:
@LPGhatguy Sorry about the delay on this one .. I'm not sure how I managed that one! Probably just an accident.

I'll close the PR since it looks like each has indeed been replaced by other PRs that are now merged 👍

PR #218, opened 2022-04-20
Preparations for rojo to support Tags/Attributes.

MaximumADHD:
- Attributes are now written in rbx_xml.
- Tags are no longer written if the container is empty.
- Attributes now have pods implemented in rbx_dom_lua.
- Improved pods for inf/nan values in Float32 and Float64.

### DIFF
--- a/rbx_xml/src/types/mod.rs
+++ b/rbx_xml/src/types/mod.rs
@@ -7,6 +7,7 @@
 //! 2. Add a 'mod' statement immediately below this comment
 //! 3. Add the type(s) to the declare_rbx_types! macro invocation
 
+mod attributes;
 mod axes;
 mod binary_string;
 mod bool;
@@ -48,6 +49,7 @@ use crate::{
 };
 
 use self::{
+    attributes::write_attributes,
     referent::{read_ref, write_ref},
     shared_string::{read_shared_string, write_shared_string},
     tags::write_tags,
@@ -99,6 +101,7 @@ macro_rules! declare_rbx_types {
         ) -> Result<(), EncodeError> {
             match value {
                 $(Variant::$variant_name(value) => value.write_outer_xml(xml_property_name, writer),)*
+                Variant::Attributes(value) => write_attributes(writer, xml_property_name, value),
 
                 // BrickColor values just encode as 32-bit ints, and have no
                 // unique appearance for reading.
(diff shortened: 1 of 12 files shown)


rbx_xml/src/types/tags.rs
```diff
@@ -14,11 +14,13 @@ pub fn write_tags<W: Write>(
...
-    let encoded = value.encode();
-
+    let encoded = base64::encode(&value.encode());
     writer.write(XmlWriteEvent::start_element(XML_TAG_NAME).attr("name", property_name))?;
-    writer.write_string(&base64::encode(&encoded))?;
-    writer.write(XmlWriteEvent::end_element())?;
 
+    if !encoded.is_empty() {
+        writer.write_string(&base64::encode(&encoded))?;
+    }
```
LPGhatguy, line 22: It looks like this code is base64-encoding the result twice.

rbx_xml/src/types/mod.rs
```diff
@@ -99,6 +101,7 @@ macro_rules! declare_rbx_types {
         ) -> Result<(), EncodeError> {
             match value {
                 $(Variant::$variant_name(value) => value.write_outer_xml(xml_property_name, writer),)*
+                Variant::Attributes(value) => write_attributes(writer, xml_property_name, value),
```
LPGhatguy, line 104: We should implement Attributes the same way that we implement other types. I don't think there's any reason to special-case them here. Look at how other types (not `Tags`!) are implemented, via `XmlType` and `write_outer_xml`.

LPGhatguy (changes requested):
Thanks for the PR!

This PR does a few things that aren't necessarily related. Can you split this PR into one PR for each of these things? We generally test each piece separately, and often times we don't need to block each piece on reviewing the others.

- rbx_xml changes
- rbx_dom_lua changes
- One PR per other unrelated chang (float serialization, any changes to rbx_binary)

rbx_xml/src/types/mod.rs
LPGhatguy: Combing back through this code, I remember why we had to implement Tags and Attributes this way, ugh. rbx_xml pre-dates most of the work we did on reflection to get rbx_binary functional, and so it wasn't built with types like Tags in mind.

MaximumADHD: I saw you made another pull request for this. Should I remove my integration?

MaximumADHD:
Seems to have been taken care of in https://github.com/rojo-rbx/rbx-dom/pull/228
Merge conflicts are busted, so I'll just close this now.

PR #223, opened 2022-04-25
Use better strategy for serializing ±inf/nan in serializeFloat function.

MaximumADHD:
The `tonumber` function handles "inf", "-inf", and "nan" correctly. Numbers passed into `tonumber` just return themselves.
As such, it should be reasonable to use the constants derived from `tostring` to serialize those values into JSON.

### DIFF
--- a/rbx_dom_lua/src/EncodedValue.lua
+++ b/rbx_dom_lua/src/EncodedValue.lua
@@ -11,13 +11,11 @@ local function unpackDecoder(f)
 end
 
 local function serializeFloat(value)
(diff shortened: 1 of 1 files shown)


LPGhatguy:
This PR is only half of the protocol. Encoding from rbx_types (like with the Rojo server) still won't produce strings, so this code will never see strings to decode.

Even so, I'm not sure this is the best approach for solving this problem. A little while back, there was some investment into implementing MessagePack in Luau. Using another format that supports infinity and NaN would be better than introducing strings into JSON to encode the missing values.

PR #232, opened 2022-08-10
fixed Property Descriptor Custom erroring

Boegie19:
property Descriptor would error here `customProperties[self.className][self.name]` since is some ceases `customProperties[self.className]` returns nil what would cause the script to error out.


LPGhatguy (approved):
Looks good to me, thank you!

LPGhatguy:
(CI errors are unrelated, we'll need to solve those in a follow-up PR since it's a clippy + insta interaction)

PR #235, opened 2022-08-12
error will always return an error type added + more asserts for better error messages.

Boegie19:
added if statements to check at every point it can error and made it so that it will always return an error type and does not make the script errror out.


1xozf:
The fix was found by another user.

Dekkonot:
Downstream PR mentions that this is only an issue with older versions of Rojo. Given that, I'm assuming that it was fixed in some other way in the last few months. Feel free to re-open this if it's still a problem in the newest release but as-is, I'm going to go ahead to close this.

Boegie19:
Fair enough

PR #239, opened 2022-10-12
safeguard for Plugins folder not existing inside the studio directory during reflection generation

blackshibe:


LPGhatguy:
Looks like you picked up too many changes in this branch; I recommend running `git checkout master -- FILE` for each file that you want to revert back to master.

Issue #222, opened 2022-04-25
Implement UniqueId type


Anaminus:
UniqueId is a property type that is currently present only in place files. It has distinct formats for both rbxl and rbxlx.

Sample property chunk for rbxl format, in hexadecimal format:

```text
| 00 00 00 00 08 00 00 00  55 6e 69 71 75 65 49 64 |........UniqueId|
| 1f 00 01 b7 86 02 4e 93  e1 e1 1e 4c 04 09 cf 82 |......N....L....|
| 88                                               |.|
```

Sample of equivalent value for rbxlx:

```xml
<UniqueId name="UniqueId">708f260204e7c144024e93e10001b786</UniqueId>
```

The first bytes describe the property as usual:

```text
00 00 00 00              Class ID (0)
08 00 00 00              Length of property name (8)
55 6e 69 71 75 65 49 64  Property name ("UniqueId")
1f                       Property type
```

The type ID for UniqueId is **0x1F**.

As usual, the remaining bytes represent an array of values. For this type, **the array is interlaced by 16 bytes.**

The sample contains one value:

	00 01 b7 86 02 4e 93 e1 e1 1e 4c 04 09 cf 82 88

For proper conversion between rbxlx, this must be interpreted as a struct:

```
{
	Index:  u32
	Time:   u32
	Random: i64
}
```

The Index and Time fields are interpreted as unsigned 32-bit integers, encoded in big-endian. The Random field is similar to the encoding of the int64 type: a signed 64-bit integer in big-endian and with **rotated encoding**.

In the rbxlx format, this value is encoded by converting each field to bytes in big-endian, then encoding to hexadecimal. In relation to the binary format, the fields are written in reverse order:

```text
    Index    Time     Random
bin 0001b786 024e93e1 e11e4c0409cf8288

    Random           Time     Index
xml 708f260204e7c144 024e93e1 0001b786
```

Decoding is simply applying this process in reverse.

Note: The Random field looks different between formats because rbxl uses rotated encoding:

```text
    Random (hex)     Random (binary)
bin e11e4c0409cf8288 1110000100011110010011000000010000001001110011111000001010001000
xml 708f260204e7c144 0111000010001111001001100000001000000100111001111100000101000100
```

Anaminus:
The following archive contains rbxl and rbxlx files with instances that have equivalent UniqueId values. These can be used to verify the correctness of implementations:

[...]

PR #241, opened 2022-10-31
Document UniqueId in the binary format

Dekkonot:
As per Issue #222, I've decided to document UniqueId in preparation for implementing it. 

EDIT: After some feedback in the linked issue, I'm going to go ahead and bite the bullet and document this properly. I would prefer to view UniqueId as a blob but it doesn't make sense unfortunately.


regginator:
(This is definitely a 0x1F moment)

Dekkonot:
> (This is definitely a 0x1F moment)

Spam is spam regardless of when or why you post it.

regginator:
> > (This is definitely a 0x1F moment)
> 
> Spam is spam regardless of when or why you post it.

Well I mean you don't have to be a soulless NPC about a single comment, cmon now? I fail to see how this is spam, I just actually find it interesting that they're adding this as an exclusive rbxm type, lol.

LPGhatguy (approved):

PR #244, opened 2022-11-22
Fix non-determinstic behaviour for shared string dictionary

apotato24:
The current way rbxm files are saved leads to some non-derminsitic behaviour due to how the shared string dictionary is filled. This PR aims to make this correct this behaviour.


apotato24:
Sorry, not sure how I forgot to test it before making the pull request. It should work fine now.

apotato24:
Closing this in favour of #254

PR #248, opened 2023-01-15
Implement Font type

Corecii:
This PR implements the Font type:
* Implements binary de/serializer
* Implements xml de/serializer
* Implements Lua/json de/serializer
* Adds some relevant tests
* Updates reflection database

This is my first time working with rbx-dom's code, so I may have gotten some things wrong! I also have some general questions and concerns:
* The Font value I saw in a hex editor looked different from what rbx-dom would output. Does rbx-dom do some transformation on all binary property data to make them sense? Maybe I didn't read deep enough about it. Should I have included information about that in my documentation of the Font type?
* The Font type has a u8 and u16 which act a lot like enums but aren't exactly enums. I made special types for these with support for an `Other` so that any new items should _just work_ even if they're not nicely represented.  Was this the correct move? I could have represented Weight with just a u16, but Style needs a u8 <-> name converter so that it can be serialized to/from xml.
* The FontStyle type serializes to xml by name, but to binary as u8. I included an `Other` enum variant here since I figured having support for unknown new values was a good thing, _but that was before I saw how xml did it_. This means unknown new FontStyles can go from binary -> binary, but not binary -> xml or xml -> anything. Is this okay, or do we want to remove the `Other` variant here? Related: in cases where the variant is unknown, I default to `Normal`. Would it be better to error?
* The default font style I have listed is unstyled Source Sans Pro. This matches the default when you make a new object in Studio, but _not_ the default when loading a type in Studio that's missing the Font property. Studio will actually give TextLabels (and such) that are missing the Font property and unstyled _Arial_ font. Should that be our default, or should Source Sans Pro be our default?


Corecii:
Oh, I just realized this is also missing conversion from the old font property to this new one, so if a very old rbxm of rbxl is loaded with rbx-dom, its FontFace will be incorrect. What's the correct way to handle that case?

docs/binary.md
```diff
@@ -639,6 +640,21 @@ When interacting with the XML format, care must be taken because `UniqueId` is s
...
+**Type ID `0x20`**
+
+The `Font` type is a struct composed of two `String` values, a `u8`, and a `u16`:
+
+| Field Name   | Format              | Value                           |
+| :----------- | :------------------ | :------------------------------ |
+| Family       | [`String`](#string) | The font family content url     |
+| Weight       | `u16`               | The weight of the font          |
+| Style        | `u8`                | The style of the font           |
+| CachedFaceId | [`String`](#string) | The cached content url of the ttf file |
```
Dekkonot, line 653:
```suggestion
| Field Name   | Format              | Value                                  |
|:-------------|:--------------------|:---------------------------------------|
| Family       | [`String`](#string) | The font family content URI            |
| Weight       | `u16`               | The weight of the font                 |
| Style        | `u8`                | The style of the font                  |
| CachedFaceId | [`String`](#string) | The cached content URI of the TTF file |
```
In the spec files, we prefer to have the delimiter row (the second one) not have leading or trailing spaces, contrary to the rest of the table. Also, the `Value` column should be aligned to the length of the max description, which in this case is the description of `CachedFaceId`.

Additionally, we capitalize acronyms and it's more accurate to refer to `Family` and `CachedFaceId` as URIs, since Roblox's `Content` type isn't necessarily a valid URL.

docs/binary.md
```diff
@@ -639,6 +640,21 @@ When interacting with the XML format, care must be taken because `UniqueId` is s
...
+
+The `Weight` and `Style` fields are stored as little-endian unsigned integers. These are usually treated like enums, and to assign them in Roblox Studio an Enum is used. Interestingly, the `Weight` is _always_ stored as a number in binary and xml, but `Style` is stored as a number in binary and as text in xml.
```
Dekkonot, line 655:
```suggestion
The `Weight` and `Style` fields are stored as little-endian unsigned integers. These are usually treated like enums, and to assign them in Roblox Studio an Enum is used. Interestingly, the `Weight` is *always* stored as a number in binary and XML, but `Style` is stored as a number in binary and as text in XML.
```

By convention, we use `*` for italics in spec files. Also as previously stated, we capitalize acronyms.

docs/binary.md
```diff
@@ -639,6 +640,21 @@ When interacting with the XML format, care must be taken because `UniqueId` is s
...
+
+The `CachedFaceId` field is always present, but is allows to be an empty string (a string of length 0). When represented in xml, this property will be omitted if it is an empty string. This property is not visible via any user APIs in Roblox Studio.
```
Dekkonot, line 657:
```suggestion
The `CachedFaceId` field is always present, but is allows to be an empty string (a string of length `0`). When represented in XML, this property will be omitted if it is an empty string. This property is not visible via any user APIs in Roblox Studio.
```

We surround literals like `0` in backticks.

Dekkonot (changes requested):
In a hurry at the moment so I can't do a more comprehensive review of the actual code, but here's one for the spec file at least.

It generally looks good, it just needs some changes to fit with the format we use throughout the spec files.

rbx_xml/src/types/font.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+impl XmlType for Font {
+    const XML_TAG_NAME: &'static str = "Font";
+
+    fn write_xml<W: Write>(&self, writer: &mut XmlEventWriter<W>) -> Result<(), EncodeError> {
+        write_content(writer, &self.family, "Family")?;
+
+        let weight: u16 = self.weight.clone().into();
+        let weight: i32 = weight.into();
```
Dekkonot, line 95: Provided the earlier suggestion for a `to_u16` method is adopted:
```suggestion
        let weight = i32::from(self.weight.to_u16());
```

rbx_xml/src/types/font.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    fn read_xml<R: Read>(reader: &mut XmlEventReader<R>) -> Result<Self, DecodeError> {
+        let family = read_content(reader, "Family")?;
+
+        let weight: i32 = reader.read_value_in_tag("Weight")?;
+        let weight: u16 = weight
+            .clamp(u16::MIN as i32, u16::MAX as i32)
+            .try_into()
+            .unwrap();
+        let weight = FontWeight::from(weight);
```
Dekkonot, line 120: Provided the suggestion for `FontWeight::from_u16` is adopted:
```suggestion
        let weight = FontWeight::from_u16(weight);
```

Dekkonot (dismissed):
Gave the files a proper look!

This is mostly fine, but there's some cleanup I'd like done with `FontWeight` and `FontStyle`. We'd ideally not use `.into()` or have to clone them, and most of my comments are related to that + cleanup from changing them.

rbx_binary/src/serializer/state.rs
```diff
@@ -709,6 +709,21 @@ impl<'dom, W: Write> SerializerState<'dom, W> {
...
+                    Type::Font => {
+                        for (i, rbx_value) in values {
+                            if let Variant::Font(value) = rbx_value.as_ref() {
+                                chunk.write_string(&value.family)?;
+                                chunk.write_le_u16(value.weight.clone().into())?;
+                                chunk.write_u8(value.style.clone().into())?;
+                                match &value.cached_face_id {
+                                    Some(id) => chunk.write_string(id)?,
+                                    None => chunk.write_string("")?,
+                                }
```
Kampfkarren, line 721: Nit: Would this look nicer?

```suggestion
                                chunk.write_string(value.cached_face_id.unwrap_or_default())?;
```

Corecii: I think so. I had to stick a `.clone()` to make the borrow checker happy, and write_string wants an `&str`, so it's a bit uglier than your suggestion but I think it's still nicer than what I had.

Corecii:
My most last commit addresses all of your comments!

Corecii:
Tests should pass now. It's unrelated to the font implementation -- probably related to updating the reflection database. Since IgnoreGuiInset no longer saves, it isn't present in the decoded or encoded rbxms.

Dekkonot (approved):
Looks good to me!

For anyone curious about the prior CI failure, the serialization of `IgnoreGuiInserts` was changed in release 554 and updating the reflection database exposed us to that change. Thus (as expected) the test involving ScreenGuis failed.

We likely should try to fix this in the future so that Roblox reflection changes don't interfere with our tests like this, but that's well outside of scope for this PR.

Dekkonot:
I meant to leave a comment a while back answering your questions but I forgor and got caught in the review sauce, so you'll have to forgive me.

> The Font value I saw in a hex editor looked different from what rbx-dom would output. Does rbx-dom do some transformation on all binary property data to make them sense? Maybe I didn't read deep enough about it. Should I have included information about that in my documentation of the Font type?

Roblox serializes their binary format with compression (every chunk is LZ4 compressed), but rbx-dom doesn't at the moment. That's the only difference I can think of.

> The Font type has a u8 and u16 which act a lot like enums but aren't exactly enums. I made special types for these with support for an Other so that any new items should just work even if they're not nicely represented. Was this the correct move? I could have represented Weight with just a u16, but Style needs a u8 <-> name converter so that it can be serialized to/from xml.

I considered this while I was doing the review, and eventually decided that it was a good call. It doesn't really impact things for rbx-dom either way, but using enums should make it easier for Rojo to implement a good format for Font property types.

> The FontStyle type serializes to xml by name, but to binary as u8. I included an Other enum variant here since I figured having support for unknown new values was a good thing, but that was before I saw how xml did it. This means unknown new FontStyles can go from binary -> binary, but not binary -> xml or xml -> anything. Is this okay, or do we want to remove the Other variant here? Related: in cases where the variant is unknown, I default to Normal. Would it be better to error?

A younger less experienced version of me would say "we should error whenever we encounter unknown data" but as I've seen rbx-dom and Rojo get surprised by things, I'm inclined to say it's not a bad idea to maintain data when we can. The XML limit is a shame but isn't our fault because there's quite literally nothing we could do about it either way. At the very least this implementation will let binary files do what they need to do.

> The default font style I have listed is unstyled Source Sans Pro. This matches the default when you make a new object in Studio, but not the default when loading a type in Studio that's missing the Font property. Studio will actually give TextLabels (and such) that are missing the Font property and unstyled Arial font. Should that be our default, or should Source Sans Pro be our default?

Anybody who cares about Font will be overriding it so it isn't a big deal and I'd rather use what Roblox's default is because it's probably what people will be expecting anyway. Not sure if this is what we do *everywhere* but I wouldn't be surprised if the same reasoning was used.

rbx_xml/src/types/font.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    fn read_xml<R: Read>(reader: &mut XmlEventReader<R>) -> Result<Self, DecodeError> {
+        let family = read_content(reader, "Family")?;
+
+        let weight: i32 = reader.read_value_in_tag("Weight")?;
+        let weight: u16 = weight
+            .clamp(u16::MIN as i32, u16::MAX as i32)
+            .try_into()
+            .unwrap();
```
LPGhatguy, line 118: We have support for adding any integer types we need to this method; we don't need to cast around. We should add an implementation of `XmlType` for `u16` like we do for `i16` here: <https://github.com/rojo-rbx/rbx-dom/blob/3a3563eeebebca949189b7a7f3002bedf3c9e017/rbx_xml/src/types/numbers.rs#L73>

LPGhatguy (changes requested):
Thank you for the PR! Here are a few comments from me, then this change looks good to me.

LPGhatguy:
The compatibility story for this property type might be a bit hairy. When merging older models without `Font` with newer models that do have the property, which property takes precedence in Studio? Will users need to re-save all of their models? Should we offer some sort of normalization or migration pass?

@Dekkonot @Corecii

> Roblox serializes their binary format with compression (every chunk is LZ4 compressed), but rbx-dom doesn't at the moment. That's the only difference I can think of.

rbx_binary does output compressed chunks. The compression code is here:

<https://github.com/rojo-rbx/rbx-dom/blob/3a3563eeebebca949189b7a7f3002bedf3c9e017/rbx_binary/src/chunk.rs#L79-L102>

and is referenced all over in the serializer:

<https://github.com/rojo-rbx/rbx-dom/blob/3a3563eeebebca949189b7a7f3002bedf3c9e017/rbx_binary/src/serializer/state.rs#L443>

> The default font style I have listed is unstyled Source Sans Pro. This matches the default when you make a new object in Studio, but not the default when loading a type in Studio that's missing the Font property. Studio will actually give TextLabels (and such) that are missing the Font property and unstyled Arial font. Should that be our default, or should Source Sans Pro be our default?

We should match the default value for what Roblox does for creating the instance with `Instance.new`, not the value when inserted via Studio's UI. That matches what happens when the `Font` property is unset, Arial. We shouldn't deviate from the strategy of using the instance's defaults instead of Studio's nice defaults since it can mask problems when things are broken. In any case, rbx_reflection should pick up the new property and property type and make the default value (Arial) explicit anyways.

Dekkonot:
> rbx_binary does output compressed chunks. The compression code is here:

That is rather embarrassing on my part! I thought I'd looked properly and I don't really have an excuse because I did in fact know that... I literally wrote the code for serializing `SSTR`. 🤦🏼 Thank you for the correction. In that case, I'm not sure how to explain the difference between rbx-dom's output and Studio's beyond maybe the compression level we use.

> When merging older models without `Font` with newer models that do have the property, which property takes precedence in Studio? Will users need to re-save all of their models? Should we offer some sort of normalization or migration pass?

Every value of the Font enum has an equivalent `Font` value with the exception of `Unknown`, which doesn't do anything when Studio reads it. This means that we can directly translate the Font enum to `Font` values when we have to. This would have to be done via a migration pass, but it could be done in the background because it isn't destructive.

Corecii:
I'm going to be away from home for the next week and may not have time to implement fixes it changes to this PR in that time. I'll get to it when I'm back, or _maybe_ if I have some downtime while I'm gone! Just want to let you all know beforehand so you don't think I've forgotten or abandoned this PR.

LPGhatguy:
No rush on addressing feedback! These days I'm not around consistently. I've been working on trying to get myself out of the pipeline and find groups of maintainers for all this stuff I'm still the main blocker on.

> That is rather embarrassing on my part! I thought I'd looked properly and I don't really have an excuse because I did in fact know that... I literally wrote the code for serializing SSTR. 🤦🏼 Thank you for the correction. In that case, I'm not sure how to explain the difference between rbx-dom's output and Studio's beyond maybe the compression level we use.

I figure that's probably it. There are usually many ways to compress the same data, and LZ4 is tunable.

> Every value of the Font enum has an equivalent Font value with the exception of Unknown, which doesn't do anything when Studio reads it. This means that we can directly translate the Font enum to Font values when we have to. This would have to be done via a migration pass, but it could be done in the background because it isn't destructive.

This is really good news. I wonder if this is something we should plan to do in rbx_binary if we get any user reports of weird behavior. Tools like Remodel and Rojo that aggregate models might produce some funny results if we don't do anything.

Dekkonot:
> This is really good news. I wonder if this is something we should plan to do in rbx_binary if we get any user reports of weird behavior. Tools like Remodel and Rojo that aggregate models might produce some funny results if we don't do anything.

If we do it in rbx_binary, we're essentially promising to do it a second time if anything like this happens again. I don't love that idea and would instead prefer we leave it up to consumers to patch files. This would let them decide how to handle it themselves and let them patch properties we don't/can't. However, if we do decide to implement it, I'd prefer we do it with a crate that wasn't rbx_binary.

The idea of a codec that modifies data in that fashion feels wrong even if the intent is to fix something Roblox did. It should be done by someone else, regardless of it it's another rbx_dom crate or an upstream user.

NecroticNanite:
Just checking if anyone is still looking into this, and if it will be released soon?
This is going to be a big problem for us very soon if not resolved

Corecii:
I'm planning to respond to feedback this week.

The code already works and you can build Rojo with this merged in for font support. We have pre-built versions of Rojo and Remodel under the UpliftGames org which you can refer to with aftman if you need it now and don't want to build it yourself.

I didn't _want_ to let this sit around for 2+ weeks, but I've been really busy lately and haven't been able to get back to this PR. I should have time within the next two days to address feedback!

LPGhatguy:
@Dekkonot

> The idea of a codec that modifies data in that fashion feels wrong even if the intent is to fix something Roblox did. It should be done by someone else, regardless of it it's another rbx_dom crate or an upstream user.

Was thinking about this and a memory popped into my head — we've done this before! Colors on parts are something that we do conversions for, from the old `BrickColor` to the new `Color3`-based properties. It should be a matter of just marking one property as canonical and implementing the correct conversions.

rbx_types/src/font.rs
```diff
@@ -0,0 +1,132 @@
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "camelCase")
```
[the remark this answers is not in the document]
Corecii: I think it makes sense for these not to be camelCase. 👍

rbx_types/src/font.rs
```diff
@@ -0,0 +1,132 @@
...
+    Thin,
+    ExtraLight,
+    Light,
+    Regular,
+    Medium,
+    SemiBold,
+    Bold,
+    ExtraBold,
+    Heavy,
+    Other(u16),
```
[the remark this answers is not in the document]
Corecii: I just cut the Other variants. Preserving these other values for binary is nice but seems like an unnecessary pain at this point.

Corecii:
All of your feedback should be addressed now!

> In any case, rbx_reflection should pick up the new property and property type and make the default value (Arial) explicit anyways.

Does this mean the default trait does not need to be changed to reflect Roblox? Should we still change it to match?

NecroticNanite:
Hi, can someone explain how to link our local aftman.toml file to the WIP Uplift games? I tried a variety of URL formats and didn't have much success

Corecii:
> Hi, can someone explain how to link our local aftman.toml file to the WIP Uplift games? I tried a variety of URL formats and didn't have much success.

Our fork versions are one version ahead plus a prerelease tag. Here's what our rojo and remodel entries look like:
```toml
[tools]
rojo = "UpliftGames/rojo@7.2.2-uplift.release.4"
remodel = "UpliftGames/remodel@0.12.0-uplift.release.9"
```
Don't forget to switch back to `rojo-rbx/rojo` when the Font fixes are merged in to the main repo. The versions above are from the very first version of this PR -- which works and is compatible, but is outdated and is missing some non-critical fixes.

NecroticNanite:
@Corecii I tested changing my aftman file to what you have above, but I still encounter the same issue where fonts are not synced
Output file has :
`				<Font name="FontFace">
					<Family><url>rbxasset://fonts/families/GothamSSm.json</url></Family>
					<Weight>700</Weight>
					<Style>Normal</Style>
				</Font>`
But I still get arial in studio after deleting and re-importing via Rojo

NecroticNanite:
Sorry to bump again, but is there an update on when this might get merged? Or an answer to my above question as to why the forked version still exhibits the same issue?

Kampfkarren (dismissed):
Small nits

Corecii:
Nits fixed!

---

@NecroticNanite thanks for this report! rbx_dom had been set to exclude Font types from the reflection database and I didn't catch that when implementing. Additionally, we must manually include the new version of rbx_dom in the plugin in the rojo repo. We didn't catch this at Uplift because we're not syncing in models with rojo -- we only needed updated rbx_dom for remodel. The fork should work now if you change the version to `rojo = "UpliftGames/rojo@7.2.2-uplift.release.5"` and install the updated plugin through either the rojo menu in vs code or the `rojo plugin install` command.

NecroticNanite:
Thanks for fixing that - however, I cannot download that version :(
Aftman gives me the following: "Aftman error: Could not find release UpliftGames/rojo@7.2.2-uplift.release.5"
Installing previous versions works fine (tried .2 and .4)
I'm guessing not published somehow?

Corecii:
@NecroticNanite You're correct! Again, a symptom of us not needing this (yet). I had GitHub generate the release draft but didn't go press publish on it when it was done. Try now!

NecroticNanite:
@Corecii Thanks a lot, that seems to have done the trick! Good timing too, we've got a tricky merge coming up and a release to sync to different environments :D

NecroticNanite:
@Corecii  Hi, we're just testing this now.
First issue:
Old files (from probably 6-12 months ago), that have not been re-exported, will prevent Rojo starting:
The file has a Font XML as follows:
`<token name="Font">9</token>
<Font name="FontFace"></Font>`

It will throw the following error when trying to start Rojo:
`[ERROR rojo] Malformed rbxm file: [path]/[to]\Conversation.rbxmx
        
        Caused by:
            line 139, column 27: Unexpected XML event EndElement(Font)`
            
 Not sure how easy it will be to add back-compat for this - if it's not possible, will likely make this MR go from a minor hotfix into a full major version release, as anyone using it will have to do a fresh export of any model/rbxmx files.
 
 The fix (at least, it seems to fix it) is to delete the blank <Font name="FontFace"></Font> lines from any rbxmx files - possibly the parser should just ignore those empty blocks?
 
 (We're testing a few other things, and will send any issues along as/when we hit them)

Corecii:
Oh that's fascinating! I'm sure we can just ignore it if it's empty. Do you happen to have any files you're willing to share which I can use to test a fix for this locally?

NecroticNanite:
> Oh that's fascinating! I'm sure we can just ignore it if it's empty. Do you happen to have any files you're willing to share which I can use to test a fix for this locally?

Conversation.rbxmx - Pastbin: [https://pastebin.com/qhAUyGzP](url)
Password: xbL1NXXtUc

NecroticNanite:
Everything else seems to be working great though!

LPGhatguy:
Updated the MSRV on master, should fix the CI build.

LPGhatguy (approved):
LGTM. I'm going to put in a follow-up commit to update the signatures of `from_u8` and `from_u16` for the sub-types of `Font`.

PR #221, opened 2022-04-24
Production version of generate_reflection

LPGhatguy:
This PR introduces a new tool, `rbx_reflector`, that is a production-quality rewrite of `generate_reflection`/`gen_reflection` that is suitable for CI and end users.

This tool has a much better interface for generating all sorts of reflection information and will let users manage their own reflection databases and patches when they encounter issues with Rojo or other tools.

PR #249, opened 2023-01-19
Create polished version of `generate_reflection` named rbx_reflector

nezuo:
This is a continuation of the #221 PR. I thought I would start a draft so I could get feedback.


nezuo:
I've added the `DefaultsPlace` command but there are still a few things I would like to fix:

1. The old `generate_reflection` used `innerput` to save the place automatically but that wasn't working for me. I think this will be necessary for a CI workflow.

2. Since it calls kill on the studio process `studio_process.kill()?;`, it leaves behind a lock file. I could delete the file but I was wondering if there is a way to stop the process more gracefully.

nezuo:
Is there a rule on comment wrapping?

nezuo:
I added the `Generate` which for now only takes an API dump and outputs a serialized `ReflectionDatabase`. I changed the API of the `Dump` command to output the API dump instead of the reflection database. I'm not sure what the API was supposed to look like so I'll focus on getting everything functional.

nezuo:
I'm not using the latest version of `serde_yaml` to deserialize the patch files because 0.9 changes the way enums are handled:
https://github.com/dtolnay/serde-yaml/releases/tag/0.9.0

nezuo:
I'm going to open this up for review because I would like some feedback on a few things.

1. `innerput` needs to be fixed.
2. `rbx_xml` and `rbx_binary` need a reflection database for their tests. Should I just generate a serialized reflection database to use for the tests? `rbx_util` also needs a reflection database.
3. What should I do about the lock file?
4. Should I use the log crate?
5. Should it be up to the user to delete files they don't need anymore like the api dump or the defaults place?
6. What was the Patch command supposed to do?

kennethloeffler:
> innerput needs to be fixed.

Elaborate on how it's not working? You could be running into one of innerput's failure cases. It fails when:
* An application using the Windows Console terminal emulator is focused (i.e. cmd.exe, powershell.exe [note that in recent releases of Windows, these use the newer Windows Terminal, which does not have this problem])
* The start menu is focused
* Some other Fun Windows Things that I don't remember at the moment (I really should have written them down)...

Since I wrote it, it has worked reliably for me as long as I avoid these cases, so either I messed up in the unsafe, or one of these things is happening. These problems must be fixed when we expose rbx_reflector to end users (can't wait 😔), but I think it can work for CI. Please leave as much information as possible and feel free to get in touch with me over Discord or email if you want to have a chat about it - it's important that this works as expected

> rbx_xml and rbx_binary need a reflection database for their tests. Should I just generate a serialized reflection database to use for the tests? rbx_util also needs a reflection database.

I think rbx_xml, rbx_binary, rbx_util can stay as-is and we'll just use rbx_reflector to generate the static databases in rbx_dom_lua and rbx_reflection_database, [same as we've been doing with generate_reflection](https://github.com/rojo-rbx/rbx-dom/blob/3a3563eeebebca949189b7a7f3002bedf3c9e017/gen-reflection). I think changing everything over to support dynamic databases is out of scope for this PR and it would be better to tackle in a future PR

> What should I do about the lock file?

I'm not totally sure on best practice for Cargo.lock in workspaces that include both libraries and binaries either. We can take a look at other large projects that need to think about this ([ripgrep](https://github.com/BurntSushi/ripgrep/blob/44fb9fce2c1ee1a86c450702ea3ca2952bf6c5a7/Cargo.toml) for example, which provides a binary as well as libraries, sets a binary target in Cargo.toml, and checks Cargo.lock into vc).

> Should I use the env_logger crate?

I think we should, the configurable logs in generate_reflection have come in handy many times for me

> Should it be up to the user to delete files they don't need anymore like the api dump or the defaults place?

generate_reflection puts these in a temp directory that gets deleted when we're done with it - I think we can do the same here

> What was the Patch command supposed to do?

I'm not sure! Maybe it was supposed to apply patches to a given database in isolation? I can see this being useful, sometimes you want to separately run and inspect the results of each step of the generation process

nezuo:
> Elaborate on how it's not working? You could be running into one of innerput's failure cases. It fails when:

It's failing to find the window for the studio process.
https://github.com/kennethloeffler/innerput/blob/5d81e78cc4a666ba31f57662b266e4655cb8a6a5/src/win32/window.rs#L210

I should have been more specific about the lock file.  I was referencing my earlier comment here:
> Since it calls kill on the studio process studio_process.kill()?;, it leaves behind a lock file. I could delete the file but I was wondering if there is a way to stop the process more gracefully.

>generate_reflection puts these in a temp directory that gets deleted when we're done with it - I think we can do the same here

`rbx_reflector` is split up into multiple commands, if I want to generate a reflection database, I have to pass in file paths for the api dump, patches, etc. That's hard to do if they are put in a temp directory and it makes it hard for the user to inspect the result of each command. Would it be reasonable to expect the user to just add these files to their `.gitignore`?

Dekkonot:
> * `innerput` needs to be fixed.

Just to check in, is this still a problem? If it is, we may need to either steal the implementation from `generate_reflection` and work from there or defer to Ken for help with fixing it.

> * `rbx_xml` and `rbx_binary` need a reflection database for their tests. Should I just generate a serialized reflection database to use for the tests? `rbx_util` also needs a reflection database.

I agree with Ken here; this tool is meant to replace `generate_reflection`, so we don't need to worry about it quite yet. We can keep generating databases and vendor them manually like we have been.

> * What should I do about the lock file?

The elegant way would probably be to use `innerput` to save and exit the process so we don't leave the lock file at all. Pending that, there's really only two routes forward: save the file in the output location and delete it; or save the file to a temp location and copy it, avoiding the lock issue altogether. I think the latter is more 'elegant' but neither of them are good.

> * Should I use the env_logger crate?
Yes please!

> * Should it be up to the user to delete files they don't need anymore like the api dump or the defaults place?

> * What was the Patch command supposed to do?

There's no way of knowing for sure what the intent was but my guess is that the idea was to allow patching of an existing database. This'd be useful for incrementally updating an existing database without needing Studio, which might be nice for platforms that don't support it.

nezuo:
> Just to check in, is this still a problem?

Not anymore, I was able to fix it.

kennethloeffler:
> The elegant way would probably be to use innerput to save and exit the process so we don't leave the lock file at all.

I tried this ([hence the existence of this commit](https://github.com/kennethloeffler/innerput/commit/e454b6102ea7a4848ea55efe7580845618987d1d)), but it seems like the defaults place crashes Roblox Studio when it's closed (possibly due to a specific instance type? dunno 😅)

Dekkonot:
> I tried this ([hence the existence of this commit](https://github.com/kennethloeffler/innerput/commit/e454b6102ea7a4848ea55efe7580845618987d1d)), but it seems like the defaults place crashes Roblox Studio on close (possibly due to a specific instance type? dunno 😅)

@kennethloeffler It feels like maybe we should track down what's causing that issue because it'd make for an interesting bug report but... That's irrelevant to this PR, I guess. Shame it causes a crash. :(

@nezuo As per the discussion on Discord, I think we should just focus on generating a database cleanly. The `patch` command (whatever it was meant to be) can come later since right now it's not clear what functionality we actually need. Let's not get caught up in design and instead just make something that's a production ready version of `generate_reflection`.

rbx_reflector/src/cli/defaults_place.rs
```diff
@@ -0,0 +1,247 @@
...
+impl<'a> Drop for PluginInjector<'a> {
+    fn drop(&mut self) {
+        log::info!("Uninstalling Studio plugin");
+
+        let plugin_path = self
+            .studio_install
+            .plugins_path()
+            .join("RbxDomDefaultsPlacePlugin.lua");
+
+        fs::remove_file(plugin_path).unwrap();
```
Dekkonot, line 245: Panicking in `Drop` implementations isn't a good idea; it'll likely just abort the program because it can't unwind the panic (the `PanicInfo` it creates will get dropped immediately).

It's almost certainly fine to abort if we panic here but I'd rather we just log if it fails:

```suggestion
        if let Err(err) = fs::remove_file(plugin_path) {
            log::error!("Could not remove plugin: {err}");
        }
```

rbx_reflector/src/cli/generate.rs
```diff
@@ -0,0 +1,267 @@
...
+        "string" => VariantType::String,
+
+        // ProtectedString is handled as the same as string
+        "ProtectedString" => VariantType::String,
+
+        // TweenInfo is not supported by rbx_types yet
+        "TweenInfo" => return Ok(None),
+
+        // Font is not supported by rbx_types yet
+        "Font" => return Ok(None),
```
Dekkonot, line 255: This has changed! Font is supported now. :-)

rbx_reflector/src/defaults.rs
```diff
@@ -0,0 +1,72 @@
...
+
+            return;
+        }
+    };
+
+    for (property_name, property_value) in &instance.properties {
+        let property_name = Cow::Owned(property_name.clone());
+
+        match property_value.ty() {
+            // We don't support usefully emitting these types yet.
```
Dekkonot, line 64: This feels like a misleading comment; we don't emit them because they're not useful to have defaults for. The default of `Ref` properties is "nothing" and no property is `SharedString` by default.

rbx_reflector/src/patches.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "PascalCase", deny_unknown_fields)]
+struct Patch {
+    #[serde(default)]
```
Dekkonot, line 97: I'm curious about the decision to move to only supporting `Change` patches and not `Add`. I'm guessing that it's because the full API dump renders it mostly moot?

nezuo: Yes, I did since all properties exist in the new full dump.

rbx_reflector/src/defaults.rs
nezuo: I'll update the comment, I had copied it from the old `generate_reflection`.

Dekkonot:
That CI failure isn't your fault, don't worry. Since this replaces `generate_reflection`, we should probably remove it once this gets merged which would fix this issue. That said, I'm just going to remove the pinned build CI (#270) because it doesn't make sense for us to do that beyond clippy maybe failing. It doesn't fix it right this minute but it will eventually.

Because this fundamentally changes how we write patch files, would it be possible for you to update [`docs/patching-database.md`](https://github.com/rojo-rbx/rbx-dom/blob/master/docs/patching-database.md)? It's a bit involved because of the nature of this change but maintaining this guide is important.

rbx_reflector/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "rbx_reflector"
+version = "0.1.0"
+edition = "2021"
+
```
Dekkonot, line 5: We should probably add the remaining useful fields. I'm not sure if we plan to publish rbx_reflector but it's good practice.

```suggestion
description = "A program for generating a reflection database for rbx-dom's projects"
license = "MIT"
homepage = "https://github.com/rojo-rbx/rbx-dom"
repository = "https://github.com/rojo-rbx/rbx-dom.git"
authors = ["Nezuo"]
```

You may put whatever name you want in the authors field, this is just a suggestion. The format is `Name <email>` but the email is optional.

Dekkonot (changes requested):
I cannot believe I almost forgot this but we need a `README` for this tool. Just a simple description of what it does and how to use it will do since it's internal. If you need an idea, you can look at the README for `generate_reflection`.

Dekkonot (approved):

PR #252, opened 2023-03-08
Patch ScaleFactor to Scale in reflection database

Dekkonot:
At the moment, we do not serialize `ScaleFactor`. This is mostly fine, except Roblox has begun emitting models with `ScaleFactor` so Rojo occasionally merges them when building models. This has the result of us producing files with `ScaleFactor` set to `0`, which is not good.

This is not critical right now because model scaling is in beta and most people don't have it enabled. It might become critical in the future though. So, this PR patches it so we won't have that problem.

I also added `Scale` as an alias for it because that's what it's exposed to users as in Studio and it's more intuitive than `ScaleFactor`.

---

This is my first time adding a patch to the database and I probably made a mistake, but I did my best. Please let me know if anything needs to change.


nezuo:
I think you need to set `Scale`'s `Scriptability` to `Custom`

rbx_dom_lua/src/customProperties.lua
```diff
@@ -61,4 +61,15 @@ return {
 			end,
 		},
 	},
+	Model = {
+		Scale = {
+			read = function()
+				-- There is no way to return the scale at the moment
```
stravant, line 67: `instance:GetScale()` ?

LPGhatguy (approved):
Cool, let's get this in before it becomes a problem.

PR #254, opened 2023-03-13
Make SSTR chunk deterministic

JohnnyMorganz:
Currently, the SSTR chunk is non-deterministic, where the ordering of the SharedStrings inside the array can change on each run.

We make this deterministic by sorting the shared strings array (based on its hash), and then assigning IDs afterwards.

Dependent on the test files introduced in https://github.com/rojo-rbx/rbx-test-files/pull/16

### DIFF
--- a/rbx_binary/CHANGELOG.md
+++ b/rbx_binary/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 ## Unreleased
 * Added support for `Font` values. ([#248])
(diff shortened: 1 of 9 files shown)


JohnnyMorganz:
Whoops, I didn't see https://github.com/rojo-rbx/rbx-dom/pull/244 which also fixes this. Feel free to close this one out if you want to merge the other one instead

LPGhatguy (approved):
Neat solution to the problem, thanks for the very easy to review PR!

PR #258, opened 2023-03-30
Document ZSTD compression and Bytecode data type in binary spec file

Dekkonot:
Roblox added support for ZSTD compression alongside existing LZ4 compression. It's not very intuitive, but I've documented it here. I can provide a link to what I mean when I say `ZSTD frame` if it's deemed necessary but it should be obvious to anybody who's implementing ZSTD compression since it's normal vocabulary for that algorithm.

This also corrects some laziness and documents the `Bytecode` data type! We haven't documented it because it's not very useful to 3rd party users but I figure we should get to it eventually. I'm optimistically assuming we won't ever support it in `rbx_dom_lua` because why would we do that, but I guess I'm open to changing it.

The documentation comes with a warning against running unsigned bytecode. In a perfect world, this would be unnecessary. Unfortunately, [Roblox literally did this](https://github.com/latte-soft/0x1D) so I figure we have an obligation to at least mention that it's dangerous.


LPGhatguy (approved):

LPGhatguy:
Nice!

PR #259, opened 2023-04-14
Patch `MaterialService.Use2022Materials`

### LINKED ISSUE #243: `MaterialService.Use2022Materials` is serialized internally as `Use2022MaterialsXml`
Currently, as of whatever version of rbx-dom rojo `7.2.1` uses, attempting to write to `MaterialService.Use2022Materials` results in the property not being saved properly. It is successfully written to the place file, but Roblox overrides it because it serializes it differently internally.

How rbx-dom writes the property:
image

How Roblox serializes the property:
image

Workaround in a rojo `*.project.json` file:
image

wackbyte:
Closes #243.

(Do I need to run `generate_reflection` myself?)

### DIFF
--- a/patches/material-service.yml
+++ b/patches/material-service.yml
@@ -1,0 +1,6 @@
+Change:
+  MaterialService:
+    Use2022Materials:
+      Serialization:
+        Type: SerializesAs
+        As: Use2022MaterialsXml


Dekkonot:
This looks good, though we'll have to regenerate the database. The instructions on how to do so aren't amazing, but you'll need to run `generate_reflection` yourself.

If you have bash available, there's a script located [here](https://github.com/rojo-rbx/rbx-dom/blob/master/gen-reflection) but if not, just run the command that's in that script and commit the changed database.

wackbyte:
As I'm on Linux, I'm not sure if I'll be able to run `generate_reflection` very easily.

Dekkonot:
Provided you have the Rust toolchain installed, it should be straightforward because like I said, there's a script available to build and run it like we need. If that's still not possible, I can just generate it myself, though as a general rule I'd prefer to not push to your branch.

wackbyte:
The script tries to locate and run Roblox Studio, which won't work

Dekkonot:
Ah yes that would be a problem! Sorry, I don't usually run Linux so I didn't think about that. 😓 

For the sake of simplicity, I've verified that this affects the database in the intended way and I'll go ahead and merge this without updating it in this PR.

PR #260, opened 2023-04-15
Fix some clippy lints, formatting in documentation, and more

wackbyte:
I've tried to break up my changes into lots of separate commits to make the review process slightly easier.


Dekkonot:
While we're definitely open to clippy and documentation fixes, I'm not sold on the changes to line endings. Is there a reason why normalizing it was necessary?

It's not important to us either way is if there's a reason I'm open to it, but it makes this PR very noisy so if there wasn't a motivation, I'd prefer we don't change them.

PR #261, opened 2023-04-24
Fix support for empty Font tags in rbx_xml

Dekkonot:
In the past, Roblox serialized `Font` properties as empty tags. Although they've stopped doing that, there are still models in the wild with that format for tags. This adds support for empty Font tags by just returning the default value for `Font` when the tag is empty.

Relies on https://github.com/rojo-rbx/rbx-test-files/pull/19.

### DIFF
--- a/rbx_xml/src/types/font.rs
+++ b/rbx_xml/src/types/font.rs
@@ -106,7 +106,7 @@ impl XmlType for Font {
     fn read_xml<R: Read>(reader: &mut XmlEventReader<R>) -> Result<Self, DecodeError> {
         // Patchwork fix for older Roblox files that were written with invalid
         // `Font` tags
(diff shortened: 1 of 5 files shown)


rbx_xml/src/types/font.rs
```diff
@@ -104,6 +104,12 @@ impl XmlType for Font {
     }
 
     fn read_xml<R: Read>(reader: &mut XmlEventReader<R>) -> Result<Self, DecodeError> {
+        // Patchwork fix for older Roblox files that were written with invalid
+        // `Font` tags
+        if let XmlReadEvent::EndElement { name: _ } = reader.expect_peek()? {
```
Kampfkarren, line 109:
```suggestion
        if let XmlReadEvent::EndElement { .. } = reader.expect_peek()? {
```

Kampfkarren (approved):

PR #269, opened 2023-05-13
Fix SignalBehavior serialize patch

Barocena:
Fixes the Serialization of SignalBehavior Property of workspace (SignalBehavior2 -> SignalBehavior)


Dekkonot:
There's a large PR pending right now that changes how we write patch files and I'm not sure if this will be compatible with it. Since we're unlikely to cut a new release until then anyway, I'm going to wait to merge this until after that one. Thank you though!

Dekkonot:
As predicted, we've ended up with an incompatible patch here. We want this to change `SignalBehavior` to be an alias to `SignalBehavior2`. You can see an example in the patch file for smoke and fire [here](https://github.com/rojo-rbx/rbx-dom/blob/master/patches/fire-and-smoke.yml) but I believe it should become literally:

```yml
Change:
  Workspace:
    SignalBehavior:
      Serialization:
        Type: SerializesAs
        As: SignalBehavior2
    SignalBehavior2:
      AliasFor: SignalBehavior
```

Dekkonot:
We're still being left with a merge conflict here so I can't merge this. While I can't actually see what the conflict is, I believe it's just that the file was deleted in master so it should be easy to fix.

PR #270, opened 2023-05-15
Remove pinned version from workflows

Dekkonot:
We don't actually get any benefit from pinning our CI version beyond knowing our MSRV. This is because we aren't keeping lock files in the repository and we've historically just bumped the pinned version until it works.

To prevent this from ever becoming a problem again, we should simply not pin our versions. This has a consequences for Clippy because it means that our CI may suddenly fail in the future, but honestly Clippy lints should be fixed anyway so I don't think that's a bad thing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: [stable, "1.63.0"]
+        rust_version: [stable]
 
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        rust_version: ["1.64.0"]
+        rust_version: [stable]
 
     steps:
     - uses: actions/checkout@v1
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ rbx-dom is a collection of cross-platform libraries that enables any software to
 
 Documentation about the project is hosted at [dom.rojo.space](https://dom.rojo.space).
 
+At this moment, we do not specify a MSRV for any project in rbx-dom. If you need to build one of these libraries with an outdated version of Rust, please open an issue explaining what blockers there are to you updating Cargo, what version you would need us to support, and why.
+
 ## [rbx_dom_weak](rbx_dom_weak)
 [![rbx_dom_weak on crates.io](https://img.shields.io/crates/v/rbx_dom_weak.svg)](https://crates.io/crates/rbx_dom_weak)
 [![rbx_dom_weak docs](https://img.shields.io/badge/docs-docs.rs-orange.svg)](https://docs.rs/rbx_dom_weak)
--- a/generate_reflection/src/main.rs
+++ b/generate_reflection/src/main.rs
@@ -52,11 +52,11 @@ fn run(options: Options) -> anyhow::Result<()> {
 
     if let Some(path) = &options.json_path {
         let encoded = serde_json::to_string_pretty(&database)?;
-        fs::write(&path, encoded)?;
+        fs::write(path, encoded)?;
     }
 
     if let Some(path) = &options.values_path {
-        fs::write(&path, values::encode()?)?;
+        fs::write(path, values::encode()?)?;
     }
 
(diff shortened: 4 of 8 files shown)


Dekkonot:
This is very straightforward but I'm wondering if we should deny clippy warnings in the actual projects as well as in CI like this. Long-term, Rust's [RFC 3389](https://rust-lang.github.io/rfcs/3389-manifest-lint.html) will let us configure it for the entire workspace but right now that'd involve putting `!#[deny(clippy::warnings)]` at the top of every crate's main library. I'm inclined to think it's worth it but I also recognize that it's a bit annoying.

It also does nothing to address new clippy lints from ruining things, so we may end up wanting to add a non-required nightly CI for clippy to warn us of any new lints. I've literally never interacted with unstable toolchains though, so I'd appreciate guidance on that if we want to do it.

Corecii:
One advantage of setting MSRV is that dependant tools know what Rust version to use. You know, like `time` does, and like how they raise it at random with no discussion. We don't actually care about the MSRV and we're practically just raising it when our dependencies raise it. Still, if we start using new Rust features without noticing because we have no MSRV, then our dependants will be forced to raise their Rust version — just like `time` did to us — right?

We should make sure we don't care about that. It's potentially fine to not care about that because Rust is supposed to always be backwards-compatible, so raising the rust version isn't that big of deal. We should listen to the rbx-dom community if they protest, though. Have we considered asking in the OSS discord if anyone cares about removing MSRV for rbx-dom and potentially requiring dependants to update to latest Rust at random? I think standardizing on "always use latest" is acceptable if the community agrees to it.

---

I am not opposed to making clippy lints optional. We can still show them without requiring them for merge.

There is one idea I have: what if we get a list of files changed in the PR diff and only show clippy results for those files? This has some disadvantages, primarily that changing an existing file would ask you to fix all clippy issues in that file even if you only changed a small part of it.

When it comes to a nightly workflow to check clippy lints, we can consider creating an issue whenever new clippy failures appear. The workflow can reuse the existing issue if it's not closed yet, too.

Dekkonot:
> Have we considered asking in the OSS discord if anyone cares about removing MSRV for rbx-dom and potentially requiring dependants to update to latest Rust at random?

Rojo doesn't afaik have a MSRV policy because it's an executable and after asking in the OSS server, it appears that nobody has any objects. Specifically, HaxHelper and Kampfkarren both expressed a desire to not have a MSRV and Lune's maintainer indicated that it explicitly doesn't have an MSRV. A check of the `crates.io` dependents indicates that we've covered our bases so I think we're fine.

> There is one idea I have: what if we get a list of files changed in the PR diff and only show clippy results for those files?

I don't even know where to begin doing that because workflows aren't intuitive but if you want to give it a shot I'm open to it.

With regards to not requiring clippy, I don't think it's a good idea to let warnings past PRs without at least looking at them and anybody with merging permissions can just bypass the CI anyway. It's not an amazing ting to do but as long as this crate is being actively maintained it'll probably be fine since any new clippy failures will just be caught by maintainers so it should be relatively rare to bypass it.

Corecii (approved):

kennethloeffler (approved):
> There is one idea I have: what if we get a list of files changed in the PR diff and only show clippy results for those files? This has some disadvantages, primarily that changing an existing file would ask you to fix all clippy issues in that file even if you only changed a small part of it.

I did a small amount of research and found a tool that is aimed at this problem: https://github.com/o0Ignition0o/cargo-scout, but it looks like it hasn't been touched in a couple years. The author also calls it "experimental and in a very rough draft" and asks to "please wait for 1.0 release" before integrating it into a CI pipline, but it could be worth a shot. If that doesn't work out I think we should still keep clippy lints regardless of how annoying they'll be (at least before we get them all cleaned up), otherwise we'll be throwing code quality to the wind!

Agreed that it's not important for us to have an MSRV - Rust doesn't have breaking changes, and in my mind it's not unreasonable to ask people to keep their Rust up-to-date. We don't care about it, we don't of know anyone does care, and if someone out there suddenly starts caring, well... they should have spoken up here. 😁 If we go forward with this, my only suggestion is to explicitly state that there is no MSRV in the README.

Dekkonot:
I think what's more important right now is getting CI working so I'm going to just merge this. I've mentioned that we don't have an MSRV in the read me and any changes to the clippy workflow can be made later

PR #271, opened 2023-05-18
Implement UniqueId

Dekkonot:
Repeat of #256 because I managed to mess that one up really badly. See that PR for any history.

~~This may still fail in CI but I'll fix it when it happens.~~ Nevermind I guess. Sometimes all you need to do is be careful, I guess.

### DIFF
--- a/rbx_types/src/unique_id.rs
+++ b/rbx_types/src/unique_id.rs
@@ -107,7 +107,7 @@ impl FromStr for UniqueId {
 mod serde_impl {
     use super::UniqueId;
     use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
-    use std::{convert::TryInto, fmt, str::FromStr};
+    use std::{convert::TryInto, fmt};
 
     impl Serialize for UniqueId {
         fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
@@ -155,7 +155,7 @@ mod serde_impl {
         where
             E: de::Error,
         {
-            UniqueId::from_str(v).map_err(E::custom)
+            v.parse().map_err(E::custom)
         }
     }
 
--- a/rbx_types/src/unique_id.rs
+++ b/rbx_types/src/unique_id.rs
@@ -29,7 +29,7 @@ pub enum UniqueIdError {
 }
 
 /// Represents a UUID with a custom epoch of midnight January 1st 2021.
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct UniqueId {
     index: u32,
--- a/rbx_types/src/unique_id.rs
+++ b/rbx_types/src/unique_id.rs
@@ -92,10 +92,11 @@ mod serde_impl {
             if serializer.is_human_readable() {
                 serializer.serialize_str(&self.to_string())
             } else {
-                let mut bytes = Vec::with_capacity(16);
-                bytes.extend_from_slice(&self.random.to_be_bytes());
-                bytes.extend_from_slice(&self.time.to_be_bytes());
-                bytes.extend_from_slice(&self.index.to_be_bytes());
+                let mut bytes = [0; 16];
+
+                bytes[0..8].copy_from_slice(&self.random.to_be_bytes());
+                bytes[8..12].copy_from_slice(&self.time.to_be_bytes());
+                bytes[12..16].copy_from_slice(&self.index.to_be_bytes());
 
                 serializer.serialize_bytes(&bytes)
             }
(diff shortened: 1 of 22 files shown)


rbx_xml/src/types/unique_id.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        let random = i64::from_str_radix(&content[0..16], 16).map_err(|e| reader.error(e))?;
+        let time = u32::from_str_radix(&content[16..24], 16).map_err(|e| reader.error(e))?;
+        let index = u32::from_str_radix(&content[24..32], 16).map_err(|e| reader.error(e))?;
+
+        Ok(UniqueId::new(index, time, random))
+    }
+
+    fn write_xml<W: Write>(&self, writer: &mut XmlEventWriter<W>) -> Result<(), EncodeError> {
+        writer.write_value(&format!(
```
kennethloeffler, line 26: I think we can use `UniqueId::to_string` here now

rbx_types/src/unique_id.rs
```diff
@@ -0,0 +1,211 @@
...
+        let time = SystemTime::now()
+            .duration_since(*EPOCH)
+            .map_err(|_| UniqueIdError::SystemPastTime)?;
+
+        Ok(Self {
+            index: INDEX.fetch_add(1, Ordering::AcqRel),
+            time: u32::try_from(time.as_secs()).map_err(|_| UniqueIdError::Overflow)?,
+            // This matches Roblox's behavior, where the value is both an i64
+            // but is also always positive.
+            random: thread_rng().gen_range(0..i64::MAX),
```
kennethloeffler, line 58: Huh, weird. Do we know why they didn't use all the bits?

rbx_types/src/unique_id.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    impl Serialize for UniqueId {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+        {
+            if serializer.is_human_readable() {
+                serializer.serialize_str(&self.to_string())
+            } else {
+                let mut bytes = Vec::with_capacity(16);
```
kennethloeffler, line 95: Can we use an array and `copy_from_slice` like we do in the binary serializer?

rbx_types/src/unique_id.rs
```diff
@@ -0,0 +1,211 @@
...
+            write!(formatter, "a sequence of 32 hexadecimal characters")
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            if v.len() == 32 {
+                Ok(UniqueId {
+                    random: i64::from_str_radix(&v[0..16], 16).map_err(E::custom)?,
```
kennethloeffler, line 134: There's some code duplication between rbx_xml and rbx_types now... maybe we should `impl FromStr for UniqueId`?

kennethloeffler (changes requested):
More comments:

kennethloeffler:
Oh, we should add UniqueId to the README's coverage table too (and font! but not here, I'll open a PR).

rbx_types/src/unique_id.rs
Dekkonot: Not a clue! My guess is just code reuse since they already have support for `i64`s internally, but that's just a guess.

rbx_types/src/unique_id.rs
Dekkonot: I'm not thrilled with how this ends up looking but I've done it. The error handling in rbx_xml is bad.

rbx_types/src/unique_id.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            write!(formatter, "a sequence of 32 hexadecimal characters")
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            UniqueId::from_str(v).map_err(E::custom)
```
kennethloeffler, line 158:
```suggestion
            v.parse().map_err(E::custom)
```

rbx_xml/src/types/unique_id.rs
```diff
@@ -0,0 +1,36 @@
...
+    serializer_core::XmlEventWriter,
+};
+
+impl XmlType for UniqueId {
+    const XML_TAG_NAME: &'static str = "UniqueId";
+
+    fn read_xml<R: Read>(reader: &mut XmlEventReader<R>) -> Result<Self, DecodeError> {
+        let content = reader.read_characters()?;
+
+        UniqueId::from_str(&content).map_err(|e| reader.error(e))
```
kennethloeffler, line 21:
```suggestion
       content.parse().map_err(|e| reader.error(e))
```

kennethloeffler (changes requested):
I suggest we use `parse` (this way we don't have to import `FromStr`) but otherwise LGTM

kennethloeffler (approved):

PR #275, opened 2023-05-25
Resolve problem with updating database using rbx_reflector 

Dekkonot:
Through a fun and exciting bug I cannot identify, using `rmp-serde` at version `1.1.1` causes all sorts of problems when generating a reflection database. It does however work just fine at version `0.14.4`, which is what `generate_reflection` was on. Debugging the issue turned out to be cumbersome, so for now I'm suggesting we downgrade `rbx_reflector` to that version. This PR also updates the database (to confirm it works) and updates snapshots.

I've also added a more useful panic to `rbx_reflection_database` that actually displays the error's message when we attempt to deserialize the database. We can't recover from the error but we shouldn't just unwrap and call it good.


kennethloeffler (commented):
The reason this didn't work is because of a rmp-serde version mistmatch between rbx_reflector and rbx_reflection_database. 
  
If you simply bump rbx_reflection_databases rmp-serde version, then you'll still get an error, because the database was serialized using the old version. For this to work, you must comment out the `apply_defaults` call in the generate command (this way we never try to load the old database), generate the database, uncomment the call, then generate the database again. If we're going to upgrade this dependency, we should probably bump the rbx_reflection_database's version.
  
The database looks fine after a quick (but comprehensive) review,  but I would have appreciated a more explicit acknowledgment that this is the first time we've run rbx_reflector in the repo. Are we sure we're comfortable shipping these databases yet?

Dekkonot:
> The database looks fine after a quick (but comprehensive) review, but I would have appreciated a more explicit acknowledgment that this is the first time we've run rbx_reflector in the repo. Are we sure we're comfortable shipping these databases yet?

I'd like to clarify this isn't the first time I'm running `rbx_reflector` in the repo. I ran it during my review to ensure the database it emitted was coherent, which it is. I didn't check the messagepack database but I assumed that because we're using serde it'd be the same across both json and messagepack.

What I failed to do, which is what lead us here, was run it twice in a row or actually try to run tests with a database it generated. I don't have an excuse for that beyond my own complacency because I should have ran `cargo test` before merging. The only explanation I can give is that I assumed that it would be fine because I checked the database JSON and trust Serde to either fail or serialize correctly. In hindsight I should have made sure a database was generated and included with the PR but that's all I can say.

I *do* trust this database because I trust that our tests would catch something being wrong with it. I've also verified it manually to make sure we're fine. If you aren't comfortable deploying this, I'd appreciate an explicit indicator of what you think we need to account for. Clearly I make mistakes and many eyes are better than just mine.

---

With regards to version migration, I could have sworn I'd literally done that. If I didn't, I'm truly sorry because that's basic debugging. I will give that a go in the morning. 🤦🏼

kennethloeffler:
> I'd like to clarify this isn't the first time I'm running rbx_reflector in the repo. I ran it during my review to ensure the database it emitted was coherent, which it is.

Oh, my bad. It did read that way at first! 😂

Dekkonot:
I've done some deeper investigating and found a better solution that doesn't involve downgrading and fixes the underlying bug. See the linked PR for details.

PR #280, opened 2023-06-02
Rewrite rbx_xml to use quick_xml instead of xml-rs

Dekkonot:
`rbx-xml` is slow and has trouble processing unicode, among other issues. This can and is solved by simply swapping to a new XML parser. Given the speed, I've decided to go with `quick-xml`.

Unfortunately, the current parser (`xml-rs`) is intertwined rather thoroughly with `rbx-xml`'s implementation and swapping it out for `quick-xml` would be impossible without refactoring large segments of it. Given there's a few other design choices that I wanted to change, I decided to bite the bullet and go for a full rewrite. Initial testing is good and suggests we're looking at 15x the parsing speed in some cases. I have tried to prioritize not copying data when possible, so overall memory should also be low.

**Update:** This is mostly done now, I believe. Many many tests need to be written but the bulk of the code will not change. With that in mind, here are my thoughts beyond "reviewing this is a nightmare":

1. This rewrite fundamentally changes how the crate is laid out. Specifically, it's split `serializer` and `deserializer` into dedicated modules and the code for (de)serializing data types now lives under those modules. This differs dramatically from the current system we use that implements `XmlType` for each type and both deserialization and serialization in one place. In my opinion, this is easier to reason with but it is significantly different and may require us to point contributors in the right direction if they're used to the old system. Also, it means we have the potential for element name mismatches between serializing and deserializing. Testing will absolutely catch that, but it's something we didn't have to deal with before.

2. The deserializer now requires the `referent` attribute to be set on `Item` elements. I was very surprised to learn it was not before, but as a result this is also a breaking change. I don't think we should care though.

---
For posterity, here's a checklist of what needs to be done:
- [x] Visit properties using the `ReflectionDatabase`
- [x] Port data types
- [x] Redo the serializer
- [x] Run every test like a million times to ensure I didn't break anything


rbx_xml/src/deserializer/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+mod reader;
+
+use std::io::Read;
+
+pub use self::core::DecodeConfig;
+pub use error::DecodeError;
+use rbx_dom_weak::WeakDom;
+pub use reader::XmlReader;
+
+pub fn decode_internal<R: Read>(reader: R, option: DecodeConfig) -> Result<WeakDom, DecodeError> {
```
nezuo, line 14:
```suggestion
pub fn decode_internal<R: Read>(reader: R, options: DecodeConfig) -> Result<WeakDom, DecodeError> {
```

nezuo: Could alternatively rename these parameters to `config`.

rbx_xml/src/deserializer/error.rs
```diff
@@ -0,0 +1,77 @@
+use rbx_dom_weak::types::VariantType;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+#[error(transparent)]
+pub struct DecodeError(Box<ErrorKind>);
```
nezuo, line 6: Does this need to be boxed?

Dekkonot: Technically no but `ErrorKind` is a very big struct and errors are being passed around by every function. It saves us a bunch of space on the stack to have it boxed.

rbx_xml/src/property_descriptor.rs
```diff
@@ -0,0 +1,113 @@
...
+                let serialized_descriptor = class_descriptor
+                    .properties
+                    .get(serialized_name.as_ref())
+                    .expect("database SerializesAs entries should point to a real property");
+
+                Some((property_descriptor, serialized_descriptor))
+            }
+            // The lifetime on 'PropertyKind' makes us do this.
+            // I don't know if we can fix it.
+            _ => unimplemented!(),
```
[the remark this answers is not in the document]
Dekkonot: I assumed it was the lifetime but that makes a lot more sense, ha ha. I suppose it's probably marked as `non_exhaustive` so that future implementations (like this one) don't break if we add new options for the database.

Dekkonot:
Rather than work out what work I'm missing since I started this rewrite, I'm going to close this and instead focus on a less invasive rewrite. I think it will result in a much faster turn around for the port, and in turn it should end up being easier to review. Sad to see my work go, but it will not die here.

PR #283, opened 2023-06-13
Implement property migration

kennethloeffler:
This PR is based off of #268  - all of the description there also applies here. The only thing that really needed to be done was supporting Migrate patches in rbx_reflector.

I've also slightly refactored the whole canonical property dance in the rbx_binary deserializer state, hopefully it's a bit easier to follow. 

TODO:
- [ ] Fix up changelogs
- [ ] Address review feedback from #268


rbx_dom_lua/src/database.json
```diff
@@ -24635,7 +24759,7 @@
           "Enum": 0
         },
         "ScriptGuid": {
-          "String": "{7EFC8201-25F5-42C7-8DDB-DD7BD2D32FE7}"
+          "String": "{7604441B-B379-4EE1-984B-AE157BC63C77}"
```
Dekkonot, line 24762: We should probably do something about ScriptGuid going forward because it seems like it isn't stable between versions... Which is probably expected behavior given it's a GUID.

It doesn't need to and shouldn't happen in this PR but it's something to think about because I'm not sure how to actually solve a property that we shouldn't store a default for like this.

rbx_dom_weak/src/instance.rs
```diff
@@ -109,6 +109,11 @@ impl InstanceBuilder {
         self.properties.insert(key.into(), value.into());
     }
 
+    /// Check if the `InstanceBuilder` already has a property with the given key.
+    pub fn has_property<K: Into<String>>(&self, key: K) -> bool {
+        self.properties.contains_key(&key.into())
+    }
+
```
Dekkonot, line 116: This needs to be added to the changelog for rbx-dom-weak

rbx_types/src/font.rs
```diff
@@ -92,3 +92,20 @@ impl Default for Font {
...
+            cached_face_id: None,
+        }
+    }
+    pub fn regular(family: &str) -> Self {
+        Self {
+            family: family.to_owned(),
+            ..Default::default()
+        }
+    }
+}
```
Dekkonot, line 111: Add these to the changelog

Dekkonot (changes requested):

rbx_binary/src/deserializer/state.rs
```diff
@@ -85,6 +85,132 @@ struct Instance {
...
+            // TODO: Do we need an additional fix here?
+            let canonical_name = &descriptors.canonical.name;
+            let canonical_type = match &descriptors.canonical.data_type {
+                DataType::Value(ty) => *ty,
+                DataType::Enum(_) => VariantType::Enum,
+                _ => {
+                    // TODO: Configurable handling of unknown types?
+                    return None;
+                }
+            };
```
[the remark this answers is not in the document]
kennethloeffler: > I realize these are copied verbatim from the old source but it does make me wonder what was actually meant by the first one.
> 
> With regards to the second one, we don't really need to worry about it because I don't foresee a world where Roblox adds something that's not a data type or an enum that also serializes.

These comments were added by https://github.com/rojo-rbx/rbx-dom/pull/230. 

The first comment might be talking about other possible ways rbx_binary can serialize properties that aren't supposed to serialize? But I'm not sure what other cases could lead to this kind of problem :shrug:

The fall through is hit on an unknown property type, and the second comment is referring to doing something smarter in that case. I think we should keep this comment.

rbx_dom_lua/src/database.json
kennethloeffler: Agreed that it doesn't make sense to have a default ScriptGuid - we ought to catalogue any properties for which defaults aren't sensible and figure out a good strategy to deal with them

Dekkonot (approved):

PR #288, opened 2023-06-18
Prevent UniqueId collisions within WeakDom's API

### LINKED ISSUE #285: rbx_dom_weak should prevent `UniqueId` collisions
It's possible to use a tool like [Lune](https://github.com/filiptibell/lune) to clone instances and place them into the same DOM. This is currently problematic because instances may contain `UniqueId` properties. Such an operation on an instance with `UniqueId` properties will cause `UniqueId` collisions, which cause Roblox Studio to crash when they're opened. For example, I've attached a base rbxl, an rbxl containing `UniqueId` collisions, and the Lune script that generated the rbxl with collisions: [uniqueid-collisions.zip](https://github.com/rojo-rbx/rbx-dom/files/11777134/uniqueid-collisions.zip).

rbx_dom_weak should probably check `UniqueId` properties on insertion, regenerating them if they collide with another instance in the DOM.

kennethloeffler:
Closes #285, using Approach 1 as described in the issue. I ended up having to touch `WeakDom::transfer` as well, because that method could also create collisions.

It feels a bit icky to hard code the property name in these checks, but it's *very* unlikely that we or Roblox will ever want to give it a different name.

This PR additionally changes `WeakDom::insert` to accept a none `Ref` for the parent (previously, it'd just fall over and panic). The reason why is that I wanted to reuse the `UniqueId` checking code in `WeakDom::new` (to uphold the invariant even when the prop is in the root builder), but `WeakDom::insert` did not allow a none parent. The only difference for insertions with none parent vs some parent is whether the parent's list of children is extended, so it's pretty minimal & I think it makes sense.

### DIFF
--- a/rbx_dom_weak/src/dom.rs
+++ b/rbx_dom_weak/src/dom.rs
@@ -248,6 +248,7 @@ mod test {
     use super::*;
 
     use crate::DomViewer;
+    use rbx_types::{UniqueId, Variant};
 
     #[test]
     fn transfer() {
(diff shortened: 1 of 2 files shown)


kennethloeffler (commented):
I think I'm fine with where it stands. I only think that the dom tests are now pretty big, and maybe they should go in a different file. or be done in a different way

Dekkonot (approved):

Dekkonot:
I'm inclined to agree that hardcoding it is icky, but this is a good solution for now. If we feel strongly later, we can add it as a new field for `Instance` in the same way `Name` is, but I don't think it's necessary right now.

PR #293, opened 2023-06-21
Fail on invalid attributes in rbx_binary deserializer

### LINKED ISSUE #292: rbx_binary should not continue parsing if it runs into unreadable attributes
The rbx_binary deserializer currently reads `Attributes` as a `BinaryString` in the case of an attribute read failure: https://github.com/rojo-rbx/rbx-dom/blob/2e78feb05e033cbca8db1d9e490f8334c096d13e/rbx_binary/src/deserializer/state.rs#L438-L462

This is surprising for consumers who expect this property to be of type `Attributes`. rbx_binary should probably match rbx_xml's behavior in this case, and fail the whole deserialization operation if it fails to read attributes. 

This means that rbx_binary will become unusable if any unknown types are present (or if the format ever changes). The attributes format doesn't give us a way to recover gracefully here, but it's probably better to blow up immediately instead of creating a `BinaryString` value.

kennethloeffler:
Closes #292.

Not really sure about `PropertyError` for the error variant name, maybe there is a better way to phrase it?

### DIFF
--- a/rbx_binary/src/deserializer/error.rs
+++ b/rbx_binary/src/deserializer/error.rs
@@ -79,4 +79,11 @@ pub(crate) enum InnerError {
         expected_type_id: u8,
         actual_type_id: u8,
     },
+
+    #[error("Failed to deserialize {class_name}.{prop_name} because {source}")]
+    PropertyError {
+        source: rbx_dom_weak::types::Error,
+        prop_name: String,
+        class_name: String,
+    },
 }
--- a/rbx_binary/src/deserializer/error.rs
+++ b/rbx_binary/src/deserializer/error.rs
@@ -81,7 +81,7 @@ pub(crate) enum InnerError {
     },
 
     #[error("Failed to deserialize {class_name}.{prop_name} because {source}")]
-    PropertyError {
+    BadPropertyValue {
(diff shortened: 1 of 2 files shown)


rbx_binary/src/deserializer/error.rs
```diff
@@ -79,4 +79,11 @@ pub(crate) enum InnerError {
         expected_type_id: u8,
         actual_type_id: u8,
     },
+
+    #[error("Failed to deserialize {class_name}.{prop_name} because {source}")]
+    PropertyError {
```
Dekkonot, line 84: I'm inclined to agree that `PropertyError` is a bad name. Maybe something like `DataTypeError` instead? That would let us perhaps fold stuff like CFrame's `BadRotationId` error variant into it in the future.

kennethloeffler: Renamed to `BadPropertyValue`, and we can replace usage of `BadRotationId` with it  (and possibly `InvalidPropData`)?

Dekkonot:
We can definitely fold the other errors into `BadPropertyValue` after this is merged, but it feels low priority.

PR #296, opened 2023-06-25
Implement CFrame attributes

### LINKED ISSUE #291: Attributes need `CFrame` support
Roblox recently added support for `CFrame` types in attributes. and we need to catch up. This will probably require pulling out some code from rbx_binary, assuming that the same compression scheme is used for axis-aligned rotation matrices

kennethloeffler:
Closes #291.

The serialization of `CFrame` attributes works mostly the same way as in the binary format, so this is fairly straightforward.

I'm not 100% sure about testing this via the rbx_xml and rbx_binary tests, but I like having the tests be backed by actual files instead of static strings, and well... it works well enough I think?


docs/attributes.md
```diff
@@ -147,6 +148,38 @@ The `Vector3` type is a struct composed of three `f32`s:
...
+|:----------------|---------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Position        | `Vector3`           | The positional component of the `CFrame`.                                                                                                                                                                                                              |
+| Rotation ID     | `u8`                | An identifier representing axis-alignment of the rotation, determining whether the rotation matrix is present (see `Rotation matrix`). If the rotation *is* axis-aligned, this field is one of the values from the table below; otherwise, it is `00`. |
+| Rotation matrix | `Optional<Matrix3>` | The rotational component of the `CFrame`, stored as a sequence of nine `f32`s. If Rotation ID is `00`, this field is the rotation matrix in **row-major order**, the XVector, the YVector, and ZVector, in that order; otherwise, it is absent.        |
```
Dekkonot, line 160: See my comment on the correction to the documentation for CFrame in the binary spec [here](https://github.com/rojo-rbx/rbx-dom/pull/297#discussion_r1242917009).

Dekkonot (approved):
LGTM

PR #297, opened 2023-06-26
Correct binary CFrame docs to say the matrix is in row-major order

kennethloeffler:
The docs for the binary format's `CFrame` representation are a bit wrong: the rotation matrix is actually stored in row-major order, and not column-major order as the current docs suggest. This can be verified by reviewing the implementation (which reads the rotation matrix in row-major order), changing rotation matrices in Roblox Studio and inspecting the output, and reviewing other community specifications such as the [RobloxFileSpec](https://www.classy-studios.com/Downloads/RobloxFileSpec.pdf).

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -411,7 +411,7 @@ If the byte is `00`, a `CFrame` looks like this:
 | Field Name  | Format                  | Value                                                                                                        |
 |:------------|:------------------------|:-------------------------------------------------------------------------------------------------------------|
 | ID          | `u8`                    | Always `00` in this case.                                                                                    |
-| Orientation | Array of 9 `f32` values | The rotation matrix of the `CFrame`. It represents the RightVector, UpVector, and LookVector, in that order. |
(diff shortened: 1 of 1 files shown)


docs/binary.md
```diff
@@ -411,7 +411,7 @@ If the byte is `00`, a `CFrame` looks like this:
 | Field Name  | Format                  | Value                                                                                                        |
 |:------------|:------------------------|:-------------------------------------------------------------------------------------------------------------|
 | ID          | `u8`                    | Always `00` in this case.                                                                                    |
-| Orientation | Array of 9 `f32` values | The rotation matrix of the `CFrame`. It represents the RightVector, UpVector, and LookVector, in that order. |
+| Orientation | Array of 9 `f32` values | The rotation matrix of the `CFrame` in **row-major order**: the XVector, YVector, and ZVector, in that order.|
```
Dekkonot, line 414: I don't feel that specifying "in row-major order" is actually helpful to would-be implementers. If we want to be specific, we can indicate the order the components are in (using the all too popular `r00` through `r22`) but I think just saying that it's the `XVector`, `YVector`, and `ZVector` should be enough for anyone familiar with Roblox's ecosystem.

Dekkonot (approved):

PR #299, opened 2023-06-27
Implement font attributes

### LINKED ISSUE #290: Attributes need `Font` support
Roblox recently added support for the `Font` type in attributes, and we need to catch up

kennethloeffler:
Closes #290.

Should the folder-with-font-attribute test files have more & varied font attributes added? I sort of want to add more, but I think the spec is correct, and I'm not sure how important it is to test against more values here

### DIFF
--- a/rbx_types/src/attributes/error.rs
+++ b/rbx_types/src/attributes/error.rs
@@ -35,4 +35,12 @@ pub(crate) enum AttributeError {
 
     #[error("couldn't read bytes to deserialize {0}")]
     ReadType(&'static str),
+
+    #[error("font contained invalid UTF-8 in {field}")]
+    FontBadUnicode {
+        #[source]
+        source: FromUtf8Error,
+
+        field: String,
+    },
 }
--- a/rbx_types/src/attributes/error.rs
+++ b/rbx_types/src/attributes/error.rs
@@ -41,6 +41,6 @@ pub(crate) enum AttributeError {
         #[source]
         source: FromUtf8Error,
 
-        field: String,
+        field: &'static str,
     },
 }
--- a/rbx_types/src/attributes/reader.rs
+++ b/rbx_types/src/attributes/reader.rs
@@ -4,9 +4,9 @@ use std::{
 };
 
 use crate::{
-    BinaryString, BrickColor, CFrame, Color3, ColorSequence, ColorSequenceKeypoint, Matrix3,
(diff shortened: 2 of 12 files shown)


Dekkonot:
I don't imagine there's any variation in the way Fonts are serialized. One test is probably enough. I won't stop you from adding more, but I don't think it's necessary.

I'll give the code a once over in the morning but at a glance, please add this to the changelog 😉

rbx_types/src/attributes/error.rs
```diff
@@ -35,4 +35,12 @@ pub(crate) enum AttributeError {
 
     #[error("couldn't read bytes to deserialize {0}")]
     ReadType(&'static str),
+
+    #[error("font contained invalid UTF-8 in {field}")]
+    FontBadUnicode {
+        #[source]
+        source: FromUtf8Error,
+
+        field: String,
```
Dekkonot, line 44:
```suggestion
        field: &'static str,
```

rbx_types/src/attributes/reader.rs
```diff
@@ -161,6 +161,43 @@ pub(crate) fn read_attributes<R: Read>(
...
+                let cached_face_id = {
+                    let buf = read_string(&mut value)?;
+
+                    if buf.is_empty() {
+                        None
+                    } else {
+                        Some(String::from_utf8(buf).map_err(|source| {
+                            AttributeError::FontBadUnicode {
+                                source,
+                                field: "cached_face_id".to_string(),
```
Dekkonot, line 186:
```suggestion
                                field: "cached_face_id",
```

rbx_types/src/attributes/reader.rs
```diff
@@ -161,6 +161,43 @@ pub(crate) fn read_attributes<R: Read>(
...
+            VariantType::Font => {
+                let weight = read_u16(&mut value)?;
+                let style = read_u8(&mut value)?;
+
+                let family = {
+                    let buf = read_string(&mut value)?;
+
+                    String::from_utf8(buf).map_err(|source| AttributeError::FontBadUnicode {
+                        source,
+                        field: "family".to_string(),
```
Dekkonot, line 173:
```suggestion
                        field: "family",
```

Dekkonot (changes requested):
No reason for us to bother with a `String` when we can just use `&'static str` but otherwise looks good to me.

kennethloeffler:
Commited from my iPhone

Dekkonot (approved):

Dekkonot:
Merged from my Android

Issue #301, opened 2023-06-29
Type mismatch: Property IntValue.Value should be Int32, but it was Int64


ThatTimothy:
A bit of background: I've been using [Lune](https://github.com/filiptibell/lune) to convert over some old files into a more usable file system format rather than arbitrary binary files. It seems like some of my older files aren't able to be deserialized.

I can reopen them and save again, and it seems to work fine after that, but I really don't want to do that another 2,000 times. If this is a non-issue, I totally understand lol. It's a very old file from early 2018.

Lune log with `RUST_LOG=trace`:

```lua
local inputFile = fs.readFile("repo.rbxl")
local game = roblox.deserializePlace(inputFile)
print(game.ClassName)
```

```txt
[TRACE rbx_binary::chunk] Chunk "PROP" (compressed: 24, len: 22, reserved: 0)
[TRACE rbx_binary::deserializer::state] PROP chunk (IntValue.Value, instance type 29, prop type 29
[TRACE rbx_binary::deserializer::state] Known prop, canonical name Value and type Int64, with None migration
Failed to read document from buffer - Type mismatch: Property IntValue.Value should be Int32, but it was Int64
```

Remodel log with `RUST_LOG=trace`:
```lua
local game = remodel.readPlaceFile("repo.rbxl")
print(game.ClassName)
```

```txt
[TRACE] Known prop, canonical name Tags and type Tags
[TRACE] Chunk "PROP" (compressed: 24, len: 22, reserved: 0)
[TRACE] PROP chunk (IntValue.Value, instance type 29, prop type 29
[TRACE] Known prop, canonical name Value and type Int64
[ERROR] callback error
        stack traceback:
                [C]: in field 'readPlaceFile'
                [string "remodel.lua"]:1: in main chunk
        caused by: Type mismatch: Property IntValue.Value should be Int32, but it was Int64
```

Repo file: https://github.com/ThatTimothy/rbx-binary-issue-repo/blob/main/repo.rbxl (didn't let me upload directly, sorry)

kennethloeffler:
[...]

PR #303, opened 2023-06-29
Add int/float widening support to rbx_xml

Dekkonot:
With issue #301, it's became clear we need a way to widen integers and floats from 32-bit to 64-bit since older models may have outdated types for number properties. This PR implements widening for `Int32` and `Float32`, giving them a very simple conversion to `Int64` and `Float64` respectively. This should solve the linked issue in the XML format and prevent it from being a problem for floats as well.

Support in `rbx_binary` is straightforward but the author of that issue has a fix ready to merge, so I'll let them do it. :smile:

### DIFF
--- a/rbx_xml/src/conversion.rs
+++ b/rbx_xml/src/conversion.rs
@@ -27,9 +27,11 @@ impl ConvertVariant for Variant {
         target_type: VariantType,
     ) -> Result<Cow<'_, Self>, String> {
         match (value.borrow(), target_type) {
-            (Variant::Int32(value), VariantType::Int64) => Ok(Cow::Owned((*value as i64).into())),
+            (Variant::Int32(value), VariantType::Int64) => {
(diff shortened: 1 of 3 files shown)


rbx_xml/src/conversion.rs
```diff
@@ -27,6 +27,10 @@ impl ConvertVariant for Variant {
         target_type: VariantType,
     ) -> Result<Cow<'_, Self>, String> {
         match (value.borrow(), target_type) {
+            (Variant::Int32(value), VariantType::Int64) => Ok(Cow::Owned((*value as i64).into())),
+            (Variant::Float32(value), VariantType::Float64) => {
+                Ok(Cow::Owned((*value as f64).into()))
+            }
```
kennethloeffler, line 33: Can we use `i64::from` and `f64::from` so it's more clear no loss can occur in these conversions?

kennethloeffler (approved):
lgtm, thanks for adding the comment

PR #307, opened 2023-07-01
Get Studio version from plugin request in rbx_reflector

### LINKED ISSUE #304: rbx_reflector does not always reflect the current Roblox Studio install's version
While I was working on #302, I noticed that the `Version` field in the database stayed at `[0, 578, 0, 5780566]` after I ran rbx_reflector, even though my Roblox Studio was on a later version:
```json
{
  "Version": [
    0,
    578,
    0,
    5780566
  ],
<truncated>
```
image

I discovered that rbx_reflector is grabbing the version from a Roblox endpoint: 
https://github.com/rojo-rbx/rbx-dom/blob/fd147f7d90cd60e81ae72a224e9760799c653ff4/rbx_reflector/src/studio_version.rs#L11
Right now, it sends me back this:
```
{"version":"0.578.0.5780566","clientVersionUpload":"version-1f12a0b5b1bf44ce","bootstrapperVersion":"1, 6, 0, 5780566"}
```

Clearly, newer versions of Roblox Studio can ship before this endpoint bumps versions. In any case, there is no guarantee that the local Roblox Studio install is the same version as the one obtained from this endpoint, so rbx_reflector shouldn't do this. Instead, it should probably obtain the version from the injected plugin, like generate_reflection: 
https://github.com/rojo-rbx/rbx-dom/blob/fd147f7d90cd60e81ae72a224e9760799c653ff4/generate_reflection/plugin/main.lua#L1-L13

This ensures that it *always* gets the version of the Roblox Studio actually installed on the machine, which is exactly what we want here.

cc @nezuo

kennethloeffler:
Closes #304.


rbx_reflector/src/cli/mod.rs
```diff
@@ -29,7 +29,7 @@ impl Args {
     pub fn run(&self) -> anyhow::Result<()> {
         match &self.subcommand {
             Subcommand::Dump(sub) => sub.run(),
-            Subcommand::DefaultsPlace(sub) => sub.run(),
+            Subcommand::DefaultsPlace(sub) => sub.run().map(|_| ()),
```
Dekkonot, line 32: ...What is this?

kennethloeffler: The defaults place subcommand now returns Result<StudioInfo> instead of Result<()>, and this top-level run function still returns a Result<()>. I wanted to continue to propagate any error information, so chose to just map back to a unit

PR #312, opened 2023-07-06
Implement instance cloning methods in WeakDom

### LINKED ISSUE #282: `rbx_dom_weak` should provide a way to clone instances
It's a pain for consumers to implement instance cloning themselves. I think `rbx_dom_weak` should give them a way to do it, and it should have the same semantics as Roblox's `Instance:Clone`:

* All of the instance's descendants are also cloned (i.e. the entire subtree is cloned),
* Any `Ref` properties in the cloned subtree that point to instances also contained in the subtree (i.e. point to instances that were cloned) are rewritten to point to the cloned instances,
* The instance at the root of the sub-tree is returned and its parent is set to the none `Ref`.

I'm a little torn about whether this should be an `impl Clone for Instance` or a method on `WeakDom` - I'd appreciate some input here.

kennethloeffler:
Closes #282 by implementing `WeakDom::clone_within` and `WeakDom::clone_into_external`.

### DIFF
--- a/rbx_dom_weak/src/dom.rs
+++ b/rbx_dom_weak/src/dom.rs
@@ -257,7 +257,7 @@ impl WeakDom {
             self.insert(cloned_parent, builder);
         }
 
-        Self::rewrite_refs(self, &ctx);
+        Self::rewrite_refs(self, ctx);
 
         root_ref
     }
@@ -278,7 +278,7 @@ impl WeakDom {
             dest.insert(cloned_parent, builder);
         }
 
-        Self::rewrite_refs(dest, &ctx);
+        Self::rewrite_refs(dest, ctx);
 
         root_ref
     }
@@ -349,7 +349,7 @@ impl WeakDom {
         builder
     }
 
-    fn rewrite_refs(dom: &mut WeakDom, ctx: &CloneContext) {
+    fn rewrite_refs(dom: &mut WeakDom, ctx: CloneContext) {
         for (_, new_ref) in ctx.ref_rewrites.iter() {
             let instance = dom
                 .get_by_ref_mut(*new_ref)
--- a/rbx_dom_weak/src/dom.rs
+++ b/rbx_dom_weak/src/dom.rs
@@ -18,6 +18,102 @@ pub struct WeakDom {
     unique_ids: HashSet<UniqueId>,
 }
 
+#[derive(Debug)]
+struct CloneContext<'dom> {
+    // TODO: Should we use a specific type for these instead of (Ref, Ref)?
+    queue: VecDeque<(Ref, Ref)>,
+
+    ref_rewrites: HashMap<Ref, Ref>,
+    dom: &'dom mut WeakDom,
+    subtree_root: Ref,
+}
+
+impl<'dom> CloneContext<'dom> {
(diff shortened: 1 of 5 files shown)


rbx_dom_weak/src/dom.rs
```diff
@@ -277,6 +325,47 @@ impl WeakDom {
...
+            .with_properties(instance.properties.clone());
+
+        let new_ref = builder.referent;
+
+        for uncloned_child in instance.children.iter() {
+            ctx.queue.push_back((new_ref, *uncloned_child))
+        }
+
+        ctx.ref_rewrites.insert(original_ref, new_ref);
+        builder
```
kennethloeffler, line 349: This builder *never* has any children and kinda makes me want a fast path in `WeakDom::insert` that doesn't create a `VecDeque` for builders with no children - kind of irrelevant for this PR, but I'll put it on my back burner...

rbx_dom_weak/src/dom.rs
```diff
@@ -235,6 +235,48 @@ impl WeakDom {
...
+        let mut ctx = CloneContext::default();
+        let root_builder = ctx.clone_ref_as_builder(self, referent);
+        let root_ref = dest.insert(Ref::none(), root_builder);
+
+        while let Some((cloned_parent, uncloned_child)) = ctx.queue.pop_front() {
+            let builder = ctx.clone_ref_as_builder(self, uncloned_child);
+            dest.insert(cloned_parent, builder);
+        }
+
+        ctx.rewrite_refs(dest);
```
Dekkonot, line 276: Cloning into an external DOM and leaving a bunch of Ref properties that point to nothing (despite not being `null`) makes me leery, so I went and checked what our serializers do when they run into one.

`rbx_binary` handles orphaned referents invisibly and correctly by just writing them equivalently to `null` but `rbx_xml` doesn't, and will increment the internal referent counter by 1 whenever it runs into a non-null referent. As a result, orphaned Ref properties will cause the rest of the file to be off by one. This isn't a problem since the deserializer skips unknown referents, but it makes this method for cloning not invisible to consumers like Lune.

We can either fix that behavior in `rbx_xml`'s serializer or rewrite orphaned Ref properties to be `None`, or do both. I'm inclined to say both but doing it in `rbx_xml` is outside of scope for this PR.

rbx_dom_weak/src/dom.rs
```diff
@@ -235,6 +235,48 @@ impl WeakDom {
...
+        root_ref
+    }
+
+    /// Clone the instance with the given `referent` and all its descendants (i.e. the
+    /// entire subtree) into the given WeakDom.
+    ///
+    /// After the operation, the root of the cloned subtree has no parent.
+    ///
+    /// Any Ref properties that point to instances contained in the subtree are
+    /// rewritten to point to the cloned instances.
```
Dekkonot, line 265:
```suggestion
    /// Any Ref properties that point to instances contained in the subtree are
    /// rewritten to point to the cloned instances. Any other Ref properties
    /// would be invalid  in `dest` and are thus rewritten to be `Ref::none()`
```

Dekkonot:
Can you add these to the Changelog also?

Dekkonot (approved):
LGTM

Issue #302, opened 2023-06-29
`Workspace.CollisionGroups` doesn't serialize anymore, and other collision group woes 


kennethloeffler:
This is a bit of a multilayered issue, so bear with me....

### Changes to the collision group API
The potential for problems with collision groups seems to begin in September 2022, when [Roblox made changes to the collision groups API](https://devforum.roblox.com/t/updates-to-collision-groups/1990215). As part of this change, it looks like `Workspace.CollisionGroups` was superseded by `Workspace.CollisionGroupData`.

### User reports of problems with collision groups
I can find reports of problems in the Rojo ecosystem involving `Workspace.CollisionGroupData` going back to January 2023. I'll provide an illuminating excerpt from a conversation in the Roblox OSS Discord server:

```
User1 — 01/13/2023 4:32 PM
    Since the new collision group API, is it still possible to set collision groups at build time?
    I'm having trouble writing out the property by hand
    Searching here the format is something like
    [User1 shows the format for `Workspace.CollisionGroups`]
```
```cobol
    "Default^0^-1\\Bullet^1^-1"
```
```
User2 — 01/13/2023 4:39 PM
    you can change the collisions and save it as rbxlx* and check what it gives for collisiongroup
```
```
User1 — 01/13/2023 4:40 PM
    Save the workspace as a model file?

    Wont let me do that.. I did try saving the place as an xml and got something hashed

    [User1 shows a snippet from a .rbxlx file, showing the Base64 form of Workspace.CollisionGroupData]
```
```xml
    <BinaryString name="CollisionGroupData"><![CDATA[AQcABPP///8HRGVmYXVsdAEEi////wZCdWxsZXQCBIT///8HVHJpZ2dlcgMEgv///whXZWFyYWJsZQQEof///wVUZWFtQQUEkf///wVUZWFtQgYEwf///wVXb3JsZA==]]></BinaryString>
```
```
User2 — 01/13/2023 5:16 PM
    ok so
    [User2 again shows the Base64-encoded form of Workspace.CollisionGroupData, this time embedded in a Rojo project]
```
```json
        "Workspace": {
          "$properties": {
            "CollisionGroupData": {
              "BinaryString": "AQcABPP///8HRGVmYXVsdAEEi////wZCdWxsZXQCBIT///8HVHJpZ2dlcgMEgv///whXZWFyYWJsZQQEof///wVUZWFtQQUEkf///wVUZWFtQgYEwf///wVXb3JsZA=="
            }
[...]

Issue #277, opened 2023-05-27
rbx_dom_weak should have a concept of "proxy properties"


kennethloeffler:
Proxy properties are those properties that are mutually dependent, for example:

* `BasePart.Position`, `BasePart.Orientation`, `BasePart.Rotation`, and `BasePart.CFrame`
* `Lighting.TimeOfDay` and `Lighting.ClockTime`
* Others?

Right now, consumers are totally responsible for handling proxy properties. This is not ideal because a consumer of rbx-dom probably expects this to work the same way as Roblox's DOM out of the box, and it's a nontrivial amount of work to both catalogue and correctly translate between proxy properties.

At the same time, this will mean more maintenance burden on us. It would be great to find some data we could use to automate this (in my cursory search of Max's client tracker I was unable to find any), but failing that we could target only the most important proxy properties.

Dekkonot:
Given that we currently expose `Instance.properties` as a public struct member, it isn't possible to keep track of this 100%. However, I think if we added something akin to `Instance.set_property` and `Instance.get_property` it wouldn't be unreasonable to have 'proxies' that could intercept what a property was being read or written as.

However, I think that might be beyond `rbx_dom_weak`. It is, by its nature weakly typed so it knowing about properties like this feels weird. Between things like UniqueId and this, it may be time to start seriously considering what an implementation of `rbx_dom_strong` would look like.

kennethloeffler:
> rbx_dom_strong 

So ... I think we can at least automate cataloguing proxy properties in rbx_reflector by using the injected plugin to set properties on each instance in a defaults place while listening for changes to any other properties that happen as a result. This will give us information about which proxy properties are linked to each other (for every property settable from Lua, or at least most of them) which we can include in the reflection database. 

This does not solve the problem of how to convert to and from proxy properties - at this point that still has to be manual

Dekkonot:
> I'm still not convinced that this would be useful, but this is a conversation for another issue

I think that's basically what we're headed for anyway, since we're finding a need to be aware of properties outside of the reflection database. I'm not saying it need to be a strongly typed DOM (I don't even know what that would look like) but we should seriously consider what the future looks like if we need to be aware of invariants and proxies inside `rbx_dom_weak` itself. If nothing else, there will need to be a layer on top of `Instance`s that is aware of proxies for this to work automatically.

[...]

PR #315, opened 2023-07-07
Patch `Lighting.Clocktime` to serialize

Dekkonot:
Like it says on the tin.


kennethloeffler:
I'm a little lost here - did `Lighting.ClockTime` used to serialize? Does Roblox actually read and set this property on deserialization? Are there any related reports for e.g. Rojo or Lune?

Dekkonot:
So, I was in a hurry when I made this PR and didn't provide as much background as I should have.

Rojo has gotten intermittent reports of `ClockTime` not being serialized (see https://github.com/rojo-rbx/rojo/issues/655 for an example), so this just makes it serialize. It doesn't seem like it's ever been serialized by Roblox, but they deserialize it just fine and have since it was introduced.

Another solution would be to provide a migration of some sort for `ClockTime` to `TimeOfDay`, though I'm not sure how necessary it is. It's unlikely someone will have multiple instances of `Lighting` in a file so there's no compatibility concern.

kennethloeffler:
Not sure about this one, tbh. 

I could totally see someone deserializing a place using Lune, (`Lighting.TimeOfDay` will exist in this case), then setting `Lighting.ClockTime`. Both properties will then exist and not be consistent with each other, similar to the concern I had in #302 with `Workspace.CollisionGroups` and `Workspace.CollisionGroupData`, except probably a lot more likely because the value formats for these lighting properties are a lot more approachable, and it's common for people to use them interchangeably. Can we test what Roblox does on deserialization when `Lighting.TimeOfDay` and `Lighting.ClockTime` are both serialized into the same file?

Dekkonot:
In what can only be described as cursed, Roblox seems to load whichever property appears first in the file. This means `ClockTime` when we build it ourselves.

Given this, I'm not sure what to do. A migration would probably work, though I'm not sure how Rojo would handle a file that had both if migrations were involved.

kennethloeffler (approved):
I think it's fine to serialize both until #277 is more fleshed out... the possible inconsistency kinda sucks, but hopefully it won't be too surprising to users

PR #318, opened 2023-07-09
Add gen-reflection equivalent for windows

Dekkonot:
Though it is heresy to admit it, I do run windows on my machine and have been manually typing out the command to generate the reflection database. I think it'd be cool if I didn't have to do that anymore.

It's functionally identical to the bash script, just for batch.

### DIFF
--- a/gen-reflection.bat
+++ b/gen-reflection.bat
@@ -1,0 +1,8 @@
+@ECHO OFF
+
+IF "%1" EQU "--dry-run" (
+    cargo run --bin rbx_reflector -- generate --patches patches
+) ELSE (
+    cargo run --bin rbx_reflector -- generate rbx_reflection_database/database.msgpack rbx_dom_lua/src/database.json --patches patches
+    cargo run --bin rbx_reflector -- values rbx_dom_lua/src/allValues.json
+)


kennethloeffler:
Are you able to use e.g. Git Bash to run the existing gen-reflection script on Windows?

nezuo:
Yes, the existing script works with Git Bash. I don't think you should need to download Git Bash though.

Dekkonot:
I tried that before opening a PR and ran into a few problems, so I decided to abandon it.

However, I've just now tried again but with mingw64 (what Git uses) instead of cygwin64. It turns out that was the difference and it runs without problem in mingw64. In cygwin, it raises a syntax error which makes me think maybe it's just a bad terminal.

It *did* only run with `./gen-reflection` instead of `gen-reflection` though. It goes without saying that this is the end of the world because it means two more characters vs if I do it with a batch script.

kennethloeffler (approved):
Since it's likely that contributors are working on Windows, I guess it makes sense to also include a batch script - but we'll have to remember to edit both scripts if we ever want to make changes (which should be rare?)

Dekkonot:
I think it'll be fairly rare, though only time will tell. I'm hoping we don't evolve our reflector generation that rapidly.

PR #321, opened 2023-07-16
RollOff Patch

### LINKED ISSUE #320: `Sound.RollOffMinDistance` and `Sound.RollOffMaxDistance` need to be aliased
As shown in rojo-rbx/rojo#742, `Sound.RollOffMinDistance` and `Sound.RollOffMaxDistance` don't save or load. When saved, it seems as though they're directly aliased to `Sound.EmitterSize` and `Sound.xmlRead_MaxDistance_3` respectively.

Assuming nothing else can affect either of those properties, it should be straightforward to patch them to serialize properly.

Barocena:
closes #320 and https://github.com/rojo-rbx/rojo/issues/742

- `RollOffMinDistance` serialize as `EmitterSize`
- `RollOffMaxDistance` serialize as `xmlRead_MaxDistance_3`

### DIFF
--- a/patches/sound.yml
+++ b/patches/sound.yml
@@ -6,3 +6,17 @@ Change:
         As: xmlRead_MaxDistance_3
     xmlRead_MaxDistance_3:
       AliasFor: MaxDistance
+
+    RollOffMinDistance:
+      Serialization:
+        Type: SerializesAs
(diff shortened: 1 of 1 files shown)


Dekkonot:
Thank you!

PR #322, opened 2023-07-18
Add QDir support to rbx_types, rbx_xml, and reflection

kennethloeffler:
This PR adds `QDir` to rbx_types (a simple wrapper around `PathBuf`), `QDir` (de)serialization to rbx_xml, and also allows rbx_reflector to reflect `QDir` type properties. It's geared towards the requirements of kampfkarren/roblox-install#33, but is generally useful when interacting with Roblox Studio settings.


rbx_types/src/qdir.rs
```diff
@@ -0,0 +1,43 @@
...
+impl<'a> QDir {
+    pub fn new() -> Self {
+        QDir {
+            path_buf: PathBuf::new(),
+        }
+    }
+
+    pub fn path_buf(&'a self) -> &'a PathBuf {
+        &self.path_buf
+    }
```
Dekkonot, line 22: It feels more reasonable to return a `&Path` here rather than `&PathBuf`.

kennethloeffler: This method is intended to be an accessor to the `path_buf` field (you can already get an `&Path` via the `AsRef<Path>` implementation). Thinking about it more, maybe this field doesn't really need an accessor?

Dekkonot: I'm not sure what the use is for a field accessor here. Maybe writing, though I don't believe this is how people generally do field mutability in Rust.

Issue #316, opened 2023-07-07
Document and implement `Terrain.MaterialColors` support


Dekkonot:
This is hopefully a self-explanatory issue. `MaterialColors` is a binary property in the same vein as Attributes and Tags, and we should have first-class support for it. This means it should (de)serialize specially like those blobs.

There is an existing spec written by Anaminus located [here](https://github.com/RobloxAPI/spec/blob/master/properties/MaterialColors.md) that we can use as a starter, but we'll have to verify it and I'd also like to document it directly.

hoontee:
Verified that Anaminus' spec is accurate.

PR #323, opened 2023-07-18
Add support for `Terrain.MaterialColors`

Dekkonot:
`MaterialColors` is its own unique binary blob format that we need to support. This PR implements it in the same way we have `Attributes` and `Tags` implemented. Nothing should be very surprising in this implementation, but feel free to ask or provide suggestions for improvement.

--- 

Mostly implements #316. This PR is missing the spec documentation because we don't actually have documentation for Tags yet either and it feels like a good time to write that down as well, but that's outside scope for this PR.


kennethloeffler (changes requested):
Looking pretty ok, go ahead and merge master into this + update the database and then I'll verify your `MaterialColors::decode` and `MaterialColors::encode` implementations

rbx_dom_lua/src/EncodedValue.lua
```diff
@@ -265,6 +265,22 @@ types = {
...
+			local real = {}
+			for name, color in pod do
+				real[Enum.Material[name]] = Color3.fromRGB(color[1], color[2], color[3])
+			end
+			return real
+		end,
+		toPod = function(roblox: { [Enum.Material]: Color3 })
+			local pod = {}
+			for material, color in roblox do
+				pod[material.Name] = { math.floor(color.R * 255), math.floor(color.G * 255), math.floor(color.B * 255) }
```
kennethloeffler, line 279: I think we should use round instead of floor here for better accuracy, similar to the Color3 to Color3uint8 conversion in rust
https://github.com/rojo-rbx/rbx-dom/blob/f899fbad5c0f7326cc852039e42ca926c7281b73/rbx_types/src/basic_types.rs#L409-L417

kennethloeffler (commented):
I'm a little concerned about future compat if new terrain material variants are added - there's not much we can do about it anyway since MaterialColors format uses implicit ordering, but we should keep it in mind

rbx_dom_lua/src/EncodedValue.lua
Dekkonot: For all values in the range `[0, 255]` there's no difference, so I'm not sure it's necessary, but we may as well. I've also clamped the individual components to be in `[0, 1]` since that's something we should probably worry about.

kennethloeffler (approved):

PR #327, opened 2023-07-24
Don't serialize `ScriptGuid`, `UniqueId`, and `HistoryId`

### LINKED ISSUE #314: Some properties should not have defaults
Some properties are meant to be unique and change everytime from Roblox so we shouldn't store a default. As of right now, I'm only aware of two:
- `Instance.UniqueId`
- `LuaSourceContainer.ScriptGuid`

I believe our solution should probably be to not serialize these, full stop, to avoid creating conflicts and to make sure diffs aren't being created for people checking models into source control.

### LINKED ISSUE #324: `UniqueId::now` fallback default in binary serializer can cause `UniqueId` collisions
While investigating filiptibell/lune#69, I found that the current usage of `UniqueId::now` in `SerializerState::fallback_default_value` can cause `UniqueId` collisions. This occurs when multiple instances of the same class exist in a `WeakDom`, and some of the instances contain `UniqueId` properties, while others do not. 

It looks like the reason why is that in the serializer, one `TypeInfo` is created for each class:
https://github.com/rojo-rbx/rbx-dom/blob/d9bf893dfdf15ff60c68bab404f5af71c6e880df/rbx_binary/src/serializer/state.rs#L268-L295

The fallback defaults are then reused among all instances of the class:
https://github.com/rojo-rbx/rbx-dom/blob/d9bf893dfdf15ff60c68bab404f5af71c6e880df/rbx_binary/src/serializer/state.rs#L524-L574

I verified this by running the repro in filiptibell/lune#69 with a version of rbx_binary that uses the "nil `UniqueId`" for the fallback default (all zeros - which Roblox will happily load, even if there are other nil `UniqueId`s)

Dekkonot:
Closes #314 and resolves #324 for now.

I've decided that `LuaSourceContainer.ScriptGuid`, `Instance.UniqueId`, and `Instance.HistoryId` shouldn't serialize for the time being. All they do is make diff noise and they aren't especially useful for Rojo or Lune users at this time.

In the process, I discovered that `rbx_reflector` actually wasn't catching `UniqueId` properties so I've corrected that. I'll add it to my list of things to check in new data type implementations going forward.


kennethloeffler (commented):
Concept and execution looks ok to me...

Should this have a changelog entry in rbx_reflection_database?

Besides package compat, are there any other drawbacks to stripping these properties we should be aware of?

Dekkonot:
I was following in the precedent of not noting patches, but this is unusual and I suppose we should note it down somewhere. Good call.

Packages are the main issue with `UniqueId` and `HistoryId`, since that's all they're used for as of this moment. This was confirmed a few versions ago by the engineer who implemented them. While it's possible that's changed, I don't think it has.

With `ScriptGuid`, as far as I'm aware the only use for it is to preserve the IDE state for scripts. When you close a place in Studio, it saves what scripts you had open and any breakpoints you had for them, and it uses the GUID of scripts to do that. This technically means that we lose what little compatibility we had with the debugger, which would suck, but it only impacts people who were running Studio generated files through rbx-dom and then reopening them in Studio for debugging. Mind you, this also means that those same people had to not be using a workflow that duplicated `ScriptGuid`, since that'd also break it.

I think those are acceptable compatibility concerns because they're not really something we support beyond it being a coincidence.

kennethloeffler (approved):

PR #328, opened 2023-07-26
Migrate part BrickColor properties to BasePart.Color

### LINKED ISSUE #319: BrickColor cannot be set because it aliases to Color without handling the type difference
We have BrickColor set as an alias for Color:
https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_dom_lua/src/database.json#L3422-L3426

However, this means that a property write will attempt to set a BrickColor Enum to the Color property that expects a Color3 value.

>  Setting Part.BrickColor to (BrickColor) Hot pink - setProperty:30
  Error(Roblox: Unable to assign property Color. Color3 expected, got BrickColor
PluginDebugService.user_Rojo.rbxm.Rojo.Packages.RbxDom.PropertyDescriptor:19 function set
PluginDebugService.user_Rojo.rbxm.Rojo.Packages.RbxDom.PropertyDescriptor:77 function write
PluginDebugService.user_Rojo.rbxm.Rojo.Plugin.Reconciler.setProperty:32 function setProperty

kennethloeffler:
Closes #319.

Note that this makes `BasePart.BrickColor` and `BasePart.brickColor` canonical, serializable properties! With this PR, they only turn into `BasePart.Color` on deserialization, but may be serialized as-is.


Dekkonot (approved):
I believe this should be fine, though we'll know soon enough if it isn't.

PR #330, opened 2023-07-27
Use generic interleaving methods for `rbx_binary` core

Dekkonot:
This migrates the various `read_interleaved_XXX_array` and `write_interleaved_XXX_array` functions to use `read_interleaved_bytes` and `write_interleaved_bytes` respectively.

The end result is a lot simpler code and has no cost associated with it (I checked the assembly, both versions give us basically the same instructions). Benchmarking also supports this, as the compilation time on my machine is basically the same before and after once you account for noise.

Also I added documentation to all these functions and the integer transforming functions because I didn't like them sitting there undocumented.

### DIFF
--- a/rbx_binary/src/core.rs
+++ b/rbx_binary/src/core.rs
@@ -122,8 +122,8 @@ pub trait RbxReadExt: Read {
         let mut read = vec![[0; mem::size_of::<i32>()]; output.len()];
         self.read_interleaved_bytes(&mut read)?;
 
-        for (chunk, n) in read.into_iter().zip(output) {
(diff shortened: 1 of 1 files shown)


kennethloeffler:
Massive nit:

The commit message is going to be 93 characters long (21 more characters than the recommended 72) when this gets merged... can you edit the title of this PR so it's more concise, maybe `Use generic interleaving methods in rbx_binary core`?

Dekkonot:
Just for you, yes, I can

kennethloeffler (changes requested):

kennethloeffler (approved):

PR #337, opened 2023-08-02
Make `rbx_binary` use nil UniqueId for default value

### LINKED ISSUE #324: `UniqueId::now` fallback default in binary serializer can cause `UniqueId` collisions

Dekkonot:
Fully resolves #324 by making `UniqueId`'s default a null value, which may appears many times in a file.

Due to #327, it shouldn't affect anyone, but it's more 'correct' behavior anyway, so it's what we should be doing.

Tested in a version where `UniqueId` serializes using the following code:
```rs
use std::fs::File;

use rbx_dom_weak::{types::UniqueId, InstanceBuilder, WeakDom};

fn main() {
    let dom = WeakDom::new(
        InstanceBuilder::new("Workspace").with_children([
            InstanceBuilder::new("Folder")
                .with_property("UniqueId", UniqueId::now().unwrap())
                .with_name("UniqueId set"),
            InstanceBuilder::new("Folder"),
            InstanceBuilder::new("Folder"),
            InstanceBuilder::new("Folder"),
        ]),
    );

    rbx_binary::to_writer(File::create("out.rbxl").unwrap(), &dom, &[dom.root_ref()]).unwrap()
}
```

### DIFF
--- a/rbx_types/src/unique_id.rs
+++ b/rbx_types/src/unique_id.rs
@@ -46,6 +46,8 @@ pub struct UniqueId {
 static INDEX: AtomicU32 = AtomicU32::new(0);
 
 impl UniqueId {
+    /// Returns a non-unique `UniqueId` that has every field set to `0`.
+    /// This value may appear multiple times in a Roblox file safely.
     pub fn non_unique() -> Self {
         Self {
             index: 0,
@@ -54,6 +56,8 @@ impl UniqueId {
         }
     }
 
+    /// Returns a new `UniqueId` with each component set to the passed
+    /// values.
     pub fn new(index: u32, time: u32, random: i64) -> Self {
         Self {
             index,
@@ -62,6 +66,7 @@ impl UniqueId {
         }
     }
 
+    /// Returns a new UniqueId.
     pub fn now() -> Result<Self, UniqueIdError> {
         let time = SystemTime::now()
             .duration_since(*EPOCH)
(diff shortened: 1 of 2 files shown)


rbx_types/src/unique_id.rs
```diff
@@ -46,6 +46,18 @@ pub struct UniqueId {
...
 impl UniqueId {
+    /// Returns a non-unique `UniqueId` that has every field set to `0`.
+    /// This value may appear multiple times in a Roblox file safely.
+    pub fn non_unique() -> Self {
+        Self {
+            index: 0,
+            time: 0,
+            random: 0,
+        }
+    }
```
kennethloeffler, line 57: Can we just call this `nil`? It'd [mirror existing terminology for UUIDs](https://datatracker.ietf.org/doc/html/rfc4122.html#section-4.1.7). UniqueId is just a Roblox-brand UUID, and I think it's just more descriptive

rbx_types/src/unique_id.rs
```diff
@@ -68,14 +81,38 @@ impl UniqueId {
...
+        // || here
+        self.time != 0 || self.index != 0 || self.random != 0
+    }
+
+    /// Returns whether this `UniqueId` represents the non-unique
+    /// `UniqueId` value.
+    pub fn is_non_unique(&self) -> bool {
+        !self.is_unique()
+    }
+
```
kennethloeffler, line 97: Again, I think `is_nil` would be a better better fit here, since `is_unique` doesn't actually do any validation of uniqueness (in other words, non-unique UniqueIds are not necessarily all zeros) and could be misleading

And two methods are probably unnecessary, I think just having `is_nil` and making consumers do `!unique_id.is_nil()` for the inverse is fine

kennethloeffler (changes requested):

rbx_types/src/unique_id.rs
Dekkonot: Sure, I wasn't aware that there was existing terminology for this and didn't want to use `null` because of the implications it has.

rbx_types/src/unique_id.rs
```diff
@@ -68,14 +81,30 @@ impl UniqueId {
         })
     }
 
+    /// Returns whether this `UniqueId` is 'nil' or not. That is, whether
+    /// every field of the UniqueId is set to `0`.
+    pub fn is_nil(&self) -> bool {
+        self.time != 0 && self.index != 0 && self.random != 0
```
kennethloeffler, line 87: Fix this oopsie and then LGTM
```suggestion
        self.time == 0 && self.index == 0 && self.random == 0
```

kennethloeffler (changes requested):

kennethloeffler (approved):

PR #338, opened 2023-08-03
Give BinaryStrings their own documentation

Dekkonot:
`MaterialColors` and `Tag` aren't documented right now. This documents them and also moves `BinaryStrings` their doc page. As a byproduct, I've moved attributes to there too.

I'm not 100% on the formatting I've used for the binary strings spec but hopefully it's good enough.


sasial-dev (approved):
LGTM!
It's clear and simple to understand - I can't ask for much more!

Issue #335, opened 2023-08-01
#332 breaks compatibility


kennethloeffler:
#332 is breaking for users who use `IgnoreGuiInset` or `Font` properties in Rojo project.json files or Lune scripts. Current releases serialize these properties, but upcoming releases will not. `BrickColor`/`brickColor` are not affected because in current releases, they are aliases and therefore do not serialize. 

It'd be really easy to add warnings to these places to at least notify users when they try to serialize migrated properties:
https://github.com/rojo-rbx/rbx-dom/blob/7b3e6fb84ceb3b02c5042d17a373f7cd42bec83b/rbx_binary/src/serializer/state.rs#L304-L318
https://github.com/rojo-rbx/rbx-dom/blob/7b3e6fb84ceb3b02c5042d17a373f7cd42bec83b/rbx_xml/src/core.rs#L53-L62

Alternatively, we could maintain compatibility by performing migrations during serialization, but that is a lot more effort, and it's not clear whether it's worth it since we don't know how many people will break, and in the future, proxy properties will render it moot.

Dekkonot:
As annoying as it is, I think the correct path is probably to run migrations during serialization as well. I don't love doing that because like you said, it's work we're going to immediately make obsolete, but the other option is pushing a release that we know will break things, which I feel is worse.

kennethloeffler:
Cool, I'll get started on it right now and try to get it finished tonight. Pray for me, our serializers are not really set up for this 😄

PR #339, opened 2023-08-03
Perform property migrations in rbx_binary serializer

kennethloeffler:
Related: #335 

Yeah, so this sucked pretty hard. rbx_xml will be easier (I hope)...

I've added descriptions to some of the commits here in an attempt to elucidate my thought process, if anything is unclear please ask questions

### DIFF
--- a/rbx_binary/src/serializer/state.rs
+++ b/rbx_binary/src/serializer/state.rs
@@ -410,6 +410,8 @@ impl<'dom, W: Write> SerializerState<'dom, W> {
                 if !prop_info.aliases.contains(prop_name) {
                     prop_info.aliases.insert(prop_name.clone());
                 }
+
+                prop_info.migration = migration;
             }
         }
 
--- a/rbx_binary/src/serializer/state.rs
+++ b/rbx_binary/src/serializer/state.rs
@@ -611,6 +611,16 @@ impl<'dom, W: Write> SerializerState<'dom, W> {
                         // reasonable default.
                         Cow::Borrowed(prop_info.default_value.borrow())
                     })
+                    .map(|value| {
+                        if let Some(migration) = prop_info.migration {
+                            match migration.perform(&value) {
+                                Ok(new_value) => Cow::Owned(new_value),
+                                Err(_) => value,
+                            }
+                        } else {
+                            value
+                        }
+                    })
                     .enumerate();
 
                 // Helper to generate a type mismatch error with context from
--- a/rbx_binary/src/serializer/state.rs
+++ b/rbx_binary/src/serializer/state.rs
@@ -137,6 +137,9 @@ struct PropInfo {
     /// present, followed by an educated guess based on the type of the value.
     default_value: Cow<'static, Variant>,
 
+    /// If a logical property has a migration associated with it (i.e. BrickColor ->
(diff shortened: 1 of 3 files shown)


rbx_binary/src/serializer/state.rs
```diff
@@ -580,6 +611,16 @@ impl<'dom, W: Write> SerializerState<'dom, W> {
...
                     })
+                    .map(|value| {
+                        if let Some(migration) = prop_info.migration {
+                            match migration.perform(&value) {
+                                Ok(new_value) => Cow::Owned(new_value),
+                                Err(_) => value,
+                            }
+                        } else {
+                            value
+                        }
```
Dekkonot, line 622: Migrations seem unlikely to fail anyway, so I think this is fine. The one we might run into is someone setting an invalid `Font` enum value, but I don't think that's something we can reasonably account for right now.

Dekkonot (changes requested):
The new `migration` field on `PropInfo` needs documentation but otherwise looks good.

rbx_binary/src/serializer/state.rs
kennethloeffler: I think this might be a little stranger than you're realizing - the Err branch will be hit under normal operation. The reason why is because in this PR, all migrated properties are collated under a single `PropInfo`. For reference, here's what `PropInfo` looks like during the migrated properties test:
```
PropInfo {
    prop_type: Color3uint8,
    serialized_name: "Color3uint8",
    aliases: {
        "BrickColor",
        "brickColor",
    },
    default_value: Color3uint8(
        Color3uint8 {
            r: 163,
            g: 162,
            b: 165,
        },
    ),
    migration: Some(
        PropertyMigration {
            new_property_name: "Color",
            migration: BrickColorToColor,
        },
    ),
}
```
This means that in `serialize_properties`, some properties will be `Variant::BrickColor`, some properties will be `Variant::Color3uint8` or `Variant::Color3`, and the migration will fail - but we expect it to!

Dekkonot: Oh that's kinda gross. I wonder if it's worth attempting a conversion in the same way we do with rbx_xml or if it's just a lost cause?

kennethloeffler: Since this is likely throwaway code I'm not sure if it matters too much - any mismatches will be caught later in the `match prop_info.prop_type`.  I tried to make this less ugly, but I wasn't able to do so while still having one prop chunk per logical property. I don't think there's a good way to do it without significantly changing the binary serializer

Dekkonot: I suppose another option is we make the migration catch multiple types of variants? That feels less technically correct but is probably safer anyway.

kennethloeffler: I don't think that's necessary. Besides applying migrations for select properties, there's no other changes to the behavior of `serialize_properties` as a result of this, and I'll want to delete this code ASAP regardless

Dekkonot: That's fair enough. We're probably good to merge this then.

Dekkonot (approved):

PR #340, opened 2023-08-04
Perform property migrations in rbx_xml serializer

Dekkonot:
A companion to #339 and related to #335 as well.

As hoped for in 339, this was in fact much easier. We already have a mechanism for replacing names and values during serialization in rbx_xml, so it wasn't hard to add migrations to that process.

Unlike in the rbx_binary implementation, there's no compatibility concern if a migration fails. I've noted that in the source, but for posterity, if one fails and we allow a property to not be migrated, it'll still load just fine in Studio so there's not any real concern.

### DIFF
--- a/rbx_xml/src/tests/snapshots/rbx_xml__tests__basic__migrated_properties.snap
+++ b/rbx_xml/src/tests/snapshots/rbx_xml__tests__basic__migrated_properties.snap
@@ -16,6 +16,7 @@ expression: "std::str::from_utf8(&encoded).unwrap()"
     <Item class="ScreenGui" referent="2">
       <Properties>
         <string name="Name">ScreenGui</string>
+        <bool name="IgnoreGuiInset">true</bool>
       </Properties>
     </Item>
     <Item class="Part" referent="3">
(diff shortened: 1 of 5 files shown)


rbx_xml/src/tests/snapshots/rbx_xml__tests__basic__migrated_properties.snap
```diff
@@ -0,0 +1,65 @@
...
+    <Item class="Part" referent="3">
+      <Properties>
+        <string name="Name">Part</string>
+        <Color3uint8 name="Color3uint8">16777215</Color3uint8>
+      </Properties>
+    </Item>
+    <Item class="Part" referent="4">
+      <Properties>
+        <string name="Name">Part</string>
+        <Color3uint8 name="Color">11829503</Color3uint8>
```
kennethloeffler, line 31: Maybe the new property for `PropertyMigration::BrickColorToColor` should really be `Color3uint8`? 🤔 It'll still work regardless though...

Dekkonot: I suspect yes, it should be. With the way Roblox handles reflection, it turns out to be fine, but we should probably fix that for correctness's sake.

kennethloeffler (approved):

PR #343, opened 2023-08-04
Update License year and copyright holder

Dekkonot:
Since we are prepping for a new release, it makes sense for us to change this; all subsequent versions of rbx-dom starting from the next release will be licensed to the Rojo authors rather than specifically LPG.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Lucien Greathouse
+Copyright (c) 2023 The Rojo Developers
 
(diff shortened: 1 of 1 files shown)


kennethloeffler:
Did you get permission from Lucien to do this?

Dekkonot:
I have, though the response is still pending. If the response is negative, we'll have to reconsider what that looks like but I'm optimistically assuming it will be positive.

Dekkonot:
It seems as though I'm not going to get a response one way or another on this, so I'll shelve this for now. We should consider what the future looks like though.

PR #344, opened 2023-08-04
Update reflection database to v588

Dekkonot:
Like it says on the tin.


Dekkonot:
We should probably develop a more reasonable strategy for diffing reflection databases. It seems like git might not be up to the task!

rbx_dom_lua/src/database.json
```diff
@@ -47969,7 +49040,113 @@
           "Enum": 256
         },
         "MaterialColors": {
-          "BinaryString": "AAAAAAAAan8/P39rf2Y/ilY+j35fi21PZmxvZbDqw8faiVpHOi4kHh4lZlw76JxKc3trhHtagcLgc4RKxr21zq2UlJSM"
+          "MaterialColors": {
```
kennethloeffler, line 49043: Why are we only seeing this now? o_O

Dekkonot: I... am not sure. It clearly updated in my [PR for MaterialColors](https://github.com/rojo-rbx/rbx-dom/pull/323/commits/564251622b194f9e2ae0163aa8dabf4919f25d42#diff-ea4e1c89d66af15167a9d1bbe0179ebbe85c21b512c69d69797345e00054abb4R47799) but it seems like the database regeneration in the PR may have just erased it?

Concerning.

kennethloeffler: Looks like [this merge ](https://github.com/rojo-rbx/rbx-dom/pull/323/commits/905687fcbab4c3fec90a41bd1b204b9aca45321f) removed it, but not sure why the following generation didn't put it back... it's probably ok? 😓

Dekkonot:
I think I have a hypothesis based on the commit I just pushed: It doesn't know about new properties until a second run for whatever reason, so it can't store their defaults unless it's ran twice

kennethloeffler (approved):

PR #358, opened 2023-09-14
Implement SecurityCapabilities in rbx_types

kennethloeffler:
This PR implements `SecurityCapabilities` in rbx_types. Its functionality is very limited right now because we have almost no idea what this datatype does or how to interact with it!

I want rojo-rbx/rbx-test-files#23 merged before adding functionality for `SecurityCapabilities` to any other of the crates. dekko pls review it thoroughly and make sure I got it right


Dekkonot (changes requested):
We should put this in the changelog, but this is overall a good barebones implementation for it.

We should also provide a method to get the inner `value` though, since we'll need it for serializing. I'll leave the exact name up to you.

kennethloeffler:
> We should also provide a method to get the inner value though, since we'll need it for serializing. I'll leave the exact name up to you.

We won't need any accessor for parsing - rbx_xml can use `u64::parse` and then pass the result to the `SecurityCapabilities` constructor. I'd prefer to keep it totally locked away from consumers for now since we may change to a different type in the future once we understand the meaning of the bits

This also has me thinking... maybe `SecurityCapabilities::new` should instead be called `SecurityCapabilities::from_bits`?

kennethloeffler:
Ok, I just removed the Display impl and added a `bits` accessor. This is probably better than Display because if we switch to bitflags then we'd probably want to Display to change too, forward compat with `bits` should work out fine, and it's similar to existing bit field types like `Axes` and `Faces`

kennethloeffler:
Should `SecurityCapabilities` not implement `Serialize` or `Deserialize` just yet? The format may well change in the future...

Dekkonot (approved):
I'm not thrilled with implementing `Serialize`/`Deserialize` but I suppose we have to for `Variant`. With that in mind, this looks fine to me.

kennethloeffler:
Cool, I'll put up PRs for reflection, rbx_binary, and rbx_xml later today

PR #359, opened 2023-09-16
Add rbx_xml SecurityCapabilites parsing, add to reflection

kennethloeffler:
Closes #356 

This PR adds `SecurityCapabilities` support to rbx_xml and adds reflection information for `SecurityCapabilities`.

PR #360, opened 2023-09-19
Update rbx_xml and rbx_reflection_database changelogs

kennethloeffler:
I forgot to add entries to the changelogs in #359 😔


Dekkonot (approved):
It might be worth adding some sort of changelog protection because I also just straight up forgot to check!

PR #364, opened 2023-09-22
Implement method for cloning multiple instances into external dom

kennethloeffler:
Related: filiptibell/lune#110

It is not possible to preserve refs that point across subtrees using `WeakDom::clone_into_external` because a new `CloneContext` is used for each invocation - this new method gives a way to use one `CloneContext` for multiple subtrees!

Not sure about the method name, docs could probably be a little better, etc.

### DIFF
--- a/rbx_dom_weak/src/dom.rs
+++ b/rbx_dom_weak/src/dom.rs
@@ -282,7 +282,7 @@ impl WeakDom {
         root_ref
     }
 
-    /// Similar to clone_into_external, but clones multiple subtrees all at once. This
+    /// Similar to `clone_into_external`, but clones multiple subtrees all at once. This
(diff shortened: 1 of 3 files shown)


rbx_dom_weak/src/dom.rs
```diff
@@ -278,6 +282,26 @@ impl WeakDom {
         root_ref
     }
 
+    /// Similar to clone_into_external, but clones multiple subtrees all at once. This
```
Dekkonot, line 285:
```suggestion
    /// Similar to `clone_into_external`, but clones multiple subtrees all at once. This
```

Dekkonot (changes requested):
Add to changelog + comment nitpick

Dekkonot (approved):

PR #367, opened 2023-10-03
Remove mention of MSRV from README

Dekkonot:


Dekkonot:
looks good to me thank you dekkonot

PR #370, opened 2023-10-11
Use ScriptEditorService for updating Source

Dekkonot:
We need to start using `UpdateSourceAsync` and `GetEditorSource` for setting and getting `Script.Source` and `ModuleScript.Source`. So, this does that, using `task.spawn` to make sure it isn't blocking as we discussed elsewhere.

This is not a permanent solution. Long-term, we should try to properly support property sets and gets being async, or properly commit to consumers like Rojo having to handle it. For now though, this should be fine.

Also I fixed a lint because it was bothering me. Let me know if I should revert that and fix it in a different PR lol


nezuo:
Doesn't this need to also support `LocalScript`?

Dekkonot:
[LocalScript inherits from Script](https://github.com/MaximumADHD/Roblox-Client-Tracker/blob/roblox/API-Dump.txt#L2876), so it already does. It's a surprising inheritance but very helpful.

rbx_dom_lua/src/customProperties.lua
```diff
@@ -116,4 +117,36 @@ return {
 			end,
 		},
 	},
+
```
nezuo, line 120: There aren't newlines between the previous entries.

nezuo (changes requested):

nezuo (approved):

PR #372, opened 2023-10-20
Remove erroring SecurityCapabilities lua coder/decoder

kennethloeffler:
It looks like this coder/decoder was introduced in imitation of some other types, but the other types that have erroring implementations either:
* Do not exist on any instance (`Region3`)
* Are special-cased elsewhere downstream (`Ref`)
* Wouldn't work anyway? (`SharedString`, although I think there is a case to be made that `SharedString` should NOT have an erroring lua coder/decoder)

`SecurityCapabilities` exists on *all* instances so an erroring implementation is inappropriate and causes problems for Rojo as per rojo-rbx/rojo#802. I think we can just remove the lua encoder/decoder, like how `UniqueId` simply doesn't have them either.

### DIFF
--- a/rbx_dom_lua/src/EncodedValue.lua
+++ b/rbx_dom_lua/src/EncodedValue.lua
@@ -412,15 +412,6 @@ types = {
 		end,
 	},
 
-	SecurityCapabilities = {
-		fromPod = function(_pod)
-			error("SecurityCapabilities is not implemented")
-		end,
-		toPod = function(_roblox)
-			error("SecurityCapabilities is not implemented")
-		end,
-	},
-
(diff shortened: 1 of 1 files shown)


Dekkonot (approved):
It does feel as though this is fixing a symptom rather than the underlying problem, since we should really not be sending properties we know aren't scriptable through Rojo to begin with and this wouldn't be a problem if we didn't. That said, this is also an issue so I'm okay with it.

In the future it may be worth looking into whether we can have a better mechanism for distinguishing between "unimplemented" and "unknown" for rbx_dom_lua, but that day is not today.

PR #375, opened 2023-11-09
Allow a database to be specified in both rbx_xml and rbx_binary

Dekkonot:
This allows the reflection database to be specified for serializing/deserializing in `rbx_binary` and `rbx_xml`. It does this by having the structs for these crates store a reflection database rather than using `rbx_reflection_database` inline. This doesn't change how the crates are used normally since they use the bundled database by default, but it provides the opportunity for tools like Rojo to specify their own database.

I took some liberties here also, and cleaned up the name of lifetimes a bit. Specifically, I've changed it so that all lifetimes that refer to a reflection database use a `'db` lifetime and all lifetimes that refer to a `WeakDom` use a `'dom` lifetime. This is to avoid confusion because `'a` is not specific or helpful.

### DIFF
--- a/rbx_binary/CHANGELOG.md
+++ b/rbx_binary/CHANGELOG.md
@@ -1,6 +1,7 @@
 # rbx_binary Changelog
 
 ## Unreleased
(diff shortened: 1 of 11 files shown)


nezuo (approved):
Looks good to me! I appreciate the approach taken in the PR as it doesn't result in any breaking changes.

PR #376, opened 2023-11-13
Change rbx_reflection_database to support loading from FS

Dekkonot:
As part of the effort to get the reflection database loadable locally, it's probably a good idea to have a central place that decides locations for the entire ecosystem. This will prevent consumers like Lune and Rojo from having to re-implement this logic.

The obvious choice for this, in my opinion, is rbx_reflection_database. It's a requirement for everything that uses the reflection database already, and its public API can be extended to take things from the file system into account without any issues. I've implemented it and documented it thoroughly in the README for rbx_reflection_database.

However, it also sacrifices the 'purity' of the module because it fundamentally alters its purpose. I'm okay with that, but I understand not everyone will be. Please be honest with your thoughts.

I know for sure this works on Windows and Linux. I'm assuming it will also work on MacOS, but I don't own a machine I can use for testing. I don't know how to test this in CI, so we'll just have to be mindful of changes.


Corecii (changes requested):
Additional comment:
Once tools are doing automatic updates, we'll need to be able to replace the database that `rbx_reflection` returns without restarting the tool. It's _probably_ fine to provide a method to "reload" the database, which just loads another version of the database permanently into memory, as long as no one is "reloading" the database tens of times in a session. Typically I'd expect one reload from a tool after it downloads a database update.

rbx_reflection_database/src/lib.rs
```diff
@@ -1,23 +1,148 @@
...
+/// not be up to date, but it will always exist.
+pub fn get_bundled() -> &'static ReflectionDatabase<'static> {
+    &BUNDLED_DATABASE
+}
+
+/// Fetches the location a [`ReflectionDatabase`] is expected to be loaded from.
+/// This may return [`None`] if the local data directory cannot be found.
+pub fn get_local_location() -> Option<PathBuf> {
+    if let Ok(location) = env::var(OVERRIDE_PATH_VAR) {
+        log::debug!("Using enviromental variable {OVERRIDE_PATH_VAR} to fetch reflection database");
```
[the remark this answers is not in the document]
Dekkonot: After thinking about it, I'm not sure I ever have either but I was just in autopilot mode and corrected for grammar I guess. Good catch.

rbx_reflection_database/src/lib.rs
```diff
@@ -1,23 +1,148 @@
...
+pub const LOCAL_DIR_NAME: &str = ".rbxreflection";
+
 lazy_static::lazy_static! {
-    static ref DATABASE: ReflectionDatabase<'static> = {
+    static ref BUNDLED_DATABASE: ReflectionDatabase<'static> = {
+        log::debug!("Loading bundled reflection database");
         rmp_serde::decode::from_slice(ENCODED_DATABASE).unwrap_or_else(|e| panic!("could not decode reflection database because: {}", e))
     };
+
+    static ref LOCAL_DATABASE: Option<ReflectionDatabase<'static>> = {
```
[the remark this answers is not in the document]
Dekkonot: Storing it in something that has interior mutability and providing a setter wouldn't be too bad. My gut says `OnceLock<Option<ReflectionDatabase<'static>>>`, which is a mouthful but would work well for what we want.

rbx_reflection_database/src/lib.rs
```diff
@@ -1,23 +1,148 @@
...
+        let location = get_local_location()?;
+        if let Ok(file) = fs::read(&location) {
+            log::debug!("Loading local reflection database from {}", location.display());
+            Some(
+                rmp_serde::decode::from_slice(&file).unwrap_or_else(|e| {
+                    panic!("could not decode reflection database because: {}", e)
```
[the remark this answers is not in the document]
Dekkonot: I did actually consider just using a a `Result` rather than panicking for exactly this reason. It's not an unfounded worry because we've seen problems with a file exactly like this for Selene.

The reason I didn't commit to it out of the gate was because I didn't want to force the `get` to suddenly be a fallible operation, but I suppose it actually is now and lying isn't doing anyone favors. I'm willing to commit to it now so we only have to break things once.

Dekkonot:
Allowing the database to be replaced once it's initialized is complicated because it involves mutating a static variable, so for the time being I'm going to ignore it. Tools like Rojo and Lune can work out solutions without that.

rbx_reflection_database/src/lib.rs
```diff
@@ -1,15 +1,144 @@
...
+/// Returns a populated [`ReflectionDatabase`]. This will attempt to load one locally and
+/// if one can't be found, it will return one that is bundled with this crate.
+///
+/// ## Errors
+///
+/// Errors if a locally stored [`ReflectionDatabase`] could not be read
+/// or is invalid MessagePack.
+pub fn get() -> Result<&'static ReflectionDatabase<'static>, Error> {
+    Ok(get_local()?.unwrap_or(&BUNDLED_DATABASE))
+}
```
kennethloeffler, line 75: I'm not totally sure if breaking this interface is the right call. The way I'm thinking about it is: if `get_local` fails, then probably 99% of the time, users of this function will immediately want to call `get_bundled`, so why not just have this function do that for them? That way it does the right thing most of the time, and we can avoid breakage.

On the other hand, then it's not clear that this function can fail, and removes the ability to e.g. print a warning if the local database is corrupt while using this function.

kennethloeffler (commented):
Looking good, I left a few comments.

RE testing - it looks like the only code paths not covered right now are here:
https://github.com/rojo-rbx/rbx-dom/blob/d95bc804238e4f10418d3d6d03206491d03f4840/rbx_reflection_database/src/lib.rs#L130-L141

Maybe we could add a test that simply compares the result of this to the expected path for the platform, and if we're very paranoid, add some CI targets for macOS and Windows?

rbx_reflection_database/src/lib.rs
Dekkonot: I'm not totally sold on breaking this either, but the other option is hiding failures from users. It might be worth taking a middleground and just emitting a message into stderr if the local database exists but is corrupted?

It may also be worth caching it if we do this though, so it doesn't print the message every time.

LouieK22:
We should probably use a more generic top-level folder name (e.g. %localappdata%/rbx_dom on Windows, and ~/.rbx_dom on Mac/Linux) as it would give a better hint as to where these files came from. The reflection cache could be something like rbx_dom/cache/reflection.proto

This crate might come in handy for picking directories: https://crates.io/crates/directories

Dekkonot:
It's currently placed in `%localappdata%/.rbx_reflection` on Windows, `$HOME/Library/Application Support/.rbxreflection` on MacOS, and `$HOME/.rbxreflection` on Linux. `.rbx_dom` or something works too. It's not really important.

rbx_reflection_database/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # rbx\_reflection_database Changelog
 
 ## Unreleased Changes
+- The database may now be loaded dynamically from the local file system. ([#376])
+  The location is OS-dependent but it will only be loaded if one exists. The location may also be manually specified using the `RBX_DATABASE` environment variable.
+
+  `get` is unchanged in use, but will return a locally stored database if it exists, and the bundled one if not. Two new methods were added: `get_bundled` will only fetch the local database and `get_local` will only fetch a locally stored one.
```
kennethloeffler, line 7: This is not accurate to the new implementation, because rbx_reflection_database::get now returns a Result. That's a breaking change, which should be pointed out.

rbx_reflection_database/src/lib.rs
```diff
@@ -1,15 +1,147 @@
...
+    }
+}
+
+/// Returns the default local location for the reflection database, without
+/// considering the env variable.
+fn get_local_location_no_var() -> Option<PathBuf> {
+    // Due to concerns about the local data directory existing
+    // on Linux, we use the home directory instead.
+    #[cfg(target_os = "linux")]
+    let mut home = dirs::home_dir()?;
```
kennethloeffler, line 138: Can we fall back to the home directory only if the local data directory is None?

kennethloeffler (changes requested):
Looks good, and will work well for consumers that want to update the reflection database without shipping a new binary.

I think the current data directory choosing scheme works fine, probably don't need to add a library for our uses

rbx_reflection_database/src/lib.rs
Dekkonot: The concern is that it will often be `Some` but it won't actually exist. This happens in WSL, as an example.

kennethloeffler (approved):

PR #378, opened 2023-11-22
Use in-progress db for rbx_reflector default place decode

### LINKED ISSUE #345: rbx_reflector is inconsistent with storing default values for properties
As discovered in #344, defaults for properties seem to not be always filled in on the first run of rbx_reflector. This seems to be resolved by running it multiple times, but it's spooky behavior so we should fix or mitigate it.

kennethloeffler:
Closes #345

Because of #375, it's possible to continue using `DecodePropertyBehavior::IgnoreUnknown` and simply swap to the in-progress database via `DecodeOptions::reflection_database`


nezuo (approved):
Nice!

Dekkonot (approved):

PR #379, opened 2023-12-17
Add `Default` implementation for `WeakDom`

DervexHero:
Title basically explains everything. Aside from the fact that it is good practise, `Default` is required e.g. when using `#[serde(skip)]` attribute.


Dekkonot:
Can you add an entry to the changelog for `rbx_dom_weak` for this? Thanks!

DervexHero:
Sure, added!

Dekkonot (approved):

Issue #313, opened 2023-07-06
Improved error messages for instance names


filiptibell:
Currently, `rbx-dom` throws a pretty nondescript error when an instance has a name containing invalid utf-8:

```
[TRACE rbx_binary::deserializer::state] Known prop, canonical name C0 and type CFrame
[TRACE rbx_binary::chunk] Chunk "PROP" (compressed: 24, len: 37, reserved: 0)
[TRACE rbx_binary::deserializer::state] PROP chunk (RotateP.C1, instance type 59, prop type 59
[TRACE rbx_binary::deserializer::state] Known prop, canonical name C1 and type CFrame
[TRACE rbx_binary::chunk] Chunk "PROP" (compressed: 20, len: 18, reserved: 0)
[TRACE rbx_binary::deserializer::state] PROP chunk (RotateP.Enabled, instance type 59, prop type 59
[TRACE rbx_binary::deserializer::state] Known prop, canonical name Enabled and type Bool
[TRACE rbx_binary::chunk] Chunk "PROP" (compressed: 30, len: 50, reserved: 0)
[TRACE rbx_binary::chunk] Chunk "PROP" (compressed: 69, len: 103, reserved: 0)
[TRACE rbx_binary::deserializer::state] PROP chunk (RotateP.Name, instance type 59, prop type 59

stream did not contain valid UTF-8
```

This particular instance is more common in Roblox than one might think and is known as a "virus" instance, one created by a malicious script or plugin to be intentionally annoying or disruptive. Developers that get this error generally don't know that they have these, and the instances are also generally harmless, but make deserialization in `rbx-dom` fail with the above error message.

It would be nice to throw a better error here (and maybe in other cases for invalid utf-8), maybe even mentioning the exact instance and where it is so developers can deal with it accordingly.

kennethloeffler:
Looked into this more, and it's going to be difficult (or impossible) to give much useful info (e.g. the instance's full name) in these errors because the DOM is necessarily incomplete in the middle of a deserialization operation.

Maybe instead of more specific errors, we could just do a lossy conversion for non UTF-8 `Variant::String` properties? The main drawback is just the fact the original data would be altered, and there may be legitimate uses for non UTF-8 string properties!

Alternatively, we could start reading `Variant::String` props into a `Vec<u8>` instead of a `String`. This would be more correct, but it would be a breaking change and makes string properties more complicated to work with for consumers.

filiptibell:
[...]

PR #380, opened 2024-01-06
Fix "stream did not contain valid UTF-8" using String::from_utf8_lossy

krakow10:
I am using `rbx_binary::from_reader(input)` to load files in my [rust program](https://git.itzana.me/StrafesNET/asset-tool), and some of the files make the function spit out the error "stream did not contain valid UTF-8".

I decided that I'm proficient enough at Rust to find out how to fix this pesky issue myself, and here's what I've come up with.  I found two places which would produce this error and have patched them using the `String::from_utf8_lossy` function.

Here is an image of an offending section of a `Script.Source` property after it was mercilessly butchered into valid UTF-8:
image

### DIFF
--- a/rbx_binary/src/deserializer/state.rs
+++ b/rbx_binary/src/deserializer/state.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Cow,
     collections::{HashMap, HashSet, VecDeque},
     convert::TryInto,
     io::Read,
@@ -374,7 +375,20 @@ impl<'db, R: Read> DeserializerState<'db, R> {
 
             for referent in &type_info.referents {
                 let instance = self.instances_by_ref.get_mut(referent).unwrap();
-                let value = chunk.read_string()?;
+                let binary_string = chunk.read_binary_string()?;
+                let value = match std::str::from_utf8(&binary_string) {
+                    Ok(value) => Cow::Borrowed(value),
+                    Err(_) => {
+                        log::warn!(
+                            "Performing lossy string conversion on property {}.{} because it did not contain UTF-8.
+This may cause unexpected or broken behavior in your final results if you rely on this property being non UTF-8.",
+                            type_info.type_name,
+                            prop_name
+                        );
+
+                        String::from_utf8_lossy(binary_string.as_ref())
+                    }
+                };
                 instance.builder.set_name(value);
             }
 
@@ -399,8 +413,22 @@ impl<'db, R: Read> DeserializerState<'db, R> {
                 VariantType::String => {
                     for referent in &type_info.referents {
                         let instance = self.instances_by_ref.get_mut(referent).unwrap();
-                        let value = chunk.read_string()?;
-                        add_property(instance, &property, value.into());
+                        let binary_string = chunk.read_binary_string()?;
+                        let value = match std::str::from_utf8(&binary_string) {
+                            Ok(value) => Cow::Borrowed(value),
+                            Err(_) => {
+                                log::warn!(
+                            "Performing lossy string conversion on property {}.{} because it did not contain UTF-8.
+This may cause unexpected or broken behavior in your final results if you rely on this property being non UTF-8.",
+                                    type_info.type_name,
+                                    property.name
+                                );
+
+                                String::from_utf8_lossy(&binary_string)
+                            }
+                        };
+
+                        add_property(instance, &property, value.as_ref().into());
                     }
                 }
                 VariantType::Content => {
--- a/rbx_binary/CHANGELOG.md
+++ b/rbx_binary/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 * Added the ability to specify what type of compression to use for serializing. This takes the form of `Serializer::compression_type`. ([#446])
 * Added support for ZSTD compressed files ([#446])
+* Implicit lossy conversion of non-UTF-8 `Instance.Name` and `*Script.Source` properties when decoding.  The previous behaviour was returning an error. ([#380])
 
 [#446]: https://github.com/rojo-rbx/rbx-dom/pull/446
 


krakow10:
I'm testing this and it didn't entirely fix the issue, only one case.  I will keep investigating.

Dekkonot:
Hi, what problem are you running into and trying to fix? Right now I have no idea beyond that you're evidently having trouble with non-UTF8 names for Instances?

krakow10:
I have .rbxl files which when I decode with
```Rust
rbx_binary::from_reader(input)
```
Throw this UTF-8 error.

Dekkonot:
Just to clarify, is there a reason why you want/need invalid UTF-8 for the name property on Instances?

We've had discussions about this before internally and it's unclear if it's actually beneficial to support names like that. If we were going to, we'd probably want to swap to using a string replacement like [`BStr`](https://crates.io/crates/bstr) over lossy conversions though.

krakow10:
So far I have found the issue with both Name and Source property.  My current project is that I'm converting the 1006 .rbxl files I downloaded off my place version history into git version history, but I've also run into this issue with map files for my actual game which I wanted to scan for scripts.  So it's both reading files I created a very long time ago, and reading files that I didn't create.

krakow10:
All 1006 files have decoded with these changes!

krakow10:
I updated the original message with some contextual information, sorry about the lack of clarity initially.

Dekkonot:
Going to tag @kennethloeffler in so that he can share his thoughts on this. I don't want to leave you hanging on this and between us we should be able to make a final call on how we want to solve this problem.

kennethloeffler:
We've had a bit of discussion about this aspect of rbx-dom already: https://github.com/rojo-rbx/rbx-dom/issues/313. 

A lossy conversion is okay, but is... well, lossy. Roblox's implementation doesn't do any conversions whatsoever on `String` property contents, and diverging from that doesn't feel great to me. It could lead to busted results - for example, when Lua code expects a certain non UTF-8 instance name, when binary data is stored in a `StringValue`, etc.

I think I'm okay with doing a lossy conversion so rbx-dom can successfully-ish decode more files, as long as we document it and print a warning when we have to do it so that there are no murky surprises. But we should really consider bringing rbx-dom's `String` properties up to parity with Roblox's by representing them as `Vec<u8>` or similar. That is a breaking change, so we'll have to think about how it fits into the next major version.

rbx_binary/src/deserializer/state.rs
```diff
@@ -374,7 +374,7 @@ impl<'db, R: Read> DeserializerState<'db, R> {
 
             for referent in &type_info.referents {
                 let instance = self.instances_by_ref.get_mut(referent).unwrap();
-                let value = chunk.read_string()?;
+                let value = String::from_utf8_lossy(chunk.read_binary_string()?.as_ref()).to_string();
```
kennethloeffler, line 377:
```suggestion
                let binary_string = chunk.read_binary_string()?;
                let value = match std::str::from_utf8(&binary_string) {
                    Ok(value) => Cow::Borrowed(value),
                    Err(_) => {
                        log::warn!(
                            "Performing lossy string conversion on property {}.{} because it did not contain UTF-8.
This may cause unexpected or broken behavior in your final results if you rely on this property being non UTF-8.",
                            type_info.type_name,
                            prop_name
                        );

                        String::from_utf8_lossy(binary_string.as_ref())
                    }
                };
```

rbx_binary/src/deserializer/state.rs
```diff
@@ -399,7 +399,7 @@ impl<'db, R: Read> DeserializerState<'db, R> {
                 VariantType::String => {
                     for referent in &type_info.referents {
                         let instance = self.instances_by_ref.get_mut(referent).unwrap();
-                        let value = chunk.read_string()?;
+                        let value = String::from_utf8_lossy(chunk.read_binary_string()?.as_ref()).to_string();
                         add_property(instance, &property, value.into());
```
kennethloeffler, line 403:
```suggestion
                        let binary_string = chunk.read_binary_string()?;
                        let value = match std::str::from_utf8(&binary_string) {
                            Ok(value) => Cow::Borrowed(value),
                            Err(_) => {
                                log::warn!(
                            "Performing lossy string conversion on property {}.{} because it did not contain UTF-8.
This may cause unexpected or broken behavior in your final results if you rely on this property being non UTF-8.",
                                    type_info.type_name,
                                    property.name
                                );

                                String::from_utf8_lossy(&binary_string)
                            }
                        };

                        add_property(instance, &property, value.as_ref().into());
```

kennethloeffler (changes requested):
Import `std::borrow::Cow` at the top of `rbx_binary/src/deserializer/state.rs` and then apply the following suggestions. Also make sure to run rustfmt before you commit - that's why the CI failed!

kennethloeffler:
Do we need to do anything to rbx_xml for this? @Dekkonot

krakow10:
How about making lossy string conversion an optional decoder setting for now?  There are lots of maps on my game that simply don't work because of this, so I'm using this patched version.  I can agree that effectively destroying strings by default can be seen as a bug that should not be relied on.

kennethloeffler:
> How about making lossy string conversion an optional decoder setting for now? There are lots of maps on my game that simply don't work because of this, so I'm using this patched version. I can agree that effectively destroying strings by default can be seen as a bug that should not be relied on.

Meh, it's better than the behavior we have now (which is just to fall over and exit without producing a result). I think it should be fine until we properly support non UTF-8 `String` properties. Most of the time, these sorts of strings are present only because of "virus" instances, which no one wants or really cares about anyway.

Make sure to import `std::borrow::Cow` too, right now this branch will not compile

Dekkonot:
This needs an entry in [the changelog](https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_binary/CHANGELOG.md) but otherwise looks good to me.

---

> Do we need to do anything to rbx_xml for this?

Unicode in general doesn't currently work in rbx_xml so no, but only because we're in a bad spot anyway. @kennethloeffler

krakow10:
Squashed

krakow10:
Added changelog

krakow10:
Does Roblox use UTF-16?  I know there's a crate for that, maybe that's desirable instead of BStr.

Dekkonot:
No, Roblox uses UTF-8 encoding. It just also allows binary data to be put into strings. What we really need to do is not pretend that Roblox strings are always UTF-8. That's outside of scope for this PR though.

Dekkonot (approved):

Dekkonot:
@kennethloeffler When you have a moment, can you take a look at this again? Turns out it needs your approval, haha.

kennethloeffler (approved):

krakow10:
Rebased onto master to resolve merge conflicts.

Dekkonot (approved):

PR #393, opened 2024-02-20
Add Model.WorldPivotData custom getter/setter

### LINKED ISSUE #391: `Model.WorldPivotData` needs a custom property setter/getter
As discovered in #238, rbx_dom_lua is missing a custom setter/getter for `Model.WorldPivotData`, which causes failure to sync model pivots downstream in Rojo. 

We should add a custom setter and getter for this property that use `PVInstance:PivotTo` and `PVInstance:GetPivot`.

kennethloeffler:
This PR closes #391 by adding a custom getter/setter for `Model.WorldPivotData`.

I'm a little shaky about what ought to happen when the setter receives a `nil`. It likely means that the model's `PrimaryPart` was set to `nil` elsewhere... I think the safest option is probably to do nothing?


Dekkonot (approved):
I'm not sure about the PrimaryPart thing. `GetPivot` works just fine without it, so requiring a `PrimaryPart` for it feels odd!

On the other hand, I get it and think us being conservative here is probably fine. Roblox doesn't necessarily have a way to convery "optional cframe" from within their engine, so it working regardless may just be an artifact of that.

Going to hesitantly approve this because it works and I understand the reasoning. If it ends up being a mistake, it's not like it's a huge ordeal to correct.

kennethloeffler:
So, in my dazed confusion I did mess this up a little bit. 

`Model.WorldPivotData` is only unoccupied for a newly created `Model`. If I set the pivot whatsoever, the property remains occupied regardless of if the model has a PrimaryPart or not. With this in mind, the `if instance.PrimaryPart == nil then` bit is crap and should be removed. 

This also means that when we *want* the property to be unoccupied (i.e. when the setter receives a `nil`), we may have to create a new `Model` instance. I'm not sure if this level of accuracy is necessary, but we should keep it in mind

PR #394, opened 2024-02-20
Remove PrimaryPart conditional in WorldPivotData custom setter

kennethloeffler:
There was a mistake in #393 caused by some confusion around when `Model.WorldPivotData` is occupied and how it interacts `Model.PrimaryPart`. This PR removes a conditional that could cause incorrect behavior.


Dekkonot (approved):
I need more coffee because I noticed this in the other PR but talked myself into thinking it was fine.

Issue #383, opened 2024-01-24
Add support for `EnumItem`-typed attributes


kennethloeffler:
[Roblox added support for `EnumItem` attributes in 570](https://create.roblox.com/docs/release-notes/release-notes-570). rbx-dom currently lacks support for them, but they're probably uncommon since they're not yet available in Roblox Studio's attribute type picker UI.

Dekkonot:
It turns out that Enum attributes require us to store the name of the `Enum` that the `EnumItem` we're referring to is a part of. This makes sense, in hindsight, but it means that supporting these is going to be potentially complex since we have to start tagging enums with their names in some capacity.

My thoughts are that we modify the `Enum` type in `rbx_types` to also hold an `Option<String>` (or something akin to that) and then `rbx_binary` and `rbx_xml` (and other consumers) can fill that in if necessary. Then when serializing enums, we would raise an error if it didn't a name. It feels appropriately simple and doesn't require us to pull reflection info into anywhere new.

Dekkonot:
Okay I have bad news: Serde's derive macro is not powerful enough to accomplish what I was hoping to above. This means we'll have to implement a potentially complex serializer and deserializer for Enums if we use the existing struct.

My main concern is that Rojo users rely upon the serde format for enums to be stable, so the easy solutions are out the window: if we represent Enum as a struct instead of a newtype, it breaks the serde format.

It's also possible I have massively overcomplicated things so I'm interested to hear others thoughts on it.

kennethloeffler:
One way we could introduce this without breaking things or adding complexity is by introducing a new `Variant`, `QualifiedEnum`or something, that is used specifically for attributes. It might be a little weird since it's not a real Roblox data type, but we already have several specialized Variants anyway (`Tags`, `Attributes`, `MaterialColors`) so I'm gonna go ahead and try this out, wish me luck!

PR #396, opened 2024-02-23
Enum attributes

krakow10:
For #383

This is my guess as to how to implement this, assuming there is some sort of string in the serialized attributes. It builds but it does not work. Some sort of reflection database error that I don't know what it means.

```
thread 'main' panicked at /home/quat/git/rbx-dom/rbx_reflection_database/src/lib.rs:7:76:
could not decode reflection database because: invalid type: integer `0`, expected struct Enum
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


krakow10:
```Rust
println!("heyyy name={} value={}", enum_name, enum_value);
```
image
:sunglasses:

Dekkonot:
> It builds but it does not work. Some sort of reflection database error that I don't know what it means.

This is because you're changing the (de)serialization of Enum! As a result, the reflection database can't be deserialized by the new version you're making.

Fixing this is an involved processs because we bootstrap the database. As a result I don't really recommend you try to fix it because it's potentially error prone and involves a bit of juggling with reflection databases.

PS: We have the enum attributes documented in [the spec file we maintain](http://dom.rojo.space/attributes.html#enumitem). You didn't have to guess. 😄

krakow10:
What about a separate explicit enum type, like this?

krakow10:
This branch is now set up so that it suppresses the enum attribute errors and provides weakly typed enum functionality.  The explicit enum is moved to a different branch, never to be heard from again.

Dekkonot:
This implementation would fail a round-trip test because Enums no longer contain information on their type.

krakow10:
Well yes... but this is still useful as an error-suppressing patch while the full functionality isn't available

Dekkonot:
I will put aside time next week to fix this properly since it's showing up in the wild now, but I'm not willing to merge a feature that doesn't roundtrip. We won't be making a new release without the full feature anyway, so it being temporary fix is moot.

If it's desperate, I'd suggest using a custom fork of whatever tool you're using for the next little while.

krakow10:
I don't think it should be merged either which is why I have it as a draft, sorry for the misunderstanding. I'm not trying to get this incomplete feature merged, just a publicly available error fixer

PR #398, opened 2024-03-03
Delete generate_reflection

kennethloeffler:
generate_reflection was superseded by rbx_reflector in #249. rbx_reflector still has a few issues here and there, but we're now using it entirely in favor of generate_reflection. generate_reflection has not received any updates since, may not work properly anymore, increases compilation times to no purpose, and its presence could confuse new contributors, so it should be removed.


Dekkonot:
I am potentially worried about the test builds failing. Do they fail on `master` right now or is that a consequence of this deletion?

I am confused either way, but knowing if it's this changes' fault feels important.

kennethloeffler:
I'm investigating currently. I can't determine why CI is failing. I can't get it to happen locally, and I'm on the same version of Rust as CI (1.76.0) - but I'll keep you updated!

kennethloeffler:
I re-ran the build job on master and it fails in the same way: https://github.com/rojo-rbx/rbx-dom/actions/runs/7977827973/job/22222750855. So it's not related to this PR, at least.

Dekkonot:
All righty then. I suppose we can merge this, then. We should prioritize fixing the CI though.

PR #400, opened 2024-03-04
Add `InstanceBuilder::with_referent` constructing method

DervexHero:
This PR adds `with_referent` method for `InstanceBuilder` that allows to specify instance `Ref` manually. I also updated `.gitignore` for macOS users.


Dekkonot (approved):
Thank you!

PR #401, opened 2024-03-06
Make pretty JSON optional for `generate` subcommand (`rbx_reflector`)

DervexHero:
This PR makes writing of database from `generate` subcommand in pretty-printed JSON optional by adding `--pretty` option. I think that it makes sense as these files are not intended to be read by humans and it reduces the file size from `1.6MB` to `796KB`.

### DIFF
--- a/rbx_reflector/src/cli/generate.rs
+++ b/rbx_reflector/src/cli/generate.rs
@@ -34,7 +34,7 @@ pub struct GenerateSubcommand {
     pub output: Vec<PathBuf>,
     /// Whether to pretty-print the JSON output. This has no effect on MessagePack.
     #[clap(long)]
-    pub pretty: bool,
+    pub no_pretty: bool,
 }
 
 impl GenerateSubcommand {
--- a/rbx_reflector/src/cli/generate.rs
+++ b/rbx_reflector/src/cli/generate.rs
@@ -77,10 +77,10 @@ impl GenerateSubcommand {
 
             match extension {
                 Some("json") => {
-                    let result = if self.pretty {
-                        serde_json::to_writer_pretty(&mut file, &database)
-                    } else {
+                    let result = if self.no_pretty {
                         serde_json::to_writer(&mut file, &database)
+                    } else {
+                        serde_json::to_writer_pretty(&mut file, &database)
                     };
 
                     result.context("Could not serialize reflection database as JSON")?;
--- a/rbx_reflector/src/cli/generate.rs
+++ b/rbx_reflector/src/cli/generate.rs
@@ -27,11 +27,14 @@ use super::{defaults_place::DefaultsPlaceSubcommand, dump::DumpSubcommand};
 /// and write it to disk.
 #[derive(Debug, Parser)]
 pub struct GenerateSubcommand {
-    #[clap(long = "patches")]
+    #[clap(long)]
     pub patches: Option<PathBuf>,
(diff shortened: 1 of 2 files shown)


rbx_reflector/src/cli/generate.rs
```diff
@@ -27,11 +27,14 @@ use super::{defaults_place::DefaultsPlaceSubcommand, dump::DumpSubcommand};
...
 pub struct GenerateSubcommand {
-    #[clap(long = "patches")]
+    #[clap(long)]
     pub patches: Option<PathBuf>,
     /// Where to output the reflection database. The output format is inferred
     /// from the file path and supports JSON (.json) and MessagePack (.msgpack).
     pub output: Vec<PathBuf>,
+    /// Whether to pretty-print the JSON output. This has no effect on MessagePack.
+    #[clap(long)]
+    pub pretty: bool,
```
kennethloeffler, line 37:
```suggestion
    /// Whether to pretty-print the JSON output. This has no effect on MessagePack.
    #[clap(long)]
    pub no_pretty: bool,
```

rbx_reflector/src/cli/generate.rs
```diff
@@ -74,8 +77,13 @@ impl GenerateSubcommand {
...
                 Some("json") => {
-                    serde_json::to_writer_pretty(&mut file, &database)
-                        .context("Could not serialize reflection database as JSON")?;
+                    let result = if self.pretty {
+                        serde_json::to_writer_pretty(&mut file, &database)
+                    } else {
+                        serde_json::to_writer(&mut file, &database)
+                    };
+
+                    result.context("Could not serialize reflection database as JSON")?;
```
kennethloeffler, line 86:
```suggestion
                    let result = if self.no_pretty {
                        serde_json::to_writer(&mut file, &database)
                    } else {
                        serde_json::to_writer_pretty(&mut file, &database)
                    };

                    result.context("Could not serialize reflection database as JSON")?;
```

kennethloeffler (changes requested):
I'm fine with this option, and it may be useful once we start publishing databases to GitHub releases. 

I slightly disagree with "not intended to be read by humans." We do want to diff and read the database with our eyes quite frequently (whenever we open PRs to update it), and some editors don't perform well on extremely long lines! I'd prefer renaming this option to something like `no-pretty` and swapping the conditional, so that the default result is pretty printed, and the behavior of our existing `gen-reflection` script remains the same.

DervexHero:
> it may be useful once we start publishing databases to GitHub releases

I think that it is already pretty useful e.g. when including `rbx_dom_lua` in other projects.

> I slightly disagree with "not intended to be read by humans." We do want to diff and read the database with our eyes quite frequently (whenever we open PRs to update it)

Oh I thought that e.g. `rbx_dom_lua` uses 100% auto generated database from `rbx_reflector`, makes sense then.

> some editors don't perform well on extremely long lines

That's good to now!

kennethloeffler:
> Oh I thought that e.g. rbx_dom_lua uses 100% auto generated database from rbx_reflector, makes sense then.

We do use rbx_reflector to generate the database in rbx_dom_lua  -  I'd just like to be able to easily see how the database has changed after an update. Pretty printed JSON diffs pretty well, but it's much harder to see what changed when it's all slammed into a single line. It's certainly possible to prettify it manually after the fact... but I'd rather not add another step 😄

kennethloeffler (approved):

PR #402, opened 2024-03-14
Add method for getting the superclasses of an Instance to rbx_reflection

Dekkonot:
It's fairly common to need a list of superclasses for a class, so we should provide a method for it. This is an implementation.

It uses a `HashSet` for convenience. It's slightly more expensive than a `Vec` but provides an easy mechanism for checking if a class is in the returned set, so it feels worth it.

### DIFF
--- a/rbx_reflection/src/database.rs
+++ b/rbx_reflection/src/database.rs
@@ -43,17 +43,15 @@ impl<'a> ReflectionDatabase<'a> {
     /// Returns a list of superclasses for the provided class name. This list
     /// will start with the provided class and end with `Instance` if the class
     /// exists.
-    pub fn superclasses(&self, class_name: &str) -> Option<Vec<Cow<'a, str>>> {
-        // Parts have 4 superclasses, and they're generally what most models
-        // are composed of so we allocate enough for them.
-        // On average each class has 2.6 superclasses, so this benefits our
-        // theoretical 'average' case too.
+    pub fn superclasses(&self, class_name: &str) -> Option<Vec<&ClassDescriptor>> {
+        // As of the time of writing (14 March 2024), the class with the most
+        // superclasses has 6 of them.
         let mut list = Vec::with_capacity(6);
         let mut current_class = self.classes.get(class_name);
         current_class?;
 
         while let Some(class) = current_class {
-            list.push(class.name.clone());
+            list.push(class);
             current_class = class.superclass.as_ref().and_then(|s| self.classes.get(s));
         }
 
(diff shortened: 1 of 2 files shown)


kennethloeffler (commented):
> It uses a HashSet for convenience. It's slightly more expensive than a Vec but provides an easy mechanism for checking if a class is in the returned set, so it feels worth it.

Not sure if I agree - is `Vec::contains` any harder to use than `HashSet::contains`? And what's the maximum possible number of superclasses returned by this method?

Dekkonot:
I hadn't actually checked what the maximum amount was since it felt like it's either 1 or like 5 for most classes.

It turns out it is 6. With that in mind, I think a Vec is fine and I'm also going to update the starting capacity to just be 6.

rbx_reflection/src/database.rs
```diff
@@ -39,6 +39,26 @@ impl<'a> ReflectionDatabase<'a> {
             enums: HashMap::new(),
         }
     }
+
+    /// Returns a list of superclasses for the provided class name. This list
+    /// will start with the provided class and end with `Instance` if the class
+    /// exists.
+    pub fn superclasses(&self, class_name: &str) -> Option<Vec<Cow<'a, str>>> {
```
kennethloeffler, line 46: This method might be more convenient to use if returns `Vec<&ClassDescriptor>` - that way, consumers can get more information than just the name without having to look up the descriptor themselves, and won't have to destructure if all they want to do is e.g. iterate over superclasses

kennethloeffler (changes requested):

rbx_reflection/src/database.rs
Dekkonot: That... is a very fair point. I have no objections to changing it.

rbx_reflection/src/database.rs
```diff
@@ -39,6 +39,24 @@ impl<'a> ReflectionDatabase<'a> {
             enums: HashMap::new(),
         }
     }
+
+    /// Returns a list of superclasses for the provided class name. This list
+    /// will start with the provided class and end with `Instance` if the class
+    /// exists.
+    pub fn superclasses(&self, class_name: &str) -> Option<Vec<&ClassDescriptor>> {
```
kennethloeffler, line 46:
```suggestion
    pub fn superclasses(&self, class_name: &str) -> Vec<&ClassDescriptor> {
```

Did you still want this return to be optional?

Dekkonot: The other option would be returning an empty vector in case of an invalid class name, which turns `value.is_some()` into `value.is_empty()`.

I think it's fine for it to be optional as a result.

kennethloeffler (approved):

PR #403, opened 2024-03-28
Harden rbx_reflector against unknown data types

kennethloeffler:
Today I noticed that `rbx_reflector generate` failed to output a database due to a new datatype: 
```
     Running `target/debug/rbx_reflector generate rbx_reflection_database/database.msgpack rbx_dom_lua/src/database.json --patches patches`
[INFO  rbx_reflector::cli::defaults_place] Installing Studio plugin
[INFO  rbx_reflector::cli::defaults_place] Starting Roblox Studio...
[INFO  rbx_reflector::cli::defaults_place] Waiting for Roblox Studio to re-save place...
[INFO  rbx_reflector::cli::defaults_place] Uninstalling Studio plugin
Please save the opened place in Roblox Studio (ctrl+s).
[INFO  rbx_reflector::cli::defaults_place] Place saved, killing Studio...
Error: Unknown type Path2DControlPoint
```

This got me thinking about how rbx_reflector can more gracefully deal with unknown data types. I intuit that we only need to care about unknown data types if they serialize. With this in mind, this PR makes `rbx_reflector generate` only bail on unknown serializable data types (with the exception of `QDir` and `QFont` - see comments).

### DIFF
--- a/rbx_reflection/src/class_tag.rs
+++ b/rbx_reflection/src/class_tag.rs
@@ -18,6 +18,7 @@ pub enum ClassTag {
 }
 
 #[derive(Debug)]
+#[allow(dead_code)]
 pub struct ClassTagFromStrError(String);
 
 impl FromStr for ClassTag {
--- a/rbx_reflection/src/property_tag.rs
+++ b/rbx_reflection/src/property_tag.rs
@@ -17,6 +17,7 @@ pub enum PropertyTag {
 }
 
 #[derive(Debug)]
+#[allow(dead_code)]
 pub struct PropertyTagFromStrError(String);
 
 impl FromStr for PropertyTag {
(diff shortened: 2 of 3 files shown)


rbx_reflection/src/class_tag.rs
```diff
@@ -18,6 +18,7 @@ pub enum ClassTag {
 }
 
 #[derive(Debug)]
+#[allow(dead_code)]
```
Dekkonot, line 21: Why this?

kennethloeffler: In Rust 1.77.0, the `unused_tuple_struct_fields` lint was merged into the `dead_code` lint: https://github.com/rust-lang/rust/pull/118297/. This caused our lint check to fail, but I think we should keep the field.

Dekkonot (approved):

rbx_reflection/src/class_tag.rs
Dekkonot: Fair enough. I guess that makes sense.

PR #404, opened 2024-04-08
Correct Type ID of ColorSequence in the Attributes documentation

1enrique3:
In the documentation the type ID of ColorSequence is set to `0x17` instead of `0x19`, which is already taken by the NumberSequence type.

https://github.com/rojo-rbx/rbx-dom/blob/c8ec8a9ac79c957c1628f93888fc6c5b0af119c5/rbx_types/src/attributes/type_id.rs#L48


Dekkonot (approved):

Dekkonot:
Thank you!

PR #405, opened 2024-04-08
Add method for getting UniqueId for referent

Dekkonot:
Adds a method for getting the UniqueId for a given referent in a WeakDom.

Technically unnecessary but a convenience function is helpful and there's nothing stopping us from linking referents more directly to their UniqueId in the future, which this method would benefit from.

### DIFF
--- a/rbx_dom_weak/CHANGELOG.md
+++ b/rbx_dom_weak/CHANGELOG.md
@@ -1,10 +1,16 @@
 # rbx_dom_weak Changelog
 
 ## Unreleased Changes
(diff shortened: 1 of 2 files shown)


kennethloeffler:
I think this is a reasonable addition. It provides an easy way to go from `Ref` to `UniqueId`, but it does make me consider that it's straight up impossible to go the other way around: `UniqueId` to `Instance` or `Ref`, assuming the consumer only has a `UniqueId`. Maybe we should have an additional method, something like

```rust
pub fn get_by_unique_id(unique_id: UniqueId) -> Option<&Instance>
```

PR #407, opened 2024-04-19
Improve `rbx_reflector` Roblox Studio automation and logging on macOS

DervexDev:
This PR brings automatic place saving to macOS, so it now works exactly the same as on Windows. It also pipes the `stdout` and `stderr` streams to `null` to get rid of extreme verbosity on macOS.

https://github.com/rojo-rbx/rbx-dom/assets/78505208/c488f168-dea0-4a3f-8f26-b83c8b5b4a66
> _New automation with old verbose output_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Cargo stuff
-target
+/target
 **/*.rs.bk
 Cargo.lock
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Cargo stuff
-/target
+target
 **/*.rs.bk
 Cargo.lock
 
--- a/rbx_reflector/plugin.lua
+++ b/rbx_reflector/plugin.lua
@@ -1,3 +1,7 @@
+if game.Name ~= "defaults-place.rbxlx" then
+	return
+end
+
 local HttpService = game:GetService("HttpService")
 
 local SERVER_URL = "http://localhost:22073"
--- a/rbx_reflector/src/cli/defaults_place.rs
+++ b/rbx_reflector/src/cli/defaults_place.rs
@@ -92,7 +92,18 @@ fn save_place_in_studio(path: &PathBuf) -> anyhow::Result<StudioInfo> {
         Command::new("osascript")
 				.args([
 					"-e",
-					"tell application \"System Events\" to tell process \"RobloxStudio\" to set frontmost to true",
+					"tell application \"System Events\"
+                        repeat with theProcess in processes whose name is \"RobloxStudio\"
+                                tell theProcess
+                                    set windowList to windows whose name contains \"defaults-place.rbxlx - Roblox Studio\"
+                                    
+                                    if (count of windowList) > 0 then
+                                        set frontmost to true
+                                        perform action \"AXRaise\" of window 1
+                                    end if
+                                end tell
+                        end repeat
+                    end tell",
 				])
 				.output()?;
 
--- a/rbx_reflector/src/cli/defaults_place.rs
+++ b/rbx_reflector/src/cli/defaults_place.rs
@@ -104,8 +104,7 @@ end tell
             .output()?;
     }
 
-    #[cfg(not(target_os = "windows"))]
-    #[cfg(not(target_os = "macos"))]
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     println!("Please save the opened place in Roblox Studio (ctrl+s).");
 
     loop {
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 # Cargo stuff
-/target
+target
```
kennethloeffler, line 2: We're not only writing Rust in this repository, so we really do want to ignore just the top-level `target` directory here (it's possible we could introduce some other `target` directory elsewhere that we do want in VC, and then we'll just have to remove this again). Also, we typically prefer to keep PRs focused on one thing, without sprinkling in other miscellaneous changes when we don't have to.

rbx_reflector/src/cli/defaults_place.rs
```diff
@@ -63,6 +63,8 @@ fn save_place_in_studio(path: &PathBuf) -> anyhow::Result<StudioInfo> {
     log::info!("Starting Roblox Studio...");
 
     let mut studio_process = Command::new(studio_install.application_path())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
```
kennethloeffler, line 67: I've never seen Roblox Studio emit any useful logs or errors, so I think this is a good change. I think it's possible we may eventually want to see what Roblox Studio outputs, but we can decide what to do about that another day.

rbx_reflector/src/cli/defaults_place.rs
```diff
@@ -85,7 +87,36 @@ fn save_place_in_studio(path: &PathBuf) -> anyhow::Result<StudioInfo> {
...
+                    end tell",
+				])
+				.output()?;
+
+        Command::new("osascript")
+            .args([
+                "-e",
+                "tell app \"System Events\" to key code 1 using command down",
+            ])
+            .output()?;
```
kennethloeffler, line 115: I went ahead and took the liberty of rewriting this section to be simpler, more robust and (I think) more readable by:
* using one osascript invocation, and doing the whole script in one tell block
* using Rust's raw string literals to avoid escaping and make the formatting a little nicer
* bringing Roblox Studio to front by process id, rather than by enumerating all Roblox Studio processes and matching on window name
  I'm not sure if the `perform action "AXRaise" of window 1` statement is necessary - do you have any more info about this?
* using `keystroke "s"` instead of `key code 1`

Let me know if it works for you:
```suggestion
        let process_id = studio_process.id();
        let script = format!(
            r#"
tell application "System Events"
    set frontmost of the first process whose unix id is {process_id} to true
    keystroke "s" using command down
end tell
"#
        );

        Command::new("osascript")
            .args(["-e", script.as_str()])
            .output()?;
```

rbx_reflector/src/cli/defaults_place.rs
```diff
@@ -85,7 +87,36 @@ fn save_place_in_studio(path: &PathBuf) -> anyhow::Result<StudioInfo> {
...
+    }
+
     #[cfg(not(target_os = "windows"))]
+    #[cfg(not(target_os = "macos"))]
```
kennethloeffler, line 119: We can use the `any` predicate to express this in one line:
```suggestion
    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
```

kennethloeffler (changes requested):
Thanks for this. A long time ago I had the idea to use AppleScript to automatically save the defaults place, but I never got around to it. It's a bit of a weird language, and I'm glad I got the opportunity to dive into some of it here! 😄 

A few comments:

rbx_reflector/src/cli/defaults_place.rs
DervexDev: > * I'm not sure if the `perform action "AXRaise" of window 1` statement is necessary - do you have any more info about this?

The `AXRaise` action "un-minimizes" the window, which after testing is actually unnecessary because even if the user minimizes Roblox Studio `set frontmost` will focus the window "enough" to send our keystroke. The window will simply remain minimized until it's eventually closed by `rbx_reflector`.

DervexDev:
I also updated the `plugin.lua` to ensure the correct Roblox Studio instance is connecting to the `rbx_reflector`

rbx_reflector/plugin.lua
```diff
@@ -1,3 +1,7 @@
+if game.Name ~= "defaults-place.rbxlx" then
+	return
+end
+
```
kennethloeffler, line 4: Nice, this will help when Roblox Studio's "Reload plugins on file changed` setting is enabled

kennethloeffler (approved):

PR #410, opened 2024-05-05
Add option to serialize human-readable MessagePack in `rbx_reflector`

DervexDev:
This PR enables ambiguous database reflection deserialization (using [rmpv](https://docs.rs/rmpv/latest/rmpv/) Value) just like it is possible with JSON.

I'm not sure whether the `--human-readable` option name actually fits this usecase as it's just based on the serde serialzer's `is_human_readable` function.

### DIFF
--- a/rbx_reflector/src/cli/generate.rs
+++ b/rbx_reflector/src/cli/generate.rs
@@ -13,6 +13,8 @@ use rbx_reflection::{
     PropertySerialization, PropertyTag, ReflectionDatabase, Scriptability,
 };
 use rbx_types::VariantType;
+use rmp_serde::Serializer;
+use serde::Serialize;
 use tempfile::tempdir;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


Dekkonot:
Just so I'm not missing anything, what's the motivation for this change? Why would we want a human-readable messagepack database?

DervexDev:
So my tool [Argon](https://github.com/argon-rbx) uses `rbx_dom_lua` as a Roblox Studio plugin backend. I migrated from JSON because of the lack of support for stuff like `math.huge` and as you probably know it is the default value of some instance properties. To build the plugin with the MessagePack database I added support for `.msgpack` files to the Argon CLI. Now I want it to be available for Argon users as well so obviously I have to deserialize everything ambiguously.

I know you want to move Rojo to MessagePack as well, so it might be useful here as well.

Dekkonot (approved):
Sounds good, thanks for the contribution.

PR #414, opened 2024-06-03
Implement Serde for SharedString

Dekkonot:
Implements Serde for SharedString. The implementation is pulled from `BinaryString` to avoid us having to do anything fancy.

This will allow users to specify SharedString properties in Rojo, which is of some importance.


kennethloeffler (approved):
LGTM, but it'd be nice to have some elaboration on how this will be integrated into Rojo

PR #417, opened 2024-06-06
Make `rbx_reflector` library

DervexDev:
I am working on an automated database reflection generation tool and it would be nice to use `rbx_reflector` in its backend. 

This PR makes `rbx_reflector` a separate crate and exposes all CLI subcommands as simple functions. I also removed `publish = false` because we can now publish it to crates.io.


Dekkonot (approved):
I'm hesitant to do this because it places a maintenance burden on us; we suddenly have to respect semver, keep a changelog, and maintain docs for rbx_reflector.

I am okay with it, since it's something we should probably do anyway but I'm not the only rbx-dom maintainer. @kennethloeffler any objections?

Issue #418, opened 2024-06-06
Defining the Archivable property on an instance causes binary serializer to write false for Archivable properties on all other instances of the same class


kennethloeffler:
Minimal repro:
```rust
use rbx_dom_weak::{InstanceBuilder, WeakDom};

fn main() {
    let dom = WeakDom::new(InstanceBuilder::new("Folder").with_children([
        InstanceBuilder::new("Folder").with_property("Archivable", true),
        InstanceBuilder::new("Folder"),
    ]));

    let file = std::fs::File::create("weird-archivable.rbxm").unwrap();
    rbx_binary::to_writer(file, &dom, &[dom.root_ref()]).unwrap()
}
```
Output of `rbx-util view-binary weird-archivable.rbxm`:
```
---
num_types: 1
num_instances: 3
chunks:
  - Inst:
      type_id: 0
      type_name: Folder
      object_format: 0
      referents:
        - 0
        - 1
        - 2
  - Prop:
      type_id: 0
      prop_name: archivable
      prop_type: Bool
      values:
        - false
        - true
        - false
  - Prop:
      type_id: 0
      prop_name: Name
      prop_type: String
      values:
        - Folder
        - Folder
        - Folder
  - Prnt:
      version: 0
      links:
        - - 2
          - 0
        - - 1
          - 0
        - - 0
          - -1
  - End
```
There are several things aligning that cause this and make it problematic:
* The reflection database is missing default values for `Instance.Archivable`. This is because Roblox Studio never writes this property, meaning rbx-reflector cannot discover its default value.
* Because the reflection database is missing default values for `Instance.Archivable`, the binary serializer writes the boolean fallback default (which is `false`) for instances that are missing the Archivable property.
* rbx-dom serializes `Instance.Archivable` as`Instance.archivable`.
* Roblox Studio does *not* load `Instance.Archivable`, but *does* load `Instance.archivable`. This means all the instances with `Instance.archivable == false` will in fact be loaded that way in Roblox Studio, and thus will not be saved when publishing to Roblox or when playtesting.

I thought of a few different ways we could solve this:
[...]

PR #420, opened 2024-06-06
Traverse inheritance tree to find defaults in rbx_binary serializer

kennethloeffler:
To satisfy some requirements for #418, this PR changes rbx_binary's serializer to traverse up the inheritance tree of the reflection database when locating default values to write for instances that have missing properties.

`find_default_property` could be a method on `ReflectionDatabase`, but since it is not clear if this operation will have any utility outside of rbx_binary's serializer, it is kept internal.

This adds a small amount of overhead when locating properties that have defaults located further up the inheritance tree, and for instances that have properties unknown to the reflection database. The overhead is likely small, but it is probably worth benchmarking to ensure it is not too great.


Dekkonot:
I think putting this method on a reflection database is probably worthwhile. A use case I imagine is lazily evaluating the default value of a property for a tool like Lune.

kennethloeffler:
I'm noticing some inconsistency between the parameters of `ReflectionDatabase::superclasses` and `ReflectionDatabase::find_default_property`. `find_default_property` accepts a class descriptor, but `superclasses` takes a class name. It might be better if these were the same. I'd rather not have `find_default_property` accept a string class name because looking up an initial `ClassDescriptor` is redundant for rbx_binary's usage, so should we change `superclasses` so it accepts a `ClassDescriptor` instead of a `&str`?

Dekkonot:
Sorry for the delay in getting back to you on this. Yeah, I suppose we should probably adjust `superclasses` to use a ClassDescriptor. It's one less lookup that it has to do and doesn't really change the use of the method. I'll handle that in a separate PR.

Dekkonot (commented):
Needs the PR number in the changelog but otherwise looks good to me.

Dekkonot (approved):

PR #432, opened 2024-08-05
Populate relevant_instances with depth-first post-order traversal 

kennethloeffler:
(in response to a critical bug downstream at https://github.com/rojo-rbx/rojo/issues/953)

This PR changes the way rbx_binary populates `SerializerState.relevant_instances` from a reversed breadth-first traversal (#411) to a depth-first post-order traversal. This is important for ensuring that the PRNT chunk has the correct order. I attempted to discover why the PRNT chunk ordering matters for correctness, but this will take a while to do in a black-box way, and I don't think we have time for it. 

With this change, while the PRNT now has the correct ordering, our overall implementation still differs slightly from Roblox's. I documented the deviations in a comment just before performing the traversal.


Dekkonot:
Do we want to document this in the binary spec file since it's important? If so, do you wanna do that in this PR or in a later one?

Dekkonot (approved):

kennethloeffler:
Let's do it later, I'd like to find out more about what's going on before committing to documenting it

PR #435, opened 2024-08-15
In rbx_reflector, support properties seralizing but not setting a default and set UniqueId properties to do this

Dekkonot:
One of the issues with keeping `UniqueId` as serializable in our database was the noise it caused by having a default value saved that was random every time. This PR adds a means to indicate that a property should not have its default value stored even if it serializes.

There's some hoop jumping in this PR because of how patch application and the borrow checker work, but I think it's fine. Let me know if you have questions.

I've also marked `Instance.UniqueId` and `Instance.HistoryId` as serializable again, just without their default stored. Technically this could also apply to `LuaSourceContainer.ScriptGuid` but seeing as that's a string property, I don't know how `rbx_binary` would deal with generating a default for it so I've left it alone for now.


kennethloeffler:
Maybe we should solve this using default overrides, like we discussed in https://github.com/rojo-rbx/rbx-dom/issues/418? It should be as simple as adding a default value field (probably an Option<Variant>) to rbx-reflector's PropertyChange, then adding them to the class' defaults during patch application, and also ensure that we don't later overwrite any existing defaults when we apply the default place

We could specify null UniqueIds and null GUIDs for these properties (ScriptGuid behaves similarly to UniqueId props when set to a null value [or even just an empty string] - it gets overwritten with a new id when loaded into studio), and then we would also be able to solve the problem in the linked issue by specifying a default for Archivable

Dekkonot:
I'll admit, I forgot we had that discussion. It does make more sense to do though. Do you reckon it's worth fixing this PR or should I close it and open a new one.

kennethloeffler:
Doesn't make too much of a difference, but a new PR would contain less noise so might be easier to read if we ever need to look back at it

PR #437, opened 2024-08-16
Serialize UniqueId properties + support setting default property value in patches

Dekkonot:
As discussed in #435 and previously in #418, this allows us to specify the default value for properties using patches in rbx-reflector.

In addition to that, this PR also sets a default for `Archivable` and also reverts #327 and overrides their defaults instead. Since several of these properties are on `Instance`, this results in a very large change to the reflection database.

We can and should in the future make it so defaults are stored in a method that respects inheritance and encourage people to use `ReflectionDatabase::find_default_property` to find defaults. The reason I'm not doing it at this moment is that historically people haven't had to care about inheritance to get defaults. Though it's not a breaking change, I feel like we should give some warning if we change it so they have to, especially for critical things like `Archivable`.


patches/instance.yml
```diff
@@ -1,22 +1,24 @@
...
     UniqueId:
-      Serialization:
-        Type: DoesNotSerialize
+      DefaultValue:
+        UniqueId: "00000000000000000000000000000000"
     HistoryId:
-      Serialization:
-        Type: DoesNotSerialize
+      DefaultValue:
+        UniqueId: "00000000000000000000000000000000"
```
kennethloeffler, line 11: Might be a good idea to leave a comment explaining why these values were chosen (likewise for ScriptGuid)

rbx_reflector/src/patches.rs
```diff
@@ -90,6 +91,65 @@ impl Patches {
...
+                    .unwrap()
+                    .properties
+                    .get(prop_name.as_str());
+                if let Some(prop_data) = prop_data {
+                    match (&prop_data.data_type, default_value.ty()) {
+                        (DataType::Enum(_), VariantType::Enum) => {},
+                        (DataType::Value(existing), new) if *existing == new => {},
+                        (existing, new) => bail!(
+                            "Property {class_name}.{prop_name}'s value type is being changed by a DefaultValue patch.\n\
+                            Existing: {existing:?}, new: {new:?}")
```
kennethloeffler, line 130: This is a bit unclear to me, I'd prefer something like
```suggestion
                        (expected, actual) => bail!(
                            "Bad type given for {class_name}.{prop_name}'s DefaultValue patch.\n\
                            Expected {expected:?}, got {actual:?}")
```

kennethloeffler (commented):
Well, I'm not jazzed about how complex the code to propagate default values to all subclasses ended up (nor the additional space it uses), but I agree with the reasoning for why. Just a couple comments:

Dekkonot:
Yeah I'm not super jazzed either, but it's what we gotta do for now because the other option is committing to a broader change. We should probably prioritize it, but I don't think it's worth crying over right now.

kennethloeffler:
Last thing: I've edited the PR's title to mention that it's making UniqueId props serialize again, since we've had problems with UniqueId in the past. If this title isn't quite to your liking feel free to change it

kennethloeffler (approved):

PR #441, opened 2024-08-22
Use correct comparison for isAttributeNameReserved

kennethloeffler:
In #440 I copy pasted some code and forgot to change the comparison operator! 🫠


Dekkonot (approved):
You know that's on me, I should've caught that. Whoops.

PR #444, opened 2024-08-29
Initial support for Terrain SmoothGrid data

checkraisefold:
**DISCLAIMER: THIS IS MY FIRST ACTUAL PROJECT IN RUST. Terrible code will bleach your eyes!**
Because of that, this PR also has edits by maintainers on.

This PR adds support for serialization and deserialization to `rbx-types` for the `Terrain` object `SmoothGrid` property data. 
This fully permits the creation of external tooling for modification and generation of smooth terrain data, as Roblox Studio, along with RCCService (presumably) generate the additionally-required PhysicsGrid data at runtime/publish-time.

As a bonus, no reverse engineering of the client had to be done here; just a lot of comparing outputs with ImHex peddled by this article.
https://zeux.io/2017/03/27/voxel-terrain-storage/

[Rendered spec.](https://github.com/checkraisefold/rbx-dom/blob/terrain/docs/smooth-grid.md)

- [x] Decoder
- [x] Decoder tests
- [x] Encoder
- [x] Encoder tests
- [x] Module splitup


docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+| Header     | `[u8; 2]`       | Magic number and chunk size.                                          |
+| Chunks     | [Chunk](#chunk) | As many chunks as needed to fill the blob. Unknown/no maximum amount. |
+
+## Data Types
+
+Terrain data is represented using a variety of different data types. Data described within them follows any conventions set in the [document conventions](#document-conventions).
+
+### Chunk
+
+The `Chunk` type is stored with a dynamic size dependent on the size of the voxels contained within, and the end of its data is marked by reaching its maximum voxel count of 32<sup>3</sup>. Voxels are ascendingly ordered by Y, then Z, then X based on their position in the chunk. Voxels are stored in rows of 32 units on each axis, each representing 4 units in world space.
```
Dekkonot, line 43: Is the maximum size fixed at 32<sup>3</sup> or is it dynamic based on the size specified in the header? It's unclear here because it says it's dynamically sized but the max value is set in stone.

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+  - [Material](#material)
+
+## Document Conventions
+
+This document assumes a basic understanding of Rust's convention for numeric types. For example:
+
+- `u32` is an unsigned 32-bit integer
+- `f32` is a 32-bit floating point number
+
+All numeric types are little endian. Floats are stored as a dividend of their type's numeric maximum; as such, a `f8` with a value of `0x31` translates to `0x31 / 0xFF`, or `~0.192`. Floats are unsigned. Signed integers are stored with their absolute value, with preceding bytes containing the signedness information; meaning the range for an `i8` is `-0xFF` -> `0xFF`, instead of `-0x7F` to `0x7F`.
```
Dekkonot, line 22: The conventions here are a bit confusing. To me, the name `i8` suggests the number is 1 byte, but if the sign is stored in the previous byte, that seems to be 2 bytes. Can you give an example of what that looks like in practice?

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Data Types
+
+Terrain data is represented using a variety of different data types. Data described within them follows any conventions set in the [document conventions](#document-conventions).
+
+### Chunk
+
+The `Chunk` type is stored with a dynamic size dependent on the size of the voxels contained within, and the end of its data is marked by reaching its maximum voxel count of 32<sup>3</sup>. Voxels are ascendingly ordered by Y, then Z, then X based on their position in the chunk. Voxels are stored in rows of 32 units on each axis, each representing 4 units in world space.
+
+Voxel data is preceded by the offset between this chunk and the last in the blob, or from `0, 0, 0` if this is the first chunk in the blob. This offset is stored using 3 vectors of `(x: u8, y: u8, z: u8)` using the signedness determination described in [document conventions](#document-conventions), with 0xFF values in _all_ unused offsets indicating a negative sign. All coordinates are in chunk space (increments of 1 chunk), not world space.
```
Dekkonot, line 45: Reading between the lines, I'm able to work out that the offset is essentially 3 arrays that get added together like `(offset_2_16 * 2^16) + (offset_2_8 * 2^8) + offset`, but it's not clear whether this is correct and it took me thinking about the struct described before to come to this conclusion.

It would be good if it were explicitly spelled out how the offset was stored and how it was converted into that format.

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+
+For example, given two chunks, one at a position of `(2, 0, 0)` and another at a position of `(4, 0, -1)` in chunk space, the latter's offset would be stored as follows:
+
+1. (0x00, 0x00, 0xFF)
+2. (0x00, 0x00, 0xFF)
+3. (0x00, 0x00, 0xFF)
+4. (0x02, 0x00, 0x01)
```
Dekkonot, line 52: It would be cool if there was an explanation of what each of these vectors represented

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+| Offset (65536) | `[i8; 3]`       | Offset from the last chunk in the blob, each axis being multiplied by 65536 chunks. |
+| Offset (256)   | `[i8; 3]`       | Offset from the last chunk in the blob, each axis being multiplied by 256 chunks.   |
+| Offset (1)     | `[i8; 3]`       | Offset from the last chunk in the blob, each axis being in singular chunks.         |
+| Voxels         | [Voxel](#voxel) | As many voxels as needed to fulfill the maximum amount per chunk.                   |
+
+### Voxel
+
+The `Voxel` type is stored with a dynamic size (between 1-4 bytes) dependent on its set [bitflags](#voxel.flag). Only stored within [chunks](#chunk). A reference for the data contained within voxels [can be found here][Terrain.WriteVoxelChannels]. Empty voxels are stored with their material set to Air.
+
+Voxels are stored within a chunk using [run-length encoding]. A count of `1` would indicate the voxel repeating itself once, meaning 2 of the same voxel in a row would be read when decoded. Occupancy values are stored using the float format described in the [document conventions](#document-conventions).
```
Dekkonot, line 66: When you say "run length encoding" do you mean that the size of the `Count` field increases dynamically or is it just a number between 0 and 255? If it's the latter, this is a misuse of the term and it's confusing.

If it's the former, this needs to be rephrased because it suggests to me that the value is just a simple "this voxel repeats between 0 and 255 times".

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+
+Water occupancy was added as an addendum to the existing format for the new [Shorelines] feature release. To avoid restricting the [Voxel.Flag](#voxel.flag)'s `Material Index` field to 5 bits, or rearchitecting the format, run-length encoding is disabled for the current voxel when it has a water occupancy of above `0.0`. Instead, the `Store Count` bit is enabled, and a count of `0` is written. Following the empty count, the water occupancy byte is stored. If a voxel has a solid occupancy of `1.0` (unless the material is Air), the water occupancy should always be `0.0`.
```
Dekkonot, line 68: This feels like it could be simplified somewhat. I would like to provide a suggestion, but it's unclear what it means to me. Under what conditions is water occupancy stored? My best guess is it's when `Store Count` is enabled and the count is written as `0`. If this is accurate, I'd write it like this:

```suggestion
Water occupancy is dynamically set based on the `Store Count` flag and the stored `Count` value. If `Store Count` is enabled and `Count` is `0x00`, then the Water Occupancy is written.

If the occupancy of a Voxel is `1.0` and the set material is not `Air`, the water occupancy must always be `0.0` if it's set.
```

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+
+| Field Name      | Format                    | Value                                                                                                          |
+| :-------------- | :------------------------ | :------------------------------------------------------------------------------------------------------------- |
+| Flag            | [Voxel.Flag](#voxel.flag) | Contains the material of this voxel, along with other bitflags.                                                |
+| Solid Occupancy | `f8`                      | Occupancy between 0-100% of the set material for this voxel. Only stored if the `Store Occupancy` flag is set. |
+| Count           | `u8`                      | Run-length count. Only stored if the `Store Count` flag is set.                                                |
+| Water Occupancy | `f8`                      | Occupancy between 0-100% of Water for this voxel. Only stored based on the conditions described above.         |
```
Dekkonot, line 75: I'd love an explanation of occupancy in the context of Terrain voxels, even if it's just "Occupancy represents roughly how full this voxel is as a percentage". It's better to overexplain than underexplain.

docs/terrain.md
```diff
@@ -0,0 +1,123 @@
...
+## File Structure
+
+The first two bytes of the blob are `0x01`, which is a magic number, followed by `0x05`, which is a logarithm base 2 of the chunk size in voxels. Currently, the chunk size is 32<sup>3</sup> (32768). Values other than `0x05` are untested.
+
+Immediately following the header is an array of chunks, each of which must contain enough voxels to reach the maximum count (32<sup>3</sup>). Chunks are ascendingly ordered by X, then Y, then Z based on their position in the world. Each chunk represents a cube of 128<sup>3</sup> units in world space.
+
+| Field Name | Format          | Value                                                                 |
+| :--------- | :-------------- | :-------------------------------------------------------------------- |
+| Header     | `[u8; 2]`       | Magic number and chunk size.                                          |
+| Chunks     | [Chunk](#chunk) | As many chunks as needed to fill the blob. Unknown/no maximum amount. |
```
Dekkonot, line 35: I'd prefer the `Header` section be split into two segments:
```suggestion
| Field Name   | Format          | Value                                                                 |
| :----------- | :-------------- | :-------------------------------------------------------------------- |
| Magic Number | `u8`            | The magic number `0x01`                                               |
| Chunk Size   | `u8`            | The size of the chunks logarithm base 2. Should be `0x05`.            |
| Chunks       | [Chunk](#chunk) | As many chunks as needed to fill the blob. Unknown/no maximum amount. |
```

Additionally, how is the `Chunks` field sized? Is it just "until the end of the blob" and you can add as many or as few chunks as you want? This should be written down.

Dekkonot (changes requested):
I'd prefer we keep the naming scheme intact and name the spec file for `SmoothGrid` to be `smooth-grid.md`.

Otherwise, I've left my comments and questions for the spec file.

docs/terrain.md
checkraisefold: I'll test it real quick- I have no idea if the engine will overwrite the size in the header, if the game will crash, or what it will do.
32<sup>3</sup> is the default size, and can't be changed by normal means in Studio or any other developer tools

docs/terrain.md
checkraisefold:
```
When you say "run length encoding" do you mean that the size of the Count field increases dynamically or is it just a number between 0 and 255? If it's the latter, this is a misuse of the term and it's confusing.
```
This is like the point of run length encoding though. The count is a number between 0 and 255 determining how long a run length is, AKA, on decode, the voxel repeats itself `count` times and should be read as such, while only taking 2-3 bytes of actual storage space.
```
Run-length encoding (RLE) is a form of [lossless data compression](https://en.wikipedia.org/wiki/Lossless_data_compression) in which runs of data (consecutive occurrences of the same data value) are stored as a single occurrence of that data value and a count of its consecutive occurrences, rather than as the original run.
```

docs/terrain.md
checkraisefold: Clarifying that changing from 32^3 / the value of 5 can not be loaded by the engine in new commit

checkraisefold: I lied it works above 5, when re-encoding/decoding it normalizes it back to 5/chunk size of 32 though it just splits up the big chunks

docs/smooth-grid.md
```diff
@@ -0,0 +1,137 @@
...
+
+All numeric types are little endian. Floats are stored as a dividend of their type's numeric maximum; as such, a `f8` with a value of `0x31` translates to `0x31 / 0xFF`, or `~0.192`. Floats are unsigned.
+
+Unless otherwise noted, all structs in this document are assumed to be stored with their components in the sequence listed without any modification. That is, if a struct is listed as being composed of an `i32` and an `f32`, it can be assumed that it is stored as an `i32` followed by an `f32`.
+
+## File Structure
+
+The first two bytes of the blob are `0x01`, which is a magic number, followed by `0x05`, which is a logarithm base 2 of the chunk size in voxels. The default chunk size is 32<sup>3</sup> (32768).
+
+Size values other than `0x05` are normally not achievable with Studio or other tools, but values between `0x00` (inclusive) and `0x08` (inclusive) can still be deserialized by the engine, giving a theoretical range of possible chunk sizes between 1<sup>3</sup> and 256<sup>3</sup>. However, the engine splits chunks differeing in size from 32<sup>3</sup> back into the default chunk size. Any size value other than `0x05` has poorly tested and undefined behavior.
```
Dekkonot, line 30:
```suggestion
Size values other than `0x05` are normally not achievable with Studio or other tools, but values between `0x00` (inclusive) and `0x08` (inclusive) can still be deserialized by the engine, giving a theoretical range of possible chunk sizes between 1<sup>3</sup> and 256<sup>3</sup>. However, the engine splits chunks differing in size from 32<sup>3</sup> back into the default chunk size. Any size value other than `0x05` has poorly tested and undefined behavior.
```

docs/smooth-grid.md
```diff
@@ -0,0 +1,137 @@
...
+
+Immediately following the header is an array of chunks, each of which must contain enough voxels to reach the maximum count (32<sup>3</sup>). Chunks are ascendingly ordered by X, then Y, then Z based on their position in the world. Each chunk represents a cube of 128<sup>3</sup> units in world space.
```
Dekkonot, line 32: I think we went over this on Discord, but chunk sizes aren't fixed and are in fact `(2^CHUNK_SIZE)^3`, correct? If this is the case then this should be adjusted to indicate that somehow.

docs/smooth-grid.md
```diff
@@ -0,0 +1,137 @@
...
+| Chunk Size   | `u8`                 | Logarithm base 2 of the chunk size. Should be `0x05`.                                |
+| Chunks       | [Vec<Chunk>](#chunk) | Dynamic number of chunks. Runs until the end of the blob. Unknown/no maximum amount. |
+
+## Data Types
+
+Terrain data is represented using a variety of different data types. Data described within them follows any conventions set in the [document conventions](#document-conventions).
+
+### Chunk
+
+The `Chunk` type is stored with a dynamic size dependent on the size of the voxels contained within, and the end of its data is marked by reaching its maximum voxel count of 32<sup>3</sup>. Voxels are ascendingly ordered by Y, then Z, then X based on their position in the chunk. Voxels are stored in rows of 32 units on each axis, each representing 4 units in world space.
```
Dekkonot, line 46: I'm confused. Is there a fixed voxel count per chunk regardless of chunk size, or is it dynamic? If it's fixed, how do larger chunks actually work?

checkraisefold: dynamic

checkraisefold:
further code commits are paused until structure is reviewed so I don't add the deserializer and have to immediately rewrite it

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+    pub fn new(x: i32, y: i32, z: i32) -> Self {
+        // Terrain isn't accepted by the engine with voxels past 2^23.
+        const CHUNK_MAX: i32 = 2i32.pow(23) / CHUNK_SIZE;
+        let (x, y, z) = (
+            x.clamp(-CHUNK_MAX, CHUNK_MAX),
+            y.clamp(-CHUNK_MAX, CHUNK_MAX),
+            z.clamp(-CHUNK_MAX, CHUNK_MAX),
+        );
+        Self(TerrainVec::new(x, y, z))
+    }
```
Dekkonot, line 90:
```suggestion
    pub fn new(x: i32, y: i32, z: i32) -> Self {
        // Terrain isn't accepted by the engine with voxels past 2^23.
        const CHUNK_MAX: i32 = 2i32.pow(23) / CHUNK_SIZE;

        Self(TerrainVec::new(
            x.clamp(-CHUNK_MAX, CHUNK_MAX),
            y.clamp(-CHUNK_MAX, CHUNK_MAX),
            z.clamp(-CHUNK_MAX, CHUNK_MAX),
        ))
    }
```
Same as the other `new` function.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+#[derive(Debug)]
+pub struct InvalidMaterialError(u8);
+
+impl std::error::Error for InvalidMaterialError {}
+
+impl fmt::Display for InvalidMaterialError {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "Invalid binary material value {:x?}", self.0)
+    }
+}
```
Dekkonot, line 173: We have `thiserror` for this. You should be able to `derive(Error)` for `InvalidMaterialError`.

Really though we want to use the `Error` type from the `error` module in this crate. You can go look at `MaterialColors` or `Attributes` for an idea of how this works.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+        };
+        voxel.set_occupancy(solid_occupancy, 0.0);
+
+        voxel
+    }
+
+    /// Constructs a new `Voxel` with a material, solid occupancy, and water
+    /// occupancy percentage. Takes advantage of the recently-released
+    /// Shorelines feature. Equivalent to data writeable from Roblox's `Terrain:WriteVoxelChannels`.
+    /// Occupancy values are between `0.0` and `1.0`, as a percentage of the voxel.
```
Dekkonot, line 206:
```suggestion
    /// Constructs a new `Voxel` with a material, solid occupancy, and water
    /// occupancy percentage.
    /// Equivalent to data writeable from Roblox's `Terrain:WriteVoxelChannels`.
    /// Occupancy values are between `0.0` and `1.0`, as a percentage of the voxel.
```
This comment is going to be here theoretically forever so we should avoid statements like 'recently-released'.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+        };
+        voxel.set_occupancy(solid_occupancy, water_occupancy);
+
+        voxel
+    }
+
+    fn get_encode_data(&self) -> (u8, u8) {
+        let solid_occupancy: u8 = (self.solid_occupancy * 255.0) as u8;
+        let water_occupancy: u8 = (self.water_occupancy * 255.0) as u8;
+        (solid_occupancy, water_occupancy)
```
Dekkonot, line 220:
```suggestion
        (
            (self.solid_occupancy * 255.0) as u8,
            (self.water_occupancy * 255.0) as u8,
        )
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+            /* Shorelines uses a new water occupancy value in the voxel data. Because of this,
+            Roblox uses a hack to avoid having to reduce their 6 bits of material ID freedom
+            by writing voxels with a count bit set to 1 and no count. This means we have to write
+            all voxels in the run length manually. */
+            return to_write
+                .iter()
+                .cycle()
+                .copied()
+                .take(to_write.len() * count as usize)
+                .collect();
```
Dekkonot, line 261:
```suggestion
            let len = to_write.len();
            return to_write
                .into_iter()
                .cycle()
                .take(len * count as usize)
                .collect();
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+    fn encode_run_length(&self, count: u16) -> Vec<u8> {
+        assert!(
+            (1..=256).contains(&count),
+            "Invalid voxel RLE count: {}",
+            count
+        );
+
+        let (solid_occupancy, water_occupancy) = self.get_encode_data();
+        let mut flag = self.material as u8;
+        let mut to_write: Vec<u8> = vec![];
```
Dekkonot, line 232:
```suggestion
        // The first value is a placeholder that'll be replaced with the `flag` later
        let mut to_write = vec![0x00];
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+            // Should we store the count (amount of voxels this run length) value?
+            flag |= 0b10000000;
+            if water_occupancy == 0 {
+                to_write.push((count - 1) as u8);
+            } else {
+                to_write.push(0);
+                to_write.push(water_occupancy);
+            }
+        }
+        to_write.insert(0, flag);
```
Dekkonot, line 249:
```suggestion
        to_write[0] = flag;
```

This avoids having to reshuffle the vector every time.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+                .copied()
+                .take(to_write.len() * count as usize)
+                .collect();
+        }
+        to_write
+    }
+
+    /// Sets occupancy data for a `Voxel`. Water occupancy is from the
+    /// new Shorelines feature. Occupancy values are between `0.0` and `1.0`,
+    /// as a percentage of the voxel.
```
Dekkonot, line 268:
```suggestion
    /// Sets occupancy data for a `Voxel`. Water occupancy is from the
    /// Shorelines feature. Occupancy values are between `0.0` and `1.0`,
    /// as a percentage of the voxel.
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+            self.solid_occupancy = water_occupancy;
+            self.water_occupancy = 0.0;
+            self.material = Material::Water;
+            return;
+        }
+
+        self.solid_occupancy = solid_occupancy;
+
+        // Full with a solid (non-air) material? We can't have any water.
+        if solid_occupancy == 1.0 && water_occupancy > 0.0 {
```
Dekkonot, line 292:
```suggestion
        if solid_occupancy == 1.0 {
```
I don't think we actually need to check the water occupancy here.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    pub fn set_material(&mut self, material: Material) {
+        self.material = material;
+
+        // Occupancy determination depends on material.
+        self.set_occupancy(self.solid_occupancy, self.water_occupancy)
+    }
+}
+
+// We don't iterate over the grid, so we can use a HashMap instead of BTreeMap to save performance.
```
Dekkonot, line 307:
```suggestion
```
I don't think anyone will ask questions about using a HashMap.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+    #[inline]
+    pub fn write_voxel(&mut self, position: &VoxelCoordinates, voxel: Voxel) {
+        self.grid.insert(*position, voxel);
+    }
+}
+
+impl TerrainSerializer for Chunk {
+    fn encode(&self) -> Vec<u8> {
+        // ~256 bytes if all voxels are air/base mat with maximum count. Double it
+        let mut data = Vec::<u8>::with_capacity(512);
```
Dekkonot, line 361:
```suggestion
        let mut data = Vec::with_capacity(512);
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+
+        let base_voxel: Voxel = Voxel {
```
Dekkonot, line 363:
```suggestion
        let base_voxel = Voxel {
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+            solid_occupancy: 1.0,
+            water_occupancy: 0.0,
+            material: self.base_material,
+        };
+
+        let mut pos_cursor = VoxelCoordinates::default();
+        let mut run_length_cursor = (0u16, &base_voxel);
```
Dekkonot, line 370: Storing these values in a tuple isn't a great idea. I would rather we just have two different variables for them.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+
+/// A container allowing the modification, encoding, and decoding of the
+/// `SmoothGrid` data used by Roblox's `Terrain` object.
+#[derive(Debug, Clone, Default, PartialEq)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(transparent)
+)]
+pub struct Terrain {
```
Dekkonot, line 422: This should be named `SmoothGrid` to keep it in theme with the other BinaryString data types.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+    /// Writes (or overwrites) a `Chunk` at the given position in this `Terrain`.
+    #[inline]
+    pub fn write_chunk(&mut self, position: &ChunkCoordinates, chunk: Chunk) {
+        self.world.insert(*position, chunk);
+    }
+}
+
+impl TerrainSerializer for Terrain {
+    fn encode(&self) -> Vec<u8> {
+        let mut data = Vec::<u8>::with_capacity(self.world.len() * 512);
```
Dekkonot, line 458:
```suggestion
        let mut data = Vec::with_capacity(self.world.len() * 512);
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+        data.extend([0x01, CHUNK_SIZE.ilog2() as u8]);
+
+        let mut chunk_cursor: Option<&ChunkCoordinates> = None;
```
Dekkonot, line 461:
```suggestion
        let mut chunk_cursor = None;
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+        for (position, chunk) in self.world.iter() {
```
Dekkonot, line 462:
```suggestion
        for (position, chunk) in &self.world {
```

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+    pub y: i32,
+    pub z: i32,
+}
+
+impl TerrainVec {
+    pub fn from_vec3(i: Vector3) -> Self {
+        Self {
+            x: i.x as i32,
+            y: i.y as i32,
+            z: i.z as i32,
```
Dekkonot, line 27: This isn't really all that rigorous. I don't know if we should expose this without actually checking if the `f32` is a valid `i32`. At the same time though, that's a hard thing to check, so maybe it's fine.

We should for sure document how it converts it though. Something like "naively converts the components to be `i32`s" would at least make it not surprising.

rbx_types/src/terrain.rs
```diff
@@ -0,0 +1,548 @@
...
+        let mut voxel = Voxel::new_with_water(Material::Grass, 1.0, 0.5);
+        for m in 2..=22 {
+            voxel.set_material(Material::try_from(m as u8).unwrap());
+            chunk.write_voxel(&VoxelCoordinates::new(m - 2, 0, 0), voxel);
+        }
+        terr.write_chunk(&ChunkCoordinates::default(), chunk.clone());
+        terr.write_chunk(&ChunkCoordinates::new(1, 0, 0), chunk.clone());
+
+        let encoded = base64::encode(terr.encode());
+        println!("{}", encoded);
```
Dekkonot, line 546: ...This test doesn't actually do anything. Can we make it panic if the output isn't identical to something Roblox emitted?

Dekkonot (commented):
Overall, I'd say that this is a good start to an encoder. The comments up above are more focused on the actual code than things like code structure.

Let's rename the `terrain.rs` module to `smooth_grid.rs`. I'd also like it if the `TerrainSerializer` trait could be killed off, since I don't think it adds anything. The `encode` and `decode` methods could be under the same `impl` blocks as the rest of the methods.

Architecturally I have some things I'd like to change but they involve files outside of this PR so I won't force them onto you.. Stuff like moving `smooth_grid` and `material_colors` into a submodule named `terrain` and having them share a `Material` enum, as an example. If you want to make those changes you're welcome to but otherwise I'll do it after this is merged.

docs/smooth-grid.md
checkraisefold: yes its dynamic
i thought i got this bit when I resolved it but alas I did not

rbx_types/src/terrain.rs
checkraisefold: this test is solely for development testing purposes while i was making the encoder so I could put it in place files and hence shouldn't make it out of the draft PR stage

rbx_types/src/terrain.rs
checkraisefold: i copypasted this from InvalidTypeError in rbx_binary iirc i dont know why i didnt use the error module for this

checkraisefold:
> Architecturally I have some things I'd like to change but they involve files outside of this PR so I won't force them onto you.. Stuff like moving `smooth_grid` and `material_colors` into a submodule named `terrain` and having them share a `Material` enum, as an example. If you want to make those changes you're welcome to but otherwise I'll do it after this is merged.

i could give this a shot but I was mainly wondering how the enum should be structured with it being like this? the easiest way would be to just put it in order but you'd have to include air and some other non-colorable materials i think and then implement an error for them??

rbx_types/src/terrain.rs
checkraisefold: the truncation is actually desirable behavior imo, supposed to be whole number and rounding down makes the most sense. or could remove this conversion entirely and force you to know exactly what you're doing

checkraisefold:
> i could give this a shot but I was mainly wondering how the enum should be structured with it being like this? the easiest way would be to just put it in order but you'd have to include air and some other non-colorable materials i think and then implement an error for them??

for this

Dekkonot:
> i could give this a shot but I was mainly wondering how the enum should be structured with it being like this? the easiest way would be to just put it in order but you'd have to include air and some other non-colorable materials i think and then implement an error for them??

There's a few ways I can imagine this working.

The first is that we just have a `TerrainMaterial` enum and error if you ever try to use a material that isn't allowed. This seems reasonable, but it's kind of a bummer that it would mean being unable to statically verify that a material was allowed.

Another option is to just say "to hell with it" and have multiple `TerrainMaterial` enums, which is how it's implemented right now. This solves the static analysis problem but it means that working with Terrain in rbx-dom requires two identically named enums, which is a terrible UX.

The final option I can see is that we have a centralized `TerrainMaterial` enum and then also module-specific material enums, which can be converted back and forth via `from` and `try_from`. I think I hate this idea, since it'd mean `SmoothGridMaterial::try_from(TerrainMaterial::Water).unwrap()` is a thing.

I'm inclined to go with the first option since it's the easiest and have the least obstructive UX possible. Moving us to a central `TerrainMaterials` enum is technically breaking, but I think I'm willing to accept it. We need to reorganize `rbx_types` either way, so if nothing else that can be done with this release. Curious to know what @kennethloeffler thinks though.

---

In terms of actual implementation, I imagine that each Terrain would have its own internal variant that it used that it just tried to convert into from the central `TerrainMaterials`. That'd allow each module to do its own thing while still giving a consistent public interface.

Dekkonot:
Can't wait for Ken for the rest of our days, so I'm gonna go ahead and say go for the plan I gave above with moving us to a central `TerrainMaterials` enum.

checkraisefold:
No that does not compile, just pushing work so far

checkraisefold:
So far my only hypothetical concern is performance - I feel like the algorithm in the encoder sucks
Decoder hopefully shouldn't be too bad in comparison at least

checkraisefold:
@Dekkonot 
module splitup is done, re-requesting review of the splitup before I move on to the decoder

checkraisefold:
needs review - only thing missing is decoder testing

rbx_types/src/terrain/smooth_grid.rs
```diff
@@ -0,0 +1,670 @@
...
+    pub fn decode(mut buffer: &[u8]) -> Result<Self, CrateError> {
+        let mut header = [0u8, 0u8];
+        if let Err(e) = buffer.read_exact(&mut header) {
+            return Err(SmoothGridError::from(e).into());
+        }
+
+        let [version, chunk_size] = header;
+        if version != 0x01 || chunk_size != 0x05 {
+            return Err(SmoothGridError::InvalidHeader(version, chunk_size).into());
+        }
```
Dekkonot, line 625: We should move these two be constants. Presumably then `CHUNK_SIZE` can be derived from the size we expect/support.

Then again, is it going to kill us to support a chunk size other than `0x05`? What would the consequence be?

Dekkonot (commented):
This obviously still needs tests, but nothing is jumping out at me right now implementation-wise. I have some questions and suggestions but they're mostly clerical.

However, outside of the implementation there's stuff we have to do for rbx-dom to actually support reading and writing files with this type!

This needs support for reading it in rbx_xml; there's a list of conversions in `rbx_xml\src\conversion.rs` that you need to add an entry for `SmoothGrid` to. 

In `rbx_binary`, this needs some conversions. You can take a look at my `MaterialColors` PR for a good idea of what that actually entails: https://github.com/rojo-rbx/rbx-dom/pull/323/files

In rbx-dom-lua, we need at least a pod in `rbx_dom_lua\src\EncodedValue.lua`. If we want to support live syncing it for Rojo, it also needs an entry in `rbx_dom_lua\src\customProperties.lua` but honestly that might be a lot of work.

Also obligatory "we need a changelog entry".

rbx_types/src/terrain/smooth_grid.rs
checkraisefold: When the engine loads chunk sizes other than `0x05`, if I recall according to my testing, it just converts the data into a normal chunk size. The behavior is pretty untested and I'm not sure if there's hidden issues with a different chunk size.

rbx_types/src/terrain/smooth_grid.rs
```diff
@@ -0,0 +1,670 @@
...
+    #[inline]
+    pub fn from_vec3(i: Vector3) -> Self {
+        Self(TerrainVec::from_vec3(i))
+    }
+}
+
+#[repr(u8)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+enum TerrainGridMaterial {
```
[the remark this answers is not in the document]
checkraisefold: Fairly sure it's to make it actually represent how it's stored in the format and also to save on memory

rbx_types/src/terrain/smooth_grid.rs
```diff
@@ -0,0 +1,670 @@
...
+            // Shorelines uses a new water occupancy value in the voxel data. Because of this,
+            // Roblox uses a hack to avoid having to reduce their 6 bits of material ID freedom
+            // by writing voxels with a count bit set to 1 and no count. This means we have to write
+            // all voxels in the run length manually.
+            let len = to_write.len();
+            return to_write
+                .into_iter()
+                .cycle()
+                .take(len * count as usize)
+                .collect();
```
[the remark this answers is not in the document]
checkraisefold: urmm... thats the whole point ! read the big comment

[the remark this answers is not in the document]
Dekkonot: I did, thank you for the snark. Now elaborate.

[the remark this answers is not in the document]
checkraisefold: because material ids are 6 bits, and the other 2 bits of the byte for the flag are taken up by write occupancy/write count toggles, they instead set the write count bit to 1 and then write a count of 0x00. then they write the water occupancy value after that. if there are multiple voxels in a row with the same water occupancy, that process is repeated for the entire amount, so the entire 3-4 byte sequence is just repeated for the entire run length. essentially run length encoding is just turned off for voxels with a nonzero water occupancy

rbx_types/src/terrain/smooth_grid.rs
Dekkonot: I suppose that's fair. We should probably just use Roblox's default.

Still, we should move both the version and chunk size to constants so that they're not just magic numbers floating around.

checkraisefold:
i have to make ChunkCoordinates/VoxelCoordinates serializable as strings that are acceptable as json keys. worst day

rbx_types/src/terrain/smooth_grid.rs
```diff
@@ -0,0 +1,656 @@
...
+pub struct ChunkCoordinates(TerrainVec);
+
+impl ChunkCoordinates {
+    /// Constructs a new `ChunkCoordinates` object.
+    #[inline]
+    pub fn new(x: i32, y: i32, z: i32) -> Self {
+        // Terrain isn't accepted by the engine with voxels past 2^23.
+        const CHUNK_MAX: i32 = 2i32.pow(23) / CHUNK_SIZE;
+        Self(TerrainVec::new(
+            x.clamp(-CHUNK_MAX, CHUNK_MAX),
```
krakow10: This should be correct upon construction rather than implicitly clamped.

krakow10 (changes requested):

rbx_types/src/terrain/material_colors.rs
```diff
@@ -32,7 +33,12 @@ impl MaterialColors {
     /// Retrieves the set color for the given material, or the default if
     /// none is set.
     #[inline]
-    pub fn get_color(&self, material: TerrainMaterials) -> Color3uint8 {
+    pub fn get_color(&self, material: TerrainMaterials) -> Result<Color3uint8, CrateError> {
```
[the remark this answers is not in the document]
Dekkonot: >  I don't believe the material_colors.rs file should be changed as part of Terrain SmoothGrid.

This was by my guidance. It's less type safe, but as a consumer it feels a lot more convenient to only have one TerrainMaterial enum to use when touching Terrain related structs.

krakow10:
Reading through the previous suggested changes above, I don't see why it's a problem to have distinct enums for TerrainGridMaterial and MaterialColor's TerrainMaterial when they are not in fact the same thing.  A centrally multi-way convertible enum sounds like a nice idea, but complicates the MaterialColors interface unnecessarily, especially when it's not actually related.  My proposal would be to use TerrainGridMaterial for smoothgrid and TerrainMaterial as-is for MaterialColors, but provide two-way conversion with From<material_colors::TerrainMaterial> and TryFrom<smooth_grid::TerrainGridMaterial>.

rbx_types/src/terrain/material_colors.rs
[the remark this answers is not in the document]
krakow10: I would agree if the conversion was infallible, but in this case it creates a new problem where previously there was none.

krakow10:
I've proposed #626 to make rbx_binary core functionality available to other crates which would make `write_interleaved_bytes` interleaved functions available to rbx_types as traits

PR #448, opened 2024-09-20
rbx_reflection: Superclass Iterator

krakow10:
I thought it was weird that database.superclasses() returns a Vec<ClassDescriptor> rather than an iterator, perhaps it should be replaced or supplemented by something like this?

### DIFF
--- a/rbx_reflection/src/database.rs
+++ b/rbx_reflection/src/database.rs
@@ -34,17 +34,17 @@ pub struct SuperClassIter<'a> {
     database: &'a ReflectionDatabase<'a>,
     descriptor: Option<&'a ClassDescriptor<'a>>,
 }
-impl<'a> SuperClassIter<'a>{
-	fn next_descriptor(&self)->Option<&'a ClassDescriptor<'a>>{
-		let superclass = self.descriptor?.superclass.as_ref()?;
-		self.database.classes.get(superclass)
-	}
+impl<'a> SuperClassIter<'a> {
+    fn next_descriptor(&self) -> Option<&'a ClassDescriptor<'a>> {
+        let superclass = self.descriptor?.superclass.as_ref()?;
+        self.database.classes.get(superclass)
+    }
 }
 impl<'a> Iterator for SuperClassIter<'a> {
     type Item = &'a ClassDescriptor<'a>;
     fn next(&mut self) -> Option<Self::Item> {
-    	let next_descriptor = self.next_descriptor();
-		std::mem::replace(&mut self.descriptor, next_descriptor)
+        let next_descriptor = self.next_descriptor();
+        std::mem::replace(&mut self.descriptor, next_descriptor)
     }
 }
 
--- a/rbx_reflection/src/database.rs
+++ b/rbx_reflection/src/database.rs
@@ -37,13 +37,9 @@ pub struct SuperClassIter<'a> {
 impl<'a> Iterator for SuperClassIter<'a> {
     type Item = &'a ClassDescriptor<'a>;
     fn next(&mut self) -> Option<Self::Item> {
-        let next_descriptor = self.descriptor.and_then(|descriptor| {
-            descriptor
-                .superclass
-                .as_ref()
-                .and_then(|class_name| self.database.classes.get(class_name))
-        });
-        core::mem::replace(&mut self.descriptor, next_descriptor)
+        let superclass = self.descriptor?.superclass.as_ref()?;
+        let next_descriptor = self.database.classes.get(superclass)?;
+        self.descriptor.replace(next_descriptor)
     }
 }
 
--- a/rbx_reflection/src/database.rs
+++ b/rbx_reflection/src/database.rs
@@ -77,7 +77,7 @@ impl<'a> ReflectionDatabase<'a> {
     pub fn superclasses_iter(&'a self, descriptor: &'a ClassDescriptor<'a>) -> SuperClassIter {
         SuperClassIter {
             database: self,
-            descriptor,
+            descriptor: Some(descriptor),
         }
     }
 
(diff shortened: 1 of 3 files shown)


rbx_reflection/src/database.rs
```diff
@@ -30,6 +30,23 @@ pub struct ReflectionDatabase<'a> {
...
+impl<'a> Iterator for SuperClassIter<'a> {
+    type Item = &'a ClassDescriptor<'a>;
+    fn next(&mut self) -> Option<Self::Item> {
+        let next_descriptor = self.descriptor.and_then(|descriptor| {
+            descriptor
+                .superclass
+                .as_ref()
+                .and_then(|class_name| self.database.classes.get(class_name))
+        });
+        core::mem::replace(&mut self.descriptor, next_descriptor)
```
Dekkonot, line 46:
```suggestion
        let superclass = self.descriptor?.superclass.as_ref()?;
        let next_descriptor = self.database.classes.get(superclass)?;
        self.descriptor.replace(next_descriptor)
```
You can just use `?` for this since we want to exit if these are `None` anyway. Also, `Option::replace` is just a better option for this.

rbx_reflection/src/database.rs
```diff
@@ -59,6 +76,15 @@ impl<'a> ReflectionDatabase<'a> {
         Some(list)
     }
 
+    /// Returns an iterator of superclasses for the provided ClassDescriptor. This
+    /// iterator will start with the provided class and end with `Instance`.
+    pub fn superclasses_iter(&'a self, descriptor: &'a ClassDescriptor<'a>) -> SuperClassIter {
+        SuperClassIter {
+            database: self,
+            descriptor,
+        }
```
Dekkonot, line 85:
```suggestion
            descriptor: Some(descriptor),
```
I have serious concerns if you did not even attempt to compile this code before submitting it, which is what this type mismatch indicates to me.

Dekkonot (changes requested):
I'm open to this conceptually, but not with the provided implementation; I've left a review of it.

In the future, try to open issues before making a pull request so we can determine if something is even something we want to do.

rbx_reflection/src/database.rs
Dekkonot: GitHub appears to have added this suggestion to the wrong line, but hopefully it's obvious what I meant

rbx_reflection/src/database.rs
krakow10: This is great!

rbx_reflection/src/database.rs
krakow10: Whoops, I was spewing some nonsense about Rust generators not being stable and got distracted.  Fixed

Dekkonot:
Would love a test (nothing too complicated, just enough to verify it works and includes the starting value and the ending values), and this needs a changelog entry.

Otherwise, it looks good now.

krakow10:
> I'm open to this conceptually, but not with the provided implementation; I've left a review of it.
> 
> In the future, try to open issues before making a pull request so we can determine if something is even something we want to do.

I figure suggesting an idea with code is better since it can be immediately iterated upon.  You're saying I should post a patch as a comment?  Or push to my fork and link the branch maybe?

Thanks for the review, the code I had was rather tragic.

Dekkonot:
What I mean is that you got lucky with this PR! There's a chance that I would have just said "no" and closed it, and your effort would've been wasted.

In the interest of everyone's time, I'd prefer we have issues where we at least go "yeah ok" before implementing things so that nobody's time is wasted.

krakow10:
> There's a chance that I would have just said "no" and closed it, and your effort would've been wasted.

An acceptable loss to me :smiley: 

My mind's compiler says this implementation (and mine) will miss the final ClassDescriptor ("Instance").  My first implementation before the PR would have missed the first one (the descriptor passed to the function) . I will look around the code and see what I can do about writing a test that looks like other tests.

krakow10:
Side question... why does the superclasses function return an Option<Vec<_>> instead of just Vec<_>?

Dekkonot (approved):

Dekkonot:
> Side question... why does the superclasses function return an Option<Vec<_>> instead of just Vec<_>?

Honest answer: It used to accept a `&str` as an argument and then I forgot to change it from a `Option<Vec<_>>` to a `Vec<_>`. I'm going to pretend that it's because you could construct a ClassDescriptor that doesn't exist in the database though, since that feels like it's more deliberate and makes sense. ;)

rbx_reflection_database/src/lib.rs
```diff
@@ -14,10 +14,32 @@ pub fn get() -> &'static ReflectionDatabase<'static> {
...
+    fn superclasses_iter_test() {
+        let database = get();
+        let part_class_descriptor = database.classes.get("Part");
+        let mut iter = database.superclasses_iter(part_class_descriptor.unwrap());
+        fn class_descriptor_eq(lhs: Option<&ClassDescriptor>, rhs: Option<&ClassDescriptor>) {
+            assert!(match (lhs, rhs) {
+                (Some(lhs), Some(rhs)) => std::ptr::addr_eq(lhs, rhs),
+                (None, None) => true,
+                _ => false,
+            })
```
Dekkonot, line 36: I don't support using pointer equality for this; it feels... unnecessary.

```suggestion
            match (lhs, rhs) {
                (Some(lhs), Some(rhs)) => assert_eq!(lhs.name, rhs.name),
                (None, None) => true,
                _ => panic!("{:?} != {:?}", lhs, rhs),
            }
```

krakow10: I figured there shouldn't be duplicate objects, but I suppose that's not part of this test.  I made some changes and worked in the print expression.

Dekkonot (approved):
Good enough. Thanks for the contribution!

krakow10:
Is it also a good enough time for a release? I'm hoping to drop the patched packages I've been maintaining now that UTF8 conversion is merged.

Dekkonot:
I can make a new release in a few days if it's necessary; I generally done cut them until we need them, but if it's a blocker for you I can make it a priority.

krakow10:
All my packages are compiling fine, so it'd just be a convenience to me.  You're welcome to release whenever you feel is best.

PR #450, opened 2024-09-24
Override the default value of `WorldPivotData` to be `null`

Dekkonot:
`Model.WorldPivotData` isn't serialized as a real value by Roblox unless it's explicitly set. We should probably mimic this behavior.

The reason it *hasn't* been this value is because of pivot migrations. If a Model is inserted without `NeedsPivotMigration` set to `false` (as is the case in rbx-reflector), Roblox generates a pivot and serializes it, rather than leaving it blank.

If you instead insert a model that has `NeedsPivotMigration` set to `false` but doesn't have any other properties set, you can see that `WorldPivotData` is left blank.


kennethloeffler (approved):
I agree with the reasoning, and I can verify that

> If you instead insert a model that has NeedsPivotMigration set to false but doesn't have any other properties set, you can see that WorldPivotData is left blank.

matches the behavior of Roblox Studio's serializer, so LGTM!

I wonder if there are any other defaults that depend on the values of other properties... 🤔

Dekkonot:
> I wonder if there are any other defaults that depend on the values of other properties... 🤔

That one might be hard to test, but we could always go and insert every Instance manually and see if anything differs dramatically...

PR #452, opened 2024-10-06
rbx_reflection: has_superclass

krakow10:
Hi, I keep writing this function and thought it could be provided here.  Open to suggestions for the function name and signature!

### DIFF
--- a/rbx_reflection/src/database.rs
+++ b/rbx_reflection/src/database.rs
@@ -86,6 +86,17 @@ impl<'a> ReflectionDatabase<'a> {
         }
     }
 
+    /// Walks the class tree, checking if superclass matches
+    /// any class names the class inherits from
+    pub fn class_is_a(
+        &'a self,
+        descriptor: &'a ClassDescriptor<'a>,
+        superclass_descriptor: &'a ClassDescriptor<'a>,
+    ) -> bool {
+        self.superclasses_iter(descriptor)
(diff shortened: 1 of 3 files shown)


rbx_reflection/src/database.rs
```diff
@@ -86,6 +86,17 @@ impl<'a> ReflectionDatabase<'a> {
         }
     }
 
+    /// This mimics the behavior of the Roblox Lua Instance:IsA(ClassName) method.
+    /// Returns whether superclass_descriptor is a superclass of descriptor.
```
Dekkonot, line 90:
```suggestion
    /// This mimics the behavior of the Roblox method `Instance:IsA(ClassName)` method.
    /// Returns whether `superclass_descriptor` is a superclass of `descriptor`.
```

Dekkonot (commented):
I'm open to including this. I think `class_is_a` is a bit weird, but the type signature is fine.

This needs an entry in `rbx_reflection`'s changelog though.

krakow10:
> I'm open to including this. I think `class_is_a` is a bit weird, but the type signature is fine.

Do you have an idea of what to change it to?  The Roblox function is also poorly named.

Dekkonot:
I think `has_superclass` is probably good. If it has a doc comment explaining what it does, I think the name is intuitive enough that it won't confuse anyone.

Dekkonot (approved):

PR #453, opened 2024-10-06
rbx_reflection: Superclass Iterator v2

krakow10:
Depends on #452 

I had an idea about how to simplify the implementation of the superclass iterator.  The function signature is more like the Roblox function where invalid class names are ignored and return false.  I could go either way on this.  On the one hand, behaviour does match Roblox better, but forcing the user to provide valid classes has some merit in terms of reporting an error rather than silently failing.

### DIFF
--- a/rbx_reflection/src/database.rs
+++ b/rbx_reflection/src/database.rs
@@ -86,6 +86,17 @@ impl<'a> ReflectionDatabase<'a> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


krakow10:
I changed my mind.  Explicit class name verification is best, the roblox behaviour of silently failing can be implemented downstream.  Long live Superclass Iterator v1

PR #458, opened 2024-10-23
Add SharedString defaults to reflection database

kennethloeffler:
After #414, `SharedString` values are now representable in serde formats, meaning that it's now possible for the reflection database to define defaults for `SharedString` properties. This will be useful for Rojo's syncback, when `SharedString` properties need to be written to JSON files: https://github.com/UpliftGames/rojo/pull/6#discussion_r1770646246.


Dekkonot (approved):

kennethloeffler:
Whoops, looks like some `SharedString` tests failed. I'll spin through in a bit to figure out what went wrong!

Dekkonot:
And after my approval too. 😔

This looks like it's just a consequence of adding defaults. Since they're present, we no longer generate an empty one for SharedString properties. This changes the index of other strings in the file though, since one is no longer present.

rbx_binary/src/serializer/state.rs
```diff
@@ -449,7 +449,7 @@ impl<'dom, 'db, W: Write> SerializerState<'dom, 'db, W> {
                 // There's no assurance that the default SharedString value
                 // will actually get serialized inside of the SSTR chunk, so we
                 // check here just to make sure.
-                if let Cow::Owned(Variant::SharedString(sstr)) = &default_value {
+                if let Variant::SharedString(sstr) = default_value.borrow() {
```
kennethloeffler, line 452: Turns out we were matching against `Cow::Owned` when checking if we need to insert default SharedStrings instead of just borrowing

Dekkonot (approved):

PR #464, opened 2024-10-29
Preallocate property tables during deserialization

kennethloeffler:
This PR adds a new method `InstanceBuilder::with_property_capacity` which can preallocates `InstanceBuilder.properties`, and makes the rbx_xml and rbx_binary deserializers use it. Since we know how many properties any given instance will (likely) contain via the reflection database, this allows us to avoid resizes during deserialization.

On my machine, this improves rbx_binary's "Deserialize 10,000 Parts" benchmark by ~17%, and reduces the number of heap allocations by ~32%.

The fact that we're using the database's default properties to determine the capacity is a little hairy - we've discussed normalizing default properties in the database in #418, so we should remain mindful of this optimization, lest we inadvertently regress performance.

### DIFF
--- a/rbx_dom_weak/CHANGELOG.md
+++ b/rbx_dom_weak/CHANGELOG.md
@@ -93,6 +93,9 @@ where
 ### Other changes
 * Added `UstrMapExt`, a helper trait providing convenience methods `UstrMap::new` and `UstrMap::with_capacity`.
 * Added re-exports for `ustr` (a convenience function for creating `Ustr`s), `Ustr`, `UstrMap`, and `UstrSet`.
+* Added `InstanceBuilder::with_property_capacity`, which can preallocate an `InstanceBuilder`'s property table. [#464]
+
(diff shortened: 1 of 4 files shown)


rbx_dom_weak/CHANGELOG.md
```diff
@@ -93,6 +93,9 @@ where
 ### Other changes
 * Added `UstrMapExt`, a helper trait providing convenience methods `UstrMap::new` and `UstrMap::with_capacity`.
 * Added re-exports for `ustr` (a convenience function for creating `Ustr`s), `Ustr`, `UstrMap`, and `UstrSet`.
+* Added `InstanceBuilder::with_property_capacity`, which can preallocate an `InstanceBuilder`'s property table. [#464]
+
+[#4641]: https://github.com/rojo-rbx/rbx-dom/pull/464
```
Dekkonot, line 98:
```suggestion
[#464]: https://github.com/rojo-rbx/rbx-dom/pull/464
```

PR #465, opened 2024-10-29
Reserve space for additional instances in binary deserializer

kennethloeffler:
This PR adds a new method `WeakDom::reserve` that can preallocate additional space for instances, and makes the binary deserializer use it before adding instances.

On my machine, this improves rbx_binary's "Deserialize 10,000 Parts" benchmark by ~4%, reduces the number of heap allocations by a negligible amount, but allocates ~2.4MB less memory. 

I'm not sure if this should also call `reserve` on `WeakDom.unique_ids`, because UniqueIds won't always be present in inputs, and it could take a possibly significant amount of space for no reason.

### DIFF
--- a/rbx_dom_weak/CHANGELOG.md
+++ b/rbx_dom_weak/CHANGELOG.md
@@ -96,7 +96,7 @@ where
 * Added `InstanceBuilder::with_property_capacity`, which can preallocate an `InstanceBuilder`'s property table. [#464]
 * Added `WeakDom::reserve`, which can preallocate additional space for instances in the `WeakDom`. [#465]
 
-[#465]: https://github.com/rojo-rbx/rbx-dom/pull/464
+[#465]: https://github.com/rojo-rbx/rbx-dom/pull/465
 [#464]: https://github.com/rojo-rbx/rbx-dom/pull/464
 
 ## 2.9.0 (2024-08-22)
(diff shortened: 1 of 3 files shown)


rbx_dom_weak/CHANGELOG.md
```diff
@@ -94,7 +94,9 @@ where
 * Added `UstrMapExt`, a helper trait providing convenience methods `UstrMap::new` and `UstrMap::with_capacity`.
 * Added re-exports for `ustr` (a convenience function for creating `Ustr`s), `Ustr`, `UstrMap`, and `UstrSet`.
 * Added `InstanceBuilder::with_property_capacity`, which can preallocate an `InstanceBuilder`'s property table. [#464]
+* Added `WeakDom::reserve`, which can preallocate additional space for instances in the `WeakDom`. [#465]
 
+[#465]: https://github.com/rojo-rbx/rbx-dom/pull/464
```
Dekkonot, line 99:
```suggestion
[#465]: https://github.com/rojo-rbx/rbx-dom/pull/465
```

Dekkonot (commented):
If we _don't_ reserve extra space in `unique_ids` it'll make deserializing a place file have a fair few allocations still. This optimization is still worth doing but it's something to keep in mind.

kennethloeffler:
It doesn't seem to make much a difference for a 6MB rbxl file on my machine, but I guess it won't hurt, since massive rbxm files aren't very common anyway

Dekkonot:
Hilariously for Adopt Me (65MB), it seems to actually make it take longer to deserialize with the UniqueId reservation. If I went through the trouble of checking our full Rojo project it might be more dramatic, since we have around 6000 models with `UniqueId` written in them, but that's work I don't feel like doing right now.

The difference for Adopt Me is like 11%, so we may want to just not bother with `UniqueId` reservation.

kennethloeffler:
Well, that's a bit weird! 🤔

Dekkonot (approved):

PR #466, opened 2024-10-30
Use ahash for most HashMaps and HashSets

kennethloeffler:
This PR replaces most `HashMap`s with `AHashMap`. I did it the lazy way with type aliases, didn't do anything to rbx_xml (yet?), and haven't touched the changelogs yet, because I am tired and I want to go to bed! It's kind of invasive, and might make the library harder to use, but we do get a nice performance boost from it.

On my machine, this improves rbx_binary's "Deserialize 10,000 Parts" and "Serialize 10,000 Parts" benchmarks by ~15% and ~19%, respectively.

### DIFF
--- a/rbx_dom_weak/CHANGELOG.md
+++ b/rbx_dom_weak/CHANGELOG.md
@@ -89,6 +89,14 @@ where
     V: Into<Variant>,
     I: IntoIterator<Item = (K, V)>,
 ```
+* Started using [ahash](https://docs.rs/ustr/latest/ustr/) for hash maps, consequently changing the signature of `WeakDom::into_raw` from
+```rust
+pub fn into_raw(self) -> (Ref, HashMap<Ref, Instance, RandomState>) {
+```
+to
+```rust
+pub fn into_raw(self) -> (Ref, HashMap<Ref, Instance, ahash::RandomState>) {
+```
 
 ### Other changes
 * Added `UstrMapExt`, a helper trait providing convenience methods `UstrMap::new` and `UstrMap::with_capacity`.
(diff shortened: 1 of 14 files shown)


Dekkonot:
>  It's kind of invasive, and might make the library harder to use, but we do get a nice performance boost from it.

I do have some concerns about changing the reflection database to use `AHashMap`, since we aren't the only consumers. The only other public interface this would impact is `WeakDom::into_raw` though and I think it's probably fine to leak that. I don't know how many people are using that API but it's probably fine to slightly break it since we're already making a breaking change anyway.

This will impact Argon and Rojo, but those are the only ones I can think of off the top of my head. They'll already have to do a bunch of changes to update to rbx_weak_dom 3.0 though, so it's probably fine.

kennethloeffler:
After some further benchmarking, using ahash in rbx_reflection is a bit of a wash. It didn't effect rbx_binary's deserialization performance at all, and if anything made rbx_binary's serialization performance worse. Good for me, fewer changelogs to touch :D

rbx_dom_weak/CHANGELOG.md
```diff
@@ -89,6 +89,14 @@ where
     V: Into<Variant>,
     I: IntoIterator<Item = (K, V)>,
 ```
+* Started using [ahash](https://docs.rs/ustr/latest/ustr/) for hash maps, consequently changing the signature of `WeakDom::into_raw` from
```
Dekkonot, line 92:
```suggestion
* Started using [ahash](https://docs.rs/ahash/latest/ahash/) for hash maps, consequently changing the signature of `WeakDom::into_raw` from
```

Dekkonot (approved):

PR #467, opened 2024-10-30
Change type of InstanceBuilder.properties to Vec<(Ustr, Variant)>

kennethloeffler:
This PR changes the type of `InstanceBuilder.properties` from `UstrMap<Variant` to `Vec<(Ustr, Variant)>` because I noticed that we were spending a significant amount of time just inserting into this HashMap during deserialization, and we get no benefit from this being a HashMap. The number of keys is usually small, and we only do a lookup here:
https://github.com/rojo-rbx/rbx-dom/blob/e909f454b592e975a48ee573922678d791e7ed84/rbx_binary/src/deserializer/state.rs#L183-L191
which is a *very* cold path.

On my machine, this improves rbx_binary's "Deserialize 10,000 Parts" benchmark by ~11%, and will be even more significant on larger workloads.

### DIFF
--- a/rbx_dom_weak/src/dom.rs
+++ b/rbx_dom_weak/src/dom.rs
@@ -123,7 +123,11 @@ impl WeakDom {
                     parent,
                     name: builder.name,
                     class: builder.class,
-                    properties: builder.properties,
+                    properties: builder
+                        .properties
+                        .into_iter()
+                        .map(|(k, v)| (k, v))
+                        .collect(),
                 },
             );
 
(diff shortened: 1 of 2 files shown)


Dekkonot (approved):
I was skeptical of this one because it turns `has_property` into an O(n) operation rather than (theoretically) O(1). However, I just did a benchmark and on my machine the worst-case scenario of having all 3691 properties took 47200 nanoseconds... So I think we're probably fine.

kennethloeffler:
RIght, linear time property lookups on an `InstanceBuilder` would only be bad if it were an incredibly common operation - but it isn't, both for us and for our consumers. In this case, the cost of hash table bookkeeping outweighs the benefit of constant time lookup. 

In contrast, attempting this kind of optimization on `Instance.properties` would actually make things worse, since the rbx_binary serializer has to look up a lots of random properties over and over again, and consumers often want to do the same thing

PR #471, opened 2024-11-01
Don't bail for MaterialColors and Attributes parse failures

kennethloeffler:
In #293, we changed the rbx_binary deserializer to bail completely when it fails to read attributes. This turns out to be a pretty bad idea, because it means when Roblox changes the attributes format, any files containing the new version cannot be parsed at all. 

This PR allows MaterialColors and Attributes to be parsed as BinaryStrings when they cannot be parsed into their specialized types in both rbx_binary and rbx_xml, and changes the corresponding rbx_dom_lua custom property codecs to fail gracefully under these conditions. I'm thinking we might want to do this for Tags too, but I'm not sure... let me know what you think!

### DIFF
--- a/rbx_binary/src/deserializer/error.rs
+++ b/rbx_binary/src/deserializer/error.rs
@@ -79,11 +79,4 @@ pub(crate) enum InnerError {
         expected_type_id: u8,
         actual_type_id: u8,
     },
-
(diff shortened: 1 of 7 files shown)


Dekkonot (approved):
I find it unlikely the serialization of Tags will change. I'm also not sure if it can even fail, since the worst case is a blob that contains no zeros so our implementation just assumes it's one Tag.

Issue #460, opened 2024-10-27
Add documentation for SecurityCapabilities


kennethloeffler:
Back in #358, we added bare-bones support for the `SecurityCapabilities` type. At the time, this data type did not have any user facing functionality, making it impossible for to infer the meaning of its bits. So, we opted to treat it as an opaque integer, and did not create any documentation. 

Since then, [Roblox has released Script Capabilities as a beta feature](https://devforum.roblox.com/t/script-capabilities-preview-client-beta/3228100), so we should be able to start poking at values of this type, and determine the meaning of the bits.

I'm not sure if we should implement anything against our findings until the feature leaves beta since it's possible that exact details may change, but this won't stop us from getting a head start on documentation!

Dekkonot:
Based on [this devforum post](https://devforum.roblox.com/t/script-capabilities-preview-client-beta/3228100/62?u=dekkonot) we probably want to wait on implementing it, yeah.

Documenting it is probably fine though; it'll if nothing else be interesting to see if it changes and _how_ it changes.

EpixScripts:
The capability bits as of v648 are:
```
0: Plugin
1: LocalUser
2: WritePlayer
3: RobloxScript
4: RobloxEngine
5: NotAccessible
8: RunClientScript
9: RunServerScript
11: AccessOutsideWrite
15: Unassigned
16: AssetRequire
17: LoadString
18: ScriptGlobals
19: CreateInstances
20: Basic
21: Audio
22: DataStore
23: Network
24: Physics
25: UI
26: CSG
27: Chat
28: Animation
29: Avatar
30: Input
31: Environment
32: RemoteEvent
61: PluginOrOpenCloud
62: Assistant
```
Notice how only some of the Luau `Enum.SecurityCapability` EnumItems map to this bitfield. The lowest 8 bits map to what was previously identity permissions.
Additionally, the Instance property setter for `Instance.Capabilities` is hardcoded to not let you extend an Instance's capabilities into one of the lowest 8 bits or one of the highest 3 bits, even if the current thread has that capability.
See:
```cpp
v2 = *newCapabilities & 0x1FFFFFFFFFFFFF00LL;
// ...
if ( (v2 & v4[4]) != v2 )
    capability_error_action(v2 & ~v4[4], "extend", "Capabilities")
```

kennethloeffler:
[...]

PR #483, opened 2025-01-02
SecurityCapabilities Documentation (#460) 

phoriah:
Loosely based on https://github.com/rojo-rbx/rbx-dom/issues/460#issuecomment-2452263778
Using the methods of trial & error (as described in https://github.com/rojo-rbx/rbx-dom/issues/460#issuecomment-2452285256)
With information from:
- https://create.roblox.com/docs/scripting/capabilities
- https://github.com/Pseudoreality/Roblox-Identities


docs/xml.md
```diff
@@ -561,6 +562,65 @@ A `Ref` value pointing to an `Item` with a `referent` of `RBX466F72207262782D646
...
+| 3                | RobloxScript         | Indicates a Roblox internal script.                                                                 |
+| 4                | RobloxEngine         | Accesses engine-level features.                                                                     |
+| 5                | NotAccessible        | Members with this security are intended to only be written to outside of a running game.            |
+| 8                | RunClientScript      | [Allows execution of LocalScript or Script with a RunContext value of Client on the client.](https://create.roblox.com/docs/scripting/capabilities#execution-control) |
+| 9                | RunServerScript      | [Allows execution of Script with a RunContext value of Server on the server.](https://create.roblox.com/docs/scripting/capabilities#execution-control) |
+| 11               | AccessOutsideWrite   | [Allows fetching and receiving instances from outside the sandboxed container.](https://create.roblox.com/docs/scripting/capabilities#instance-access-control) |
```
Dekkonot, line 595: I would prefer that these links be split out of this table and placed at the bottom, as it makes aligning the table + reading it easier.

docs/xml.md
```diff
@@ -561,6 +562,65 @@ A `Ref` value pointing to an `Item` with a `referent` of `RBX466F72207262782D646
...
+<SecurityCapabilities name="Example">0</SecurityCapabilities>
+```
+
+A `SecurityCapabilities` value with `RunClientScript` (2 ^ 8 = 256) & `RunServerScript` (2 ^ 9 = 512) capabilities set appears as follows:
+
+```xml
+<SecurityCapabilities name="Example">768</SecurityCapabilities>
+```
+
+The following table maps each bit in the field to a corresponding capability:
```
Dekkonot, line 583: I am assuming that bit 0 corresponds to the least-significant bit, but we should specify.

Dekkonot (commented):
Sorry for the delay in getting to this, I've been busy!

This looks good, though I haven't verified it. We'll cross that bridge when we implement it. For now, I've left some feedback and this needs to be replicated for the binary spec file.

PR #488, opened 2025-02-08
Generic WeakDom using AsRef<Instance>

krakow10:
Hey I had this idea for allowing WeakDom to be usable with a generic type parameter.  The problem I am trying to solve is that additional metadata about instances (such as `RBXScriptSignal`) cannot be directly owned by the WeakDom.  Each additional piece of metadata has to live in a `HashMap<Ref,_>` somewhere else and be managed every time the dom is updated, so this essentially becomes a garbage collection problem.  Let me know your thoughts!

<details>
<summary>Here's a code example of how I imagine this being used:</summary>

```Rust
use my_lib::RBXScriptSignal;
use rbx_dom_weak::{GenericWeakDom, Instance, InstanceBuilder};

struct MyInstance {
    instance: Instance,
    render_stepped: RBXScriptSignal,
}
impl AsRef<Instance> for MyInstance {
    fn as_ref(&self) -> &Instance {
        &self.instance
    }
}
impl AsMut<Instance> for MyInstance {
    fn as_mut(&mut self) -> &mut Instance {
        &mut self.instance
    }
}
impl From<Instance> for MyInstance {
    fn from(instance: Instance) -> Self {
        Self {
            instance,
            render_stepped: RBXScriptSignal::new(),
        }
    }
}
impl Into<Instance> for MyInstance {
    fn into(self) -> Instance {
        self.instance
    }
}

fn main() {
    let builder = InstanceBuilder::new("RunService");

    let cool_dom = GenericWeakDom::<MyInstance>::new(builder);

    dbg!(cool_dom.root().render_stepped);
}
```

Of course `MyInstance` would have some sort of collection of `RBXScriptSignal` instead of every instance having a `render_stepped` property, but this is just an example.

</details>

### DIFF
--- a/rbx_binary/src/deserializer/mod.rs
+++ b/rbx_binary/src/deserializer/mod.rs
@@ -21,11 +21,12 @@ pub use self::error::Error;
 /// use std::io::BufReader;
 ///
 /// use rbx_binary::Deserializer;
+/// use rbx_dom_weak::WeakDom;
 ///
 /// let input = BufReader::new(File::open("File.rbxm")?);
 ///
 /// let deserializer = Deserializer::new();
-/// let dom = deserializer.deserialize(input)?;
+/// let dom: WeakDom = deserializer.deserialize(input)?;
 ///
 /// // rbx_binary always returns a DOM with a DataModel at the top level.
 /// // To get to the instances from our file, we need to go one level deeper.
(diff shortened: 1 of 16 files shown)


krakow10:
https://github.com/rojo-rbx/rbx-dom/issues/514#issuecomment-2822758516 makes it clear that this is not upstreamable.

PR #495, opened 2025-03-10
Implement support for `Content` type

Dekkonot:
- Renames old `Content` to `ContentId`
- Adds new `Content` to represent Roblox's [`Content`](https://create.roblox.com/docs/reference/engine/datatypes/Content) type
- Implements support for both serializing and deserializing to rbx_xml, rbx_binary, and rbx_dom_lua
- Re-bootstraps the database with the new changes, which makes it incompatible with old versions of rbx-dom
- Does not include migrations for ContentId properties to Content

Any notes, comments, or complaints are welcome and appreciated.


jovannic:
Thanks so much for doing this.

rbx_binary/src/deserializer/state.rs
```diff
@@ -1439,6 +1453,57 @@ rbx-dom may require changes to fully support this property. Please open an issue
...
+                    }
+
+                    let object_count = chunk.read_le_u32()? as usize;
+                    let mut objects: VecDeque<i32> = vec![0; object_count].into();
+                    chunk.read_referent_array(objects.make_contiguous())?;
+
+                    let external_count = chunk.read_le_u32().unwrap() as usize;
+                    // We are advised by Roblox to just ignore this, as it's
+                    // meant for internal use. If we want to use it in the
+                    // future, it's a referent array.
```
jovannic, line 1474: If you want to say it's sometimes used for patching up references to the original object if it still exists, that's cool. If you ever see referents >= `objects.len()` in general that's usually what they're for, patching up external refs.

jovannic (commented):
FWIW, looks correct to me.

rbx_types/src/content.rs
```diff
@@ -1,61 +1,135 @@
+use crate::Ref;
+
 /// A reference to a Roblox asset.
 ///
-/// When exposed to Lua, this is just a string.
+/// This is exposed in Roblox the `Content` type. For the legacy type equivalent
```
kennethloeffler, line 5:
```suggestion
/// This is exposed in Roblox as the `Content` type. For the legacy type equivalent
```

rbx_binary/src/deserializer/state.rs
```diff
@@ -440,11 +441,23 @@ This may cause unexpected or broken behavior in your final results if you rely o
...
-                        let value: Content = chunk.read_string()?.into();
-                        add_property(instance, &property, value.into());
+                        let value = chunk.read_string()?;
+                        let content = if value.is_empty() {
+                            Content::none()
+                        } else {
+                            Content::from_uri(value)
+                        };
+                        add_property(instance, &property, content.into());
                     }
```
kennethloeffler, line 461: Does this arm need to be here? Are there any properties that have a serialized type of `String` with a canonical type of `Content`?

kennethloeffler (commented):
Implementation looks accurate to the spec. We will need to make sure to add migrations from the old `ContentId` properties to the new `Content` properties before releasing this, because I suspect that mixing them in the same file will result in incorrect behavior like with `Font` and `FontFace`.

Just a couple comments:

rbx_binary/src/deserializer/state.rs
jovannic: Legacy `ContentId` uses `string` serialization for binary format, but we don't support `string` for new `Content` properties.

jovannic:
Yeah migrations would be important. Internally in engine the legacy `ContentId` properties just read/write from/to `Content` as storage. The ContentId property just becomes a filtered view of the Content property, e.g. if `blah.XContent.Object` is non-nil then `blah.XId` will be empty string, and writes to `blah.XId` do the equivalent of `blah.XContent = Content.fromUri(blah.XId)`

It's also important to only write one in a serialized file, because they get read/written in undefined order and Content can represent values that ContentId cannot. If the latter is read last then it can overwrite valid object states.

rbx_binary/src/deserializer/state.rs
Dekkonot: None that I know of! I can just remove this; it was part of me being on autopilot and replacing various parts of `ContentId` with `Content`.

Dekkonot:
Yeah migrations are coming. The plan was to add them to a separate PR, but I can add them to this one.

kennethloeffler (approved):

PR #506, opened 2025-03-31
Update database and release new rbx_reflection_database version

Dekkonot:
Thanks to #505 we are now aware that a mistake was made when patching things. We should publish a new version of the reflection database ASAP to avoid problems. This can happily just be a patch.

Once this is live, I'll probably yank `1.0.0+roblox-666` in favor of `1.0.1+roblox-666`.


Dekkonot:
The information that was communicated by Roblox didn't say so:
image

That said, we can for sure add one. I was just using the list Roblox gave when I made the patch list.

kennethloeffler:
Fine with leaving it out for now then, but we should make sure to check whether we'll need it when Roblox rolls out the serialization changes

Dekkonot:
Optimistically assuming this is fine and I'm going to merge it.

PR #511, opened 2025-04-16
Add `Content::as_object` and `Content::as_uri`

krakow10:
This is convenient if you want to always assume the Content is a Uri or an Object.

### DIFF
--- a/rbx_types/src/content.rs
+++ b/rbx_types/src/content.rs
@@ -58,7 +58,7 @@ impl Content {
 
     /// Returns Some(&str) if `Content` is a Uri.
     #[inline]
-    pub fn get_uri(&self) -> Option<&str> {
+    pub fn as_uri(&self) -> Option<&str> {
         match self {
             Self(ContentType::Uri(uri)) => Some(uri.as_str()),
             _ => None,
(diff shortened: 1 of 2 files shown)


rbx_types/src/content.rs
```diff
@@ -55,6 +55,24 @@ impl Content {
     pub fn into_value(self) -> ContentType {
         self.0
     }
+
+    /// Returns Some(&str) if `Content` is a Uri.
```
Dekkonot, line 59: This isn't a very helpful doc comment. It's essentially just saying what the type signature already communicates. I'd prefer something like this:
```suggestion
    /// If this `Content` is a URI, returns the URI. Otherwise, returns `None`.
```

rbx_types/src/content.rs
```diff
@@ -55,6 +55,24 @@ impl Content {
...
+    #[inline]
+    pub fn as_uri(&self) -> Option<&str> {
+        match self {
+            Self(ContentType::Uri(uri)) => Some(uri.as_str()),
+            _ => None,
+        }
+    }
+
+    /// Returns Some(Ref) if `Content` is an Object.
```
Dekkonot, line 68: Same as the other function. This is not a very helpful doc comment since it's essentially just repeating the type signature without any more information.

Dekkonot (approved):

PR #513, opened 2025-04-18
Add Userdata Field To rbx_dom_weak::Instance Behind Feature Flag

krakow10:
Alternative to #488 using only 11 lines of code.  The goal is to implement RBXScriptSignal at runtime, this is one such solution.


krakow10:
https://github.com/rojo-rbx/rbx-dom/issues/514#issuecomment-2822758516 makes it clear that this is not upstreamable.

PR #516, opened 2025-04-22
Const Ref Constructor

krakow10:
There is no way to have a Ref as a compile time constant like so:
```rust
pub(crate) const NIL_REF:Ref=Ref::from_value(u128::MAX);
```

The alternative is to litter the code with runtime construction using FromStr:
```rust
let nil_ref:Ref="ffffffffffffffffffffffffffffffff".parse().unwrap();
```
The intended use case for this is to implement runtime nil instances with a specific constant Ref that refers to the "root nil instance" within the dom. The implementation maps the None variant of `Option<Ref>` into `NIL_REF` when assigning parent:
```rust
pub struct Instance{
	referent:Ref,
}
impl Instance{
	pub const fn nil()->Self{
		Self{
			referent:NIL_REF
		}
	}
}
impl mlua::UserData for Instance{
	fn add_fields<F:mlua::UserDataFields<Self>>(fields:&mut F){
		fields.add_field_method_set("Parent",|lua,this,new_parent:Option<Instance>|{
			let parent=new_parent.unwrap_or(Instance::nil());
			dom.transfer_within(this.referent,parent.referent);
			Ok(())
		});
	}
}
```

### DIFF
--- a/rbx_types/CHANGELOG.md
+++ b/rbx_types/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased Changes
 
-* Added `Ref::from_value` and  `Ref::from` to construct a Ref from a u128. ([#516])
+* Added `Ref::some` to construct a Ref from a u128. ([#516])
 * Added `Content::as_object` and `Content::as_uri` to assume the respective type (optional value). ([#511])
 * Added `Content::into_value` to support converting a `Content` into its underlying value. ([#507])
 
--- a/rbx_types/CHANGELOG.md
+++ b/rbx_types/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased Changes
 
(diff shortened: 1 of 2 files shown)


krakow10:
My above implementation of nil instances is stupid when I can just be using `Ref::none()`, but that requires #517.  This may still be a useful change regardless.

Dekkonot:
I'm not sold on allowing people to potentially pass `0` as a parameter, as it'd result in the `None` referent, and requiring them to provide a `NonZeroU128` is leaking an implementation detail. I think we're better off considering #517.

krakow10:
I've changed the type signature of `from_value` to return `Option<Ref>` and removed `From<u128>`.

> pass 0 as a parameter, as it'd result in the None referent

How is this different from FromStr?  In the tests the None referent is constructed using a string:
```rust
Ref::from_str("00000000000000000000000000000000").unwrap()
```

PR #517, opened 2025-04-23
Allow `WeakDom::transfer_within` dest_parent_ref to be the None Ref

krakow10:
Allow the following code to run without panic:
```rust
dom.transfer_within(instance.referent(),Ref::none());
```


krakow10:
@Dekkonot Hi, this is the only out-of-tree patch I'm currently relying on in my fork, can it be included in the next release?

PR #522, opened 2025-04-24
Swap LZ4 crate to `lz4_flex`

### LINKED ISSUE #255: rbx_binary: Replace C-binding crate `lz4` with a native Rust LZ4 Library
This one library has been a blocker in building rbx_binary for WebAssembly.

LouieK22:
closes #255

### DIFF
--- a/rbx_binary/src/chunk.rs
+++ b/rbx_binary/src/chunk.rs
@@ -41,7 +41,7 @@ impl Chunk {
                 zstd::bulk::decompress(&compressed_data, header.len as usize)?
             } else {
                 log::trace!("LZ4 compression");
-                lz4::block::decompress(&compressed_data, Some(header.len as i32))?
+                lz4_flex::block::decompress(&compressed_data, header.len as usize).unwrap()
             }
         };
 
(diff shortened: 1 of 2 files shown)


rbx_binary/src/chunk.rs
```diff
@@ -41,7 +41,7 @@ impl Chunk {
                 zstd::bulk::decompress(&compressed_data, header.len as usize)?
             } else {
                 log::trace!("LZ4 compression");
-                lz4::block::decompress(&compressed_data, Some(header.len as i32))?
+                lz4_flex::block::decompress(&compressed_data, header.len as usize).unwrap()
```
Dekkonot, line 44:
```suggestion
                lz4_flex::block::decompress(&compressed_data, header.len as usize).map(io::Error::other)
```

We probably don't want to unwrap in the off chance the decompression doesn't succeed. If nothing else, this should be `expect` instead of `unwrap`.

Dekkonot (approved):
This is something we should do since it blocks building to WASM, but I'm going to leave this open for a few days to give people who have opinions a chance to weigh in.

Dekkonot (approved):
Nobody spoke against it which means this is something we are doing

PR #526, opened 2025-05-03
Remove Database from `TypeInfos`, Remove Unwrap

krakow10:
This work is motivated by recognizing that ~two uses~ one use of unwrap() can be removed by using btree_map::Entry. TypeInfos is shrunk by 8 bytes with the duplicate database reference removed, hooray!


Dekkonot:
This doesn't seem to improve performance at all on my system, but I also had to make patches to it to make it actually run. Is this the code you're testing?

krakow10:
Ah, I built the two pair of commits separately, but never together.  The first commit showed the higher throughput, but the second two didn't seem to have an impact.  I ran another benchmark for this and got the weird 70% result so I posted it to https://github.com/rojo-rbx/rbx-dom/issues/529#issuecomment-2848889670

krakow10:
This used to include two BTreeMap::entry changes, but one has been obsoleted by other changes, so now is there just one.

krakow10:
fixed merge conflicts

krakow10:
Please review this as a code improvement (entry api) rather than a performance change.

PR #527, opened 2025-05-03
wip: Refactor DecodeOptions & EncodeOptions

krakow10:
Change `DecodeOptions` to an enum to more accurately model the data it represents.  When `NoReflection` is used, the database does not need to exist.

TODO: refactor `EncodeOptions` similarly

### DIFF
--- a/rbx_reflector/src/defaults.rs
+++ b/rbx_reflector/src/defaults.rs
@@ -17,9 +17,7 @@ pub fn apply_defaults(
 ) -> anyhow::Result<()> {
     let file = BufReader::new(File::open(defaults_place).context("Could not find defaults place")?);
 
-    let decode_options = rbx_xml::DecodeOptions::new()
-        .property_behavior(rbx_xml::DecodePropertyBehavior::IgnoreUnknown)
-        .reflection_database(database);
+    let decode_options = rbx_xml::DecodeOptions::IgnoreUnknown { database };
 
     let tree =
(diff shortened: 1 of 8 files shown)


Dekkonot:
We aren't interested in performing this refactor. It's a breaking change, and it doesn't add anything. This change would essentially just be noise.

I don't think it's fair to say that this more accurately models the data it represents: we may add more options in the future, and the database exists separately from how it's used.

krakow10:
Thanks for the clear feedback, I really appreciate that.

krakow10:
> the database exists separately from how it's used.

This is what I was trying to accomplish. It's _not used_ with the `NoReflection` setting, and an enum models that correctly by having it not exist.

Dekkonot:
Right but my point is that the property decode behavior is technically unrelated to the database. We could potentially do things with classes in the future, and we need it for decoding enums even when we don't use the database for property name resolution.

Even if we agree to disagree on the modeling part though, I think it's a bad idea to have a bunch of enum variants that essentially all contain the same information just from a UX standpoint. I don't believe it solves a real problem and just moves the UX issue somewhere else.

krakow10:
This was motivated by needing a string interner for some variants an not for others when using string internment.  For example, the `NoReflection` variant must always intern strings because it can't grab references to strings in the database, whereas `IgnoreUnknown` needs a database to determine what is known or not, but no string interner because unknown properties are not represented.

I think you are correct, the database should probably just be moved to the Deserializer like rbx_binary instead of sprinkling it into every variant here, and this PR can remain closed.

Issue #515, opened 2025-04-18
Ustr Concerns & Footguns


krakow10:
- Ustr will memory leak if used with input strings continuously
  - Ustr::from_existing can be used to actively mitigate this, but it's not obvious
  - Using a wrapper function with a static lifetime requirement can help prevent doing this accidentally
```rust
// disallow non-static lifetimes
fn static_ustr(s:&'static str)->Ustr{
	ustr(s)
}
```
- Ustr is never dropped even if the a Roblox file is only decoded once
  - Not a problem for me because I can do it ahead of time, but this put limits on how the library can be used

In my opinion the switch to Ustr is an overall benefit, but perhaps it can be even better with something else.

Ideas:
Perhaps use an owned dictionary. Could be a simple as `AHashSet<String>` or a newtype of it.  Decode and encode functions accept the dictionary as an argument (mutably for decode), but there are convenient versions which create their own local dictionary if advanced deduplication is not an issue.  This means that properties and names can all be &str referencing the dictionary, a total replacement for Ustr with more memory control.

kennethloeffler:
I'm not sure if there's a problem. It's true that entries in ustr's string cache are never freed, but rbx-dom only uses ustr for class and property names (a finite set of strings), and will not cause an unbounded memory leak.

> Ustr will memory leak if used with input strings continuously 

Are you creating `Ustr`s from an endless stream of unique strings? That's not the intended use of ustr. 

The idea behind ustr is that when there's a finite set of strings that can ever possibly appear in your program, you may create `Ustr`s from them. Then behind the scenes, the library only allocates if the string hasn't been cached yet, otherwise it uses the existing allocation from the cache. Note that creating `Ustr` from an already-cached string value will not allocate, so memory usage is bounded as long as the set of strings is finite.

This greatly improves the memory usage and runtime performance of rbx-dom because it avoids many, many allocations. We'd also be hard-pressed to beat ustr for our use case without just re-implementing it. Common workloads require *a lot* of lookups, assignments, and comparisons of class and property names, and `Ustr`s are very efficient for these because they have precomputed hashes and are compared and assigned by reference.

> Ustr is never dropped even if the a Roblox file is only decoded once 
Are you concerned about the additional space required for the string cache? The total size all of all property and class names  is about 56 kilobytes. In practice, the actual space required will be smaller than that since it's unlikely that a file contains every possible property and class name.

[...]

PR #528, opened 2025-05-03
Use Zero-Copy in Reflection Database

krakow10:
This replaces all occurences of `Cow<'a, str>` and `String` in rbx_reflection with `&'a str` and a `#[serde(borrow)]` annotation.  ~~The rbx_reflector implementation is rather lazy.~~  No performance impact after the database is initialized, but will have somewhat reduced memory usage for the in-memory database as well as reduced initialization time.  There could be little fix-ups I missed because of how coerce-able &str is.

Notes:
- I don't actually know when the `#[serde(borrow)]` annotation is required and when it's not
- `PropertyMigration` gains a wacky constructor for use in rbx_reflector
- Automatically loading the reflection database from a file uses a OnceLock to store the file data so that the deserialized reflection database can hold 'static references to the data.
- Lint check is failing due to #614


Dekkonot (commented):
I measured the impact by swapping out the global allocator to measure the memory used by the database, and it looks like this would save us 685kb on startup, which amounts to 24% of the memory footprint of the reflection database. At least on startup; no promises that none of the other code involved isn't causing any noise.

Unfortunately though I'm not sure if I can merge this right now. It's a _good_ change, but it's a breaking one for `rbx_reflection` and I'm not sure it's worth the performance gain. I'm not going to close this, but a breaking change for a 685kb change in memory feels not worth it. I doubt anyone is running rbx-dom in an environment where they need that memory back urgently.

In terms of implementation, this is mostly fine but I notice a trend with using `&Enum::Variant` when pattern matching. Presumably this is to avoid ownership trouble, but probably the right way to do this is either borrowing the value you're matching or using the `ref` keyword.

e.g. instead of 
```rust
match value {
  &Enum::Variant(value) => {}
}
```

you should probably be doing

```rust
match &value {
  Enum::Variant(value) => {}
}
```

or `Enum::Variant(ref value) => {}` if necessary.

krakow10:
> Unfortunately though I'm not sure if I can merge this right now.

That's fine, you're right that I'm not hurting for 1MB of memory here.  I put this together to implement the desired "lifetime borrowing" semantics described in https://github.com/rojo-rbx/rbx-dom/issues/515#issuecomment-2816354467 after I realized I had seen an implementation in another project.  If it can be eventually combined with other breaking changes to make a worthwhile upgrade then that's great.

> a trend with using &Enum::Variant

I had always just assumed it would generate the same assembly whether you dereference or bind by reference.  I was using the `&Enum::Variant(value) => {}` syntax simply because I prefer the look of deconstruction using `&` to the dereference operator `*`.  I haven't thought deeply about it, I just like the type to be in the correct form to be used directly after without dereference or coercion.  For example, without prefixing the match arm with `&`, the type of `alias_for` is `&&str`.  If there's a performance reason not to do this that I don't understand then I would love to learn about it.  I checked it out in Compiler Explorer and it seems like there is actually a difference: https://godbolt.org/z/5E8KxY1xh I don't really understand what the assembly difference means other than there is one extra instruction using `&`.

Dekkonot:
> Without prefixing the match arm with &, the type of alias_for is &&str. If there's a performance reason not to do this that I don't understand then I would love to learn about it. I checked it out in Compiler Explorer and it seems like there is actually a difference: https://godbolt.org/z/5E8KxY1xh I don't really understand what the assembly difference means other than there is one extra instruction using &.

I'll be honest: I don't know either. My preference is entirely just because I consider `&Enum::Variant(value) => value` to be worse than `Enum::Variant(value) => *value` visually. The fact that you're using `&` for dereferencing is horrifying to me, haha.

krakow10:
I updated the dereferencing deconstructions I found according to your preference.  By the way, what do you think about `Into<&'a str>`?  I actually specifically needed to do this instead of `AsRef<str>` for [hash_str](https://github.com/krakow10/hash_str/blob/982061ae3e709881dbe47ed25e7290bd65e1417d/src/cache.rs#L97) because AsRef has no lifetime information from the source type.  It looks like it's also applicable here (as in `AsRef<str>` will not work), although simply accepting `&'a str` would definitely make more sense.  Notably `&String` does not implement `Into<&'a str>` so it's instantly worse.

krakow10:
@Dekkonot While cutting a new release on my fork, I noticed that #376 is a breaking change and will require a new database release.  Perhaps this PR can be included in the next major database version?

Dekkonot:
I'm not opposed to that, though the breaking change in rbx-reflection-database doesn't necessarily correspond to one in rbx-reflection.

krakow10:
Rebased to fix the merge conflicts.
- ~~Note that because of the neat new code for loading the database from a file, the entire file must be memory leaked to get a 'static lifetime on the database!  See 5efd6a8e178562387e3db4934ffddad95a5efad7~~ Changed to OnceLock instead of Box::leak

krakow10:
I went to go make a companion pull request for Rojo but... it just compiles fine with these changes.  There are three spots that as_ref() calls can be removed, but that's all I could find: https://github.com/krakow10/rojo/commit/ea20aa4fe48aed3687bc0c8e609b1c737a9d95f6

krakow10:
An alternative implementation that would be more agreeable with loading the database from a file would be to use String or `Box<str>`. Either way, Cow should not be used in the reflection database.

krakow10:
This branch is now failing stylesheet tests, but it's not my fault! :point_right: #602

Edit:
Nevermind it is my fault, updating submodules is confusing, that's why the test was failing for me on master branch and here

krakow10:
I fixed the merge conflicts, but I can't figure out how to borrow correctly in the `PropertyMigration` Deserialize implementation! 51b5008e469467048d11ee038bbe95848f904499

Edit:
I fixed it but in a bad way.  I moved the is_empty check and the tests to rbx_reflector/patches.rs

Edit2:
I've reverted the bad commits and pushed a proper fix.

rbx_reflection_database/src/lib.rs
```diff
@@ -65,7 +65,11 @@ static LOCAL_DATABASE: LazyLock<ResultOption<ReflectionDatabase<'static>>> = Laz
         return Ok(None);
     };
     if path.exists() {
-        let database: ReflectionDatabase<'static> = rmp_serde::from_slice(&fs::read(path)?)?;
+        static DATABASE: std::sync::OnceLock<Box<[u8]>> = std::sync::OnceLock::new();
+        let database_file = fs::read(path)?.into_boxed_slice();
+        // The database is guaranteed to be set, though not guaranteed to be set by this thread.
+        let _ = DATABASE.set(database_file);
+        let database: ReflectionDatabase<'static> = rmp_serde::from_slice(DATABASE.get().unwrap())?;
```
Dekkonot: I assume this change is because we need the database to be `'static` to decode it, but I had to go and read your explanation in the PR to know that. It would be nice if there was a comment explaining it here.

rbx_reflector/src/patches.rs
```diff
@@ -239,22 +243,38 @@ struct Patch {
...
+    Enum(String),
+}
+impl<'db> From<&'db DataType> for rbx_reflection::DataType<'db> {
+    fn from(value: &'db DataType) -> Self {
+        match value {
+            DataType::Value(variant_type) => rbx_reflection::DataType::Value(*variant_type),
+            DataType::Enum(e) => rbx_reflection::DataType::Enum(e),
+        }
+    }
+}
```
Dekkonot, line 268: I admit, I am not entirely sure why this is required. I assume it has to do with lifetimes but a proper explanation would be appreciated.

`rbx_reflector` is one of the only places we dogfood `rbx_reflection` so if this is a result of some kind of implementation pain, we should probably fix it instead of lazily working around it.

Dekkonot: Okay I have tried to fix it locally and I understand. That's... frustrating. It makes me concerned about potential consumers of `rbx_reflection` if lifetimes start to cause problems like this.

Dekkonot (commented):
The usage of replacement structs in `rbx_reflector` makes me concerned about potential consumers of `rbx_reflection` once we do this, but it otherwise looks fine. I'd like an explanatory comment on the new static in `rbx_reflection_database` though, and it's probably worth discussing the lifetime problem a bit more.

rbx_reflector/src/patches.rs
krakow10: It may be possible to make the decode of the roblox json use more zero copy, the rbx_reflection types may get more reuse that way.  This is an implementation detail of rbx_reflector building the database though.

As for the users of rbx_reflection, the lifetimes are 'static when dealing with the bundled database.  If you manually deserialize a database from a file, zero-copy makes managing the lifetimes more complicated.  The OnceLock provides an example solution which turns lifetimes to 'static.

rbx_reflection_database/src/lib.rs
krakow10: Added a comment and changed OnceLock to use get_or_init.

rbx_reflector/src/patches.rs
Dekkonot: Hesitantly going to say 'ok' because I don't care much about `rbx_reflector`'s efficiency. Don't love it though.

Dekkonot (dismissed):

krakow10:
> and it's probably worth discussing the lifetime problem a bit more

The lifetimes are always static when dealing with rbx_reflection_database's bundled database.  If it was a problem, you would have already seen it with Cow, which exposes the same zero-copy style lifetimes in its type signature, it just never uses the Cow::Borrowed variant at runtime.

The case that becomes more complicated is manually building (ala rbx_reflector) or mutating a `ReflectionDatabase<'a>` structure with custom classes / properties at runtime.  This doesn't seem like an intended or supported use case of rbx_reflection.  I don't know why anyone would want to inherit Roblox's class mess via upstream rbx_reflection types instead of making their own custom mess.

Despite what I said about not being intended, custom reflection databases can still be accomplished after the zero-copy changes.

rbx_reflection/src/database.rs
```diff
@@ -16,18 +13,20 @@ use crate::{ClassTag, PropertyMigration, PropertyTag};
 /// and enums.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "PascalCase")]
+#[serde(bound(deserialize = "'a: 'de"))]
```
kennethloeffler: The only reason we need these bound attributes is because PropertyMigration's Deserialize impl is not general enough...

rbx_reflection/src/migration.rs
```diff
@@ -22,29 +22,50 @@ pub enum MigrationError {
...
+        };
+        Some(Self {
+            new_property_names,
+            migration,
+        })
+    }
+}
 
-impl<'de> Deserialize<'de> for PropertyMigration {
+impl<'de> Deserialize<'de> for PropertyMigration<'de> {
```
kennethloeffler: ...because here we say that the input data has exactly the same lifetime as the the references stored in PropertyMigration. See the serde docs: https://serde.rs/lifetimes.html#the-deserializede-lifetime. The constraint we want instead is that 'de lives at least as long as 'a:
```suggestion
impl<'de: 'a, 'a> Deserialize<'de> for PropertyMigration<'a> {
```
Then we don't need the additional bound attributes.

kennethloeffler (commented):
Some comments on serde lifetimes:

rbx_reflection/src/database.rs
```diff
@@ -168,9 +169,9 @@ pub struct PropertyDescriptor<'a> {
 
 impl<'a> PropertyDescriptor<'a> {
     /// Creates a new `PropertyDescriptor` with the given name and type.
-    pub fn new<S: Into<Cow<'a, str>>>(name: S, data_type: DataType<'a>) -> Self {
+    pub fn new(name: &'a str, data_type: DataType<'a>) -> Self {
```
kennethloeffler: Would prefer to continue using a trait bound here to avoid altering this method's contract too much:

```suggestion
    pub fn new<S: Into<&'a str>>(name: S, data_type: DataType<'a>) -> Self {
```

kennethloeffler (commented):
With respect to #[serde(borrow)], serde only implicitly borrows fields that are directly &str or &[u8]: https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl. If we want to make sure that we're borrowing as much as possible, we should use the attribute on any fields containing borrows that are different than those types.

krakow10:
> If we want to make sure that we're borrowing as much as possible, we should use the attribute on any fields containing borrows that are different than those types.

Shouldn't it fail to compile if it doesn't borrow?  What would a deserialized reference be referencing if not the source data?  This makes me think we don't need to eagerly annotate everything with #[serde(borrow)], and can go with minimum annotation.

rbx_reflection/src/migration.rs
krakow10: Thanks, I didn't realize this.

kennethloeffler:
Yes that's correct, I misinterpreted the serde documentation, sorry! 

In the case of the `ReflectionDatabase` struct:
```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "PascalCase")]
#[non_exhaustive]
pub struct ReflectionDatabase<'a> {
    /// The Roblox release that this reflection database was generated from.
    pub version: [u32; 4],

    /// All of the the known classes in the database.
    #[serde(serialize_with = "crate::serde_util::ordered_map")]
    #[serde(borrow)]
    pub classes: HashMap<&'a str, ClassDescriptor<'a>>,

    /// All of the known enums in the database.
    #[serde(default, serialize_with = "crate::serde_util::ordered_map")]
    pub enums: HashMap<&'a str, EnumDescriptor<'a>>,
}
```
We only need one `#[serde(borrow)]` on either the `classes` field or the `enums` field, because the presence on either one is enough for serde to generate the correct`'de: 'a` bound for the entire struct. 

In another example, `PropertyDescriptor`:
```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "PascalCase")]
#[non_exhaustive]
pub struct PropertyDescriptor<'a> {
    /// The name of the property, like "Position" or "heat_xml".
    pub name: &'a str,

    /// The maximum access to this property available to Lua.
    pub scriptability: Scriptability,

    /// The type of the value described by this descriptor.
    pub data_type: DataType<'a>,

    /// A set of the tags that apply to this property.
    #[serde(serialize_with = "crate::serde_util::ordered_set")]
    pub tags: HashSet<PropertyTag>,

    /// The kind of property this is, including whether it is canonical.
    pub kind: PropertyKind<'a>,
}
```
We don't need any `#[serde(borrow)]`, because serde has already inferred the intended bound because of the `&'a str` field. 

So the rule seems to be that `#[serde(borrow)]` is only required on at least one field when there are no `&str` or `&[u8]` fields.

There's some argument to be made that annotating any field that is not `&str` or `&[u8]` would better document the code (because semantically, these fields do borrow from the input), but I'm fine leaving it as-is.

krakow10:
I've added a `#[serde(borrow)]` to the enums field.  Alternatively `#[serde(bound(deserialize = "'de: 'a"))]` can be added to the ReflectionDatabase struct.

rbx_reflector/src/patches.rs
kennethloeffler: I don't love the code duplication, and I'm wondering if there's a way we could just start borrowing from YAML patch contents to avoid needing to define separate "owned" types. I'll quickly cook up a PR against this branch to do it because it should be pretty mechanical, just have to find a serde parser than can do zero copy (serde_yaml can't).

krakow10: It's an implementation detail of rbx_reflector, so it doesn't involve breaking changes.  You could merge first and make it into a separate PR.  I can help review it either way.

kennethloeffler (approved):

PR #530, opened 2025-05-05
Zero-Copy ReadSlice Trait

krakow10:
The point of these changes is to reduce the number of copies where possible. To achieve this, we split off chunks of a byte slice instead of allocating a String or Vec<u8> where applicable.

### Performance
5-8% faster on the deserialize microbenchmarks according to my testing.  This mostly affects performance on a per-prop-chunk basis rather than per-property which explains why it doesn't seem to have a big impact on the Miner's Haven benchmark.  This is probably also why #545 didn't see a big performance uplift.


rbx_binary/src/core.rs
```diff
@@ -315,6 +315,48 @@ pub trait RbxWriteExt: Write {
...
+/// makes a semantic differentiation between strings and binary buffers,
+/// which makes it more strict than Roblox but more likely to be correct.
+///
+/// Unlike RbxReadExt::read_binary_string, this function only operates on a byte slice.
+pub fn read_binary_string_slice<'a>(slice: &mut &'a [u8]) -> io::Result<&'a [u8]> {
+    let length = slice.read_le_u32()?;
+
+    let out;
+    // split_at can panic if the slice is shorter than length
+    // but we do not expect that to happen.
```
Dekkonot, line 331: Do you think it's worth checking? We could produce a meaningful error if this invariant was violated, which might help consumers.

Then again, well-made producers of rbxm files will never have this problem so it may be a moot point; if someone is hand-crafting binary files they should be careful.

krakow10: I don't think it's worth it.  Seems like the kind of thing that would be quickly found and fixed in any encoder implementation.  It can always be added later if the situation calls for it, like perhaps if some legacy tool that will never be fixed is creating files malformed in this specific way.  It'd be simple to call `.split_at_checked` and `.ok_or` a `UnexpectedEof` io error, but not sure how that would affect performance (probably negligable).  Let me know if you want it.

rbx_binary/src/core.rs
```diff
@@ -315,6 +315,48 @@ pub trait RbxWriteExt: Write {
...
+    (out, *slice) = slice.split_at(length as usize);
+
+    let Ok(str) = core::str::from_utf8(out) else {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            "stream did not contain valid UTF-8",
+        ));
+    };
+
+    Ok(str)
```
Dekkonot, line 357:
```suggestion
    core::str::from_utf8(out).map_err(|_| {
        io::Error::new(
            io::ErrorKind::InvalidData,
            "stream did not contain valid UTF-8",
        )
    })
```

No need for the explicit return here; we can just use `map_err`.

Dekkonot (changes requested):
I'm a clever boy and can work out that the new functions aren't a part of `RbxReadExt` because they operate explicitly on a slice instead of a nebulous `impl Read`, but it might be worth writing it down for future generations. I'll leave that to your discretion, but I tend to err on the side of writing things down even if they feel obvious because it lowers the bus factor for the project.

Otherwise this seems fine to me.

krakow10:
> `RbxReadExt` ... might be worth writing it down

Done.  I was considering a trait too, but decided this was simpler.

krakow10:
This now combines changes from #561 to create a RbxReadZeroCopy trait.

Dekkonot:
> I would like to open stacked PRs, but they need to be created with a branch within the repository itself, which I don't have access to.

If you are too keen, you can make PRs on your own fork of rbx-dom and link them. I'd probably just merge them with a squash though so you can theoretically rebase after they're merged.

krakow10:
I've split out the RbxReadInterleaved change into its own PR.  This is just the ReadSlice trait now.

kennethloeffler (dismissed):
Nice, this looks to be a significant optimization for light workloads: I saw about +10% on the small benchmarks. It makes sense, since we have proportionally more strings to copy in larger files. Code looks great to me, nice work!

krakow10:
It's good to hear that I'm not just on a wild goose chase trying to find improvements to this PR, that the end result is actually desirable.  Thanks for the feedback.

krakow10:
Fixed merge conflicts.

PR #532, opened 2025-05-20
Decompress All Chunks Before Deserializing

krakow10:
This is required for #533.  In order for the dom to reference decompressed data, it must be available from a let binding.  The `DecompressedFile` struct name is a placeholder, so please recommend a more sensible name.  This has no significant performance impact as far as I can tell.  This also opens the potential for parallel chunk decompression.  However, decompression is only around 8% of the workload most of the time, so the performance improvement potential is limited.

### DIFF
--- a/rbx_binary/src/chunk.rs
+++ b/rbx_binary/src/chunk.rs
@@ -55,6 +55,33 @@ impl Chunk {
     }
 }
 
+/// Decompressed chunks.  Referenced by WeakDom properties
+/// when the identity string interner is used.
+pub struct Chunks {
+    chunks: Vec<Chunk>,
+}
+impl Chunks {
+    pub fn decode<R: Read>(mut reader: R) -> io::Result<Chunks> {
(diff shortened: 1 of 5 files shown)


Dekkonot:
It decoded all the chunks in parallel and after it was done unified the data by creating Instances, inserted properties for them, and processing parents. There are some performance optimizations we could benefit from in rbx-binary (in particular, being able to pre-allocate the number of properties for every Instance) but the actual decoding process it turns out does not take very much time. Unification is the bulk of the time spent but you can get most of the benefits of parallelization by just pre-allocating and trying to cut down on wasted memory.

It makes sense when you think about it: we are blocked by the allocator when deserializing. The allocator has to synchronize when allocating anyway, so we lose a lot of the performance gain. On Windows, at least. No promises that MacOS or Linux don't have better global allocators.

We could potentially make up for that by using arenas that were pre-allocated, but that has the potential to explode the memory usage of rbx-binary for unknown gains.

Issue #92, opened 2020-01-17
rbx_dom_strong


LPGhatguy:
Right now, we have a _weakly-typed_ DOM implementation. It'd be cool to have a strongly-typed one, too.

krakow10:
How do you imagine this working? Is there any proof of concept? I wrote a small demo that implements inheritance using the Deref trait, and StrongInstance is an enum of Boxed strong types:
https://github.com/rojo-rbx/rbx-dom/compare/5d67bb08be7a94da97a68b86e2b22f8a75b0afcd...krakow10:rbx-dom:strong

However it's implemented, it probably should be done by invoking codegen using the reflection database into intermediate committed code, or using a build.rs script with no intermediate committed code.

Another point is that the dom code could possibly be shared between strong and weak, since the surface area of intersection with the weak Instance type is not intractably large as seen in #488.

rbx_binary and rbx_xml working for both strong and weak would be much more work, especially to make it properly generic over the Serialize + Deserialize trait which would also allow deserialization into a custom user dom type.  Prior art similar to this is some work I did on the [vbsp crate](https://github.com/icewind1991/vbsp/commit/265cf21caf31631ba8f82ab868c8d7ae1e05d15f) which made Valve entity deserialization into [types unique to the game](https://github.com/krakow10/vbsp-entities) (tf2, css) possible.

krakow10:
I wrote some codegen to generate types from the database:
https://github.com/rojo-rbx/rbx-dom/compare/7ce28d972e275153472f96080bafa806cab08ba5...krakow10:rbx-dom:strong

PR #533, opened 2025-05-20
Drop Ustr Dependency

### LINKED ISSUE #515: Ustr Concerns & Footguns

krakow10:
- Depends on #528 to avoid Cow AsRef using 'de lifetime instead of 'db
- Depends on #530 to hold &str references of chunk data
- Depends on  #532 to prevent dropping decompressed chunks while holding references

Closes #515.

Here's a draft of what it could look like to drop the ustr sledgehammer in favour of fine-grained memory control. This addresses all my concerns in #515.  There is only enough code here to run rbx_binary benches, but the rest of the code for rbx_xml, rbx_util, rbx_reflector, etc. can be adapted from my raw-str branch https://github.com/krakow10/rbx-dom/compare/1482a07ec4d6184fee6769f62745fdcc177c6d49...raw-str, which is where most of this code came from as well.

Here are just the commits on top of the prerequisite dependencies:
branch https://github.com/krakow10/rbx-dom/compare/prereq...lifetimes

Overview:
- `Ustr` is completely dropped in favour of `&'a str` for properties and class names
- Introduce lifetimes to `WeakDom` and `Instance` for properties
- Introduce `DecodeOptions` to rbx_binary with an optional string interner

Performance:
Deserialization is within 2%, but Serialization performance seems to be significantly (40%) slower, which I don't understand why.  Most of the code affects deserialization only.  I suspect it could have to do with using `HashSet<&str>` instead of `UstrSet`.

There are three supported use cases:
1. Ignore or deny invalid properties, get `WeakDom<'static>`:
```Rust
let dom = rbx_binary::from_reader(reader, DecodeOptions::ignore_unknown())?;
let dom = rbx_binary::from_reader(reader, DecodeOptions::error_on_unknown())?;
```
2. Allow invalid properties by referencing `DecompressedFile` binding, required to outlive `WeakDom<'dom>`:
```Rust
let file = DecompressedFile::from_reader(reader)?;
let dom = file.deserialize(DecodeOptions::read_unknown())?;
```
3. Allow invalid properties using a custom string interner, enabling the DecompressedFile data to be immediately dropped to save memory:
```Rust
let host = bumpalo::Bump::new();
let mut cache = HashSet::new();
let dom = rbx_binary::from_reader(reader, DecodeOptions::read_unknown_with(|str: &str|{
	match cache.get(str) {
		Some(interned) => interned,
		None => {
			let interned = host.alloc_str(str) as &str;
			cache.insert(interned);
			interned
		}
	}
}))?;
```

### DIFF
--- a/rbx_binary/benches/suite/util.rs
+++ b/rbx_binary/benches/suite/util.rs
@@ -1,4 +1,5 @@
 use criterion::{measurement::Measurement, BatchSize, BenchmarkGroup, Throughput};
+use rbx_binary::{DecodeOptions, DecompressedFile};
 
 pub(crate) fn bench<T: Measurement>(group: &mut BenchmarkGroup<T>, bench_file: &'static [u8]) {
     serialize_bench(group, bench_file);
@@ -6,7 +7,8 @@ pub(crate) fn bench<T: Measurement>(group: &mut BenchmarkGroup<T>, bench_file: &
 }
 
 fn serialize_bench<T: Measurement>(group: &mut BenchmarkGroup<T>, buffer: &[u8]) {
-    let tree = rbx_binary::from_reader(buffer).unwrap();
+    let file = DecompressedFile::from_reader(buffer).unwrap();
+    let tree = file.deserialize(DecodeOptions::read_unknown()).unwrap();
     let root_ref = tree.root_ref();
     let mut buffer = Vec::new();
 
@@ -36,7 +38,10 @@ fn deserialize_bench<T: Measurement>(group: &mut BenchmarkGroup<T>, buffer: &[u8
         .throughput(Throughput::Bytes(buffer.len() as u64))
         .bench_function("Deserialize", |bencher| {
             bencher.iter(|| {
-                rbx_binary::from_reader(buffer).unwrap();
+                DecompressedFile::from_reader(buffer)
+                    .unwrap()
+                    .deserialize(DecodeOptions::read_unknown())
+                    .unwrap();
             });
         });
 }
--- a/rbx_binary/src/chunk.rs
+++ b/rbx_binary/src/chunk.rs
@@ -55,6 +55,33 @@ impl Chunk {
     }
 }
 
+/// Decompressed chunks.  Referenced by WeakDom properties
+/// when the identity string interner is used.
+pub struct Chunks {
+    chunks: Vec<Chunk>,
+}
+impl Chunks {
+    pub fn decode<R: Read>(mut reader: R) -> io::Result<Chunks> {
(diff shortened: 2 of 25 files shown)


Dekkonot:
I haven't commented on this because I was curious what the results would be. I felt it unlikely that it would improve things much, but I as curious.

Given that the results are unimpressive or just downright bad, I am going to put a pin in this and say thank you for the exploration but we aren't interested in doing this. We just adopted Ustr, and it's a much better solution than attaching lifetimes to everything just from a DX standpoint.

The memory footgun you ran into with the string internment is something we can work to make more obvious or safer (I would be open to a method for `Instance` that tried to get a property with a string but did not allocate any new `Ustr`s). But we are not dropping Ustr.

krakow10:
> Given that the results are unimpressive or just downright bad

Is this referring to the serialization performance dropping or the developer experience when managing lifetimes?  I assume the latter.  I figured as much and was prepared for this to be rejected, but I thought it was worth a shot with how far it's come.  Thanks for taking a look.

Dekkonot:
> Is this referring to the serialization performance dropping or the developer experience when managing lifetimes?

Both! If it doesn't make a difference for deserializing, make serialization performance worse, and makes the DX worse I think it's obvious that we can't do it.

krakow10:
I expect that the performance issue can be addressed, but exposing lifetimes seems like a dealbreaker which is why I didn't look into it.  The next best thing to dropping ustr is replacing rbx_dom_weak with rbx_dom_strong, and I've already put in some work there #92 to generate the types.  I can't seem to get rbx-dom out of my head, so see you around!

krakow10:
I've been considering whether to put more effort into this, but I think my effort is better spent on #92.  Closing in favour of working on that.

PR #534, opened 2025-05-22
Avoid Hashtable Clone in `clone_ref_as_builder`

krakow10:
This clones each element in the iterator rather than cloning the entire hashtable and then consuming it into an iterator.  Probably has a negligible performance impact, but it avoids an intermediate allocation.  Also avoids cloning the name string twice since the definition for String::from(str) in the rust standard library is:
```rust
impl From<&str> for String {
    fn from(s: &str) -> String {
        s.to_owned()
    }
}
```


krakow10:
My assumptions here are incorrect.  I read the Clone implementation for hashbrown HashMap again and it's obviously copying the entire structure verbatim without rehashing, I don't know how I interpreted it in any other way.

PR #537, opened 2025-05-22
Use extend in `generate_referents`

krakow10:
Use a primitive cast and refactor the loop into an iterator.  I kept the panic condition that was being checked for every sequential index, except factored out of the loop.  This is somehow 4% faster on "Miner's Haven/Serialize"?  Everything helps I guess.


krakow10:
`assert!(self.relevant_instances.len() <= i32::MAX as usize);`

This check is off by one, since the maximum number generated by .enumerate() will be
`relevant_instances.len() - 1` so maybe it should be written as:
```rust
assert!(self.relevant_instances.len() <= i32::MAX as usize + 1);
```
to avoid underflowing with 0 - 1 on the LHS.  Then again, the list is unlikely to get this long in the first place.

Dekkonot (approved):

PR #538, opened 2025-05-24
Add support for StyleRule `PropertiesSerialize`

cameronpcampbell:
NOTE: This PR has a companion PR in the rojo repo: https://github.com/rojo-rbx/rojo/pull/1058

- - -

This pull requests adds support for StyleRule's `PropertiesSerialize` property.

## Changes
- As `PropertiesSerialize` is serialized the same as `AttributesSerialize` I've renamed the `Attributes` type in `rbx_types` to `SerializedMap` as it will no longer be used for just Attributes. The other crates, packages, `database.json` and `database.msgpack` have also been updated to reflect this change.

- Added support for `PropertiesSerialize` in `customProperties.lua`.


kennethloeffler:
> As PropertiesSerialize is serialized the same as AttributesSerialize I've renamed the Attributes type in rbx_types to SerializedMap as it will no longer be used for just Attributes. The other crates, packages, database.json and database.msgpack have also been updated to reflect this change.

I don't see any point to doing this - this PR is essentially a straightforward property patch add, but by doing this it's turned into a massive change set that includes a breaking change. For what purpose? I think Attributes remains a fine name.

PR #540, opened 2025-05-29
rbx_reflection: Add `DataType.ty()`

krakow10:
All call sites become simplified with this conversion function.  The non-exhaustive cases are no longer applicable because the matching is handled from within the same crate as `DataType` is defined.


krakow10:
Added changelog entry and rebased.

PR #545, opened 2025-06-01
rbx_binary: Interleaved Iterators

krakow10:
- Supersedes #541

This makes the interleaved bytes readers and writers work directly with iterators.  Many intermediate allocations and copies are omitted, and much of the serialization and deserialization code dealing with interleaved bytes is greatly simplified.  I'm shocked to see a lack of performance uplift, there must be something weird going on.  All the benchmarks are ±3%, except the unreliable benchmark.

More effort could be put into avoiding intermediate allocations in the `Content` serialize, deserialize, and text_deserializer.

### DIFF
--- a/rbx_binary/src/core.rs
+++ b/rbx_binary/src/core.rs
@@ -137,7 +137,10 @@ pub trait RbxReadExt: Read {
     /// Creates an iterator of `len` big-endian i32 values.
     /// The bytes are read into a buffer immediately,
     /// and the values are transformed during iteration.
-    fn read_interleaved_i32_array(&mut self, len: usize) -> io::Result<impl Iterator<Item = i32>> {
+    fn read_interleaved_i32_array(
+        &mut self,
+        len: usize,
+    ) -> io::Result<impl Iterator<Item = i32> + use<Self>> {
         Ok(self
             .read_interleaved_bytes(len)?
             .map(|out| untransform_i32(i32::from_be_bytes(out))))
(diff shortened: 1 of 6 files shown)


Dekkonot:
> Edit: Not sure why this is failing to compile on the CI. It builds fine and passes all tests on rustc 1.87. Looks like it has to do with the syntax in https://github.com/rojo-rbx/rbx-dom/commit/367999315ff7857f866cbf2eda9047b3c6bde7f6

You're using Nightly. That might explain your performance woes too.

krakow10:
I'm very certain I'm not using nightly.
<details>

<summary>Proof</summary>

```
quat@quat-desktop ~ [127]> rustc --version
rustc 1.87.0 (17067e9ac 2025-05-09) (Arch Linux rust 1:1.87.0-2)
```

```rust
fn main(){
	dbg!(rustc_version::version_meta());
}
```

```
[src/main.rs:2:2] rustc_version::version_meta() = Ok(
    VersionMeta {
        semver: Version {
            major: 1,
            minor: 87,
            patch: 0,
        },
        commit_hash: Some(
            "17067e9ac6d7ecb70e50f92c1944e545188d2359",
        ),
        commit_date: Some(
            "2025-05-09",
        ),
        build_date: None,
        channel: Stable,
        host: "x86_64-unknown-linux-gnu",
        short_version_string: "rustc 1.87.0 (17067e9ac 2025-05-09) (Arch Linux rust 1:1.87.0-2)",
        llvm_version: Some(
            LlvmVersion {
                major: 20,
                minor: 1,
            },
        ),
    },
)
```

</details>

Installing 1.86 with rustup and building with that shows the same errors, so it must be something brand new in 1.87.  One way to express this in earlier versions would be to spell out the exact type of the returned iterator.

rbx_binary/src/core.rs
```diff
@@ -99,89 +123,79 @@ pub trait RbxReadExt: Read {
...
+        Ok(it)
+    }
+
+    /// Creates an iterator of `len` big-endian i32 values.
+    /// The bytes are read into a buffer immediately,
+    /// and the values are transformed during iteration.
+    fn read_interleaved_i32_array(
+        &mut self,
+        len: usize,
+    ) -> io::Result<impl Iterator<Item = i32> + use<Self>> {
```
Dekkonot: I don't know what `+ use<Self>` does here! As far as I can tell, `Self` isn't captured by the iterator type so it shouldn't be needed.

rbx_binary/src/core.rs
```diff
@@ -1,17 +1,41 @@
...
 pub static FILE_MAGIC_HEADER: &[u8] = b"<roblox!";
 pub static FILE_SIGNATURE: &[u8] = b"\x89\xff\x0d\x0a\x1a\x0a";
 pub const FILE_VERSION: u16 = 0;
 
+pub struct ReadInterleavedBufferIter<const N: usize> {
+    buffer: Vec<u8>,
+    index: usize,
+    len: usize,
+}
+impl<const N: usize> ReadInterleavedBufferIter<N> {
```
Dekkonot:
```suggestion
}

impl<const N: usize> ReadInterleavedBufferIter<N> {
```

Would rather we have new lines between blocks like this.

Dekkonot (changes requested):
Very minor nitpick for formatting and a question about type signatures. Otherwise looks good.

I expect the reason you aren't seeing a huge performance boost is that LLVM was already doing this optimization, just behind the scenes. I don't know for certain but a lot of these are allocations that are immediately consumed into iterators so it should be relatively trivial to optimize as a compiler.

I'm not too concerned about performance though. I'm most interested in the fact that it simplifies the type serde a fair bit.

rbx_binary/src/core.rs
krakow10: Writing the type using 'impl' implicitly captures the &mut self lifetime.  Here is the relevant documentation: https://doc.rust-lang.org/stable/edition-guide/rust-2024/rpit-lifetime-capture.html

krakow10:
Rebased and updated for latest code changes from the master branch.

Dekkonot (approved):

PR #546, opened 2025-06-02
Convert Lister To Use Iterators

krakow10:
Not the most important part of rbx-dom, but hey it's -38 loc.   The only functional difference should be that the faces are printed in all caps.  Updating bitflags shouldn't be a breaking change because it's a private type in a private field.


Dekkonot:
I'm not actually convinced this provides any value. Technically it improves Lister... but as you said, that's used for what, the Debug implementation for Faces and Axes?

It doesn't provide enough to warrant changing the debug output for Faces IMO, but I'll invite @kennethloeffler to share his thoughts.

Dekkonot:
It's been several months and my opinion on this hasn't changed. I don't really see the value. Lines of code don't really matter to me.

PR #554, opened 2025-09-21
Migrate Content properties added since v684

Dekkonot:
- Migrates `BackpackItem.TextureId` to `TextureContent`
- Migrates `ClickDetector.CursorIcon` to `CursorIconContent`
- Migrates `DragDetector.ActivatedCursorIcon` to `ActivatedCursorIconContent`
- Migrates `Decal.ColorMap` to `ColorMapContent` (just in case, as they don't serialize but Roblox could change that)
- Migrates `Mouse.Icon` to `IconContent`
- Migrates `UIDragDetector.ActivatedCursorIcon` to `ActivatedCursorIconContent`
- Migrates `UIDragDetector.CursorIcon` to `CursorIconContent`
- Migrates `UserInputService.MouseIcon` to `MouseIconContent`

### DIFF
--- a/patches/audio-player.yml
+++ b/patches/audio-player.yml
@@ -1,0 +1,7 @@
+Change:
+  AudioPlayer:
+    Asset:
+      Serialization:
+        Type: Migrate
+        To: AudioContent
+        Migration: ContentIdToContent
--- a/patches/backpack-item.yml
+++ b/patches/backpack-item.yml
@@ -1,0 +1,7 @@
+Change:
+  BackpackItem:
+    TextureId:
+      Serialization:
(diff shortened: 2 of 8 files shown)


Dekkonot:
Done, though I'm not sure if we can actually prove that those are migrations yet so we should keep an eye on it just to make sure.

Dekkonot:
> What I mainly want to avoid here is the possibility of having a mixture of old and new properties in the same dom, since that breaks the rbxm format.

Same, which is why I'm okay with doing it pre-emptively. I just haven't verified that the migrations in question are actually valid, is all.

PR #556, opened 2025-10-06
Update implementation of `PhysicalProperties` to include new `AcousticAbsorption` field

Dekkonot:
Roblox has added a new field to `CustomPhysicalProperties` called `AcousticAbsorption`. This PR adds support for that field in rbx_types, **which is breaking**.

It also implements support for reading and writing this new field to rbx_xml and rbx_binary. This is not breaking.

I have confirmed with two separate Roblox engineers working on the acoustics system that the default for `AcousticAbsorption` when reading an existing `CustomPhysicalProperties` is `1.0`, so it's what our implementation goes with rather than making the field optional.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -586,7 +586,7 @@ Two encoded `Rect` values with values `Rect.new(-1, -10, 8, 9)` and `Rect.new(0,
 ### PhysicalProperties
 **Type ID `0x19`**
 
-The `PhysicalProperties` type contains a bitfield which indicates whether the value is custom, along with whether it has an `AudioAbsorption` field. This flag is a single byte. These bits are described below, where bit `0` is the least-significant bit.
+The `PhysicalProperties` type contains a bitfield which indicates whether the value is custom, along with whether it has an `AudioAbsorption` field. This flag is a single byte. These bits are described below, where bit `0` is the least-significant bit. When there are multiple `PhysicalProperties` present, they are stored in sequence with no transformations or interleaving.
 
 | Bit Number | Purpose                                                                    |
 |:-----------|:---------------------------------------------------------------------------|
@@ -606,9 +606,9 @@ If bit `0` is set, then the value is followed by a `CustomPhysicalProperties`. T
 
 The `AcousticAbsorption` field is only present if bit `1` is set. Otherwise, it is left out and may be assumed to be `1.0` when deserializing a `CustomPhysicalProperties`.
 
-If bit `0` is not set, but bit `1` is set, the is no `CustomPhysicalProperties` present.
+If bit `0` is not set, but bit `1` is set, there is no `CustomPhysicalProperties` present.
 
-The `PhysicalProperties` type is stored as a `u8` representing the bitfield, followed by a `CustomPhysicalProperties` if bit `0` is set. When there are multiple `PhysicalProperties` present, they are stored in sequence with no transformations or interleaving.
+The `PhysicalProperties` type is stored as a `u8` representing the bitfield, followed by a `CustomPhysicalProperties` if bit `0` is set.
 
 
 If you had 4 physical properties with the following characteristics:
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -586,19 +586,43 @@ Two encoded `Rect` values with values `Rect.new(-1, -10, 8, 9)` and `Rect.new(0,
 ### PhysicalProperties
 **Type ID `0x19`**
 
-The `PhysicalProperties` type contains a flag which may be followed by a `CustomPhysicalProperties` value. `CustomPhysicalProperties` is a struct composed of five `f32` values:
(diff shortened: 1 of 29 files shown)


docs/binary.md
```diff
@@ -586,19 +586,43 @@ Two encoded `Rect` values with values `Rect.new(-1, -10, 8, 9)` and `Rect.new(0,
...
+| Density            | `f32`  | The density set for the custom physical properties             |
+| Friction           | `f32`  | The friction set for the custom physical properties            |
+| Elasticity         | `f32`  | The elasticity set for the custom physical properties          |
+| FrictionWeight     | `f32`  | The friction weight set for the custom physical properties     |
+| ElasticityWeight   | `f32`  | The elasticity weight set for the custom physical properties   |
+| AcousticAbsorption | `f32`  | The acoustic absorption set for the custom physical properties |
+
+The `AcousticAbsorption` field is only present if bit `1` is set. Otherwise, it is left out and may be assumed to be `1.0` when deserializing a `CustomPhysicalProperties`.
+
+If bit `0` is not set, but bit `1` is set, the is no `CustomPhysicalProperties` present.
```
kennethloeffler, line 609:
```suggestion
If bit `0` is not set, but bit `1` is set, there is no `CustomPhysicalProperties` present.
```

docs/binary.md
```diff
@@ -586,19 +586,43 @@ Two encoded `Rect` values with values `Rect.new(-1, -10, 8, 9)` and `Rect.new(0,
...
+
+The `PhysicalProperties` type is stored as a `u8` representing the bitfield, followed by a `CustomPhysicalProperties` if bit `0` is set. When there are multiple `PhysicalProperties` present, they are stored in sequence with no transformations or interleaving.
+
```
kennethloeffler, line 612: This passage reads a bit awkwardly here and has redundant information (we already know the bitfield is 1 byte wide, and how it can be followed by a `CustomPhysicalProperties`). 

I think we can move the "stored in sequence with no transformations or interleaving" bit to the opening passage and nix the rest.

rbx_binary/src/deserializer/state.rs
```diff
@@ -1174,18 +1174,30 @@ rbx-dom may require changes to fully support this property. Please open an issue
...
-                                    elasticity: chunk.read_le_f32()?,
-                                    friction_weight: chunk.read_le_f32()?,
-                                    elasticity_weight: chunk.read_le_f32()?,
-                                },
-                            ))
-                        } else {
-                            Variant::PhysicalProperties(PhysicalProperties::Default)
+                        let discriminator = chunk.read_u8()?;
+                        let value = match discriminator {
+                            0b00 | 0b10 => Variant::PhysicalProperties(PhysicalProperties::Default),
```
kennethloeffler, line 1179: Does this match Roblox's implementation (i.e. both the custom physical properties and acoustic absorption fields must be set to deserialize a custom acoustic absorption)?

rbx_binary/src/serializer/state.rs
```diff
@@ -1077,14 +1077,15 @@ impl<'dom, 'db, W: Write> SerializerState<'dom, 'db, W> {
...
+                                    chunk.write_u8(0b11)?;
+                                    chunk.write_le_f32(props.density())?;
+                                    chunk.write_le_f32(props.friction())?;
+                                    chunk.write_le_f32(props.elasticity())?;
+                                    chunk.write_le_f32(props.friction_weight())?;
+                                    chunk.write_le_f32(props.elasticity_weight())?;
+                                    chunk.write_le_f32(props.acoustic_absorption())?;
                                 } else {
-                                    chunk.write_u8(0)?;
+                                    chunk.write_u8(0b10)?;
```
kennethloeffler, line 1088: Don't love how we'll end up modifying preexisting PhysicalProperties -- maybe we could write `0b00` if there are no custom properties, `0b01` if there are custom properties and acoustic absorption == 1, and `0b11` if there are custom properties and acoustic absorption ~= 1?

rbx_dom_lua/src/EncodedValue.lua
```diff
@@ -378,13 +378,26 @@ types = {
...
 			else
-				return PhysicalProperties.new(
-					pod.density,
-					pod.friction,
-					pod.elasticity,
-					pod.frictionWeight,
-					pod.elasticityWeight
-				)
+				-- Passing `nil` instead of not passing anything gives
+				-- different results, so we have to branch here.
```
kennethloeffler, line 382: Thanks, I hate it

kennethloeffler (commented):
Implementation looks good, a few minor comments:

rbx_binary/src/deserializer/state.rs
Dekkonot: Yeah, Roblox is essentially using the acoustic absorption field as a "version" field to indicate whether a value supports `AcousticAbsorption` at all. It's why it's set even for new "not custom" values.

rbx_dom_lua/src/EncodedValue.lua
Dekkonot: Turns out `nil` becomes `0` when cast to a number. Love Roblox.

PR #561, opened 2025-11-08
rbx_binary: Eliminate Buffer From Interleaved Iterators

krakow10:
I missed a pretty obvious copy optimization here.  By constructing the iterator using a slice, we can use the original buffer and avoid copying the data into an intermediate buffer.  I was inspired by looking over #530.  I'm not actually seeing performance improvements in my benchmarking, even when omitting the interleaved index bounds checks with unsafe.  Do interleaved byte reads actually only occupy a very minor portion of the decode process?

~~Note: The function signature is infallible, but in fact it is possible for the provided slice to be too short, and split_at will panic, similarly to #530.  This could be changed to use split_at_checked and generate an EOF io error.~~

### DIFF
--- a/rbx_binary/src/core.rs
+++ b/rbx_binary/src/core.rs
@@ -269,43 +269,14 @@ impl ChunkBuilder {
 }
 
 impl<W> RbxWriteExt for W where W: Write {}
-pub trait RbxReadZeroCopy<'a> {
-    fn read_interleaved_bytes<const N: usize>(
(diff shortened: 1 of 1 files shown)


krakow10:
I've changed to a trait-based implementation.  Even though it's a bit more code, I think it's nicer, and it reduces the diff since it looks the same as the original RbxReadExt functions that it's replacing at the call site.

PR #562, opened 2025-11-08
rbx_binary: Clear ustr cache in benches

krakow10:
The ustr cache warming up causes subsequent runs to neither hash nor allocate new strings, leading to an unfair comparison to the first cold run, which is a more typical usage scenario.  The cache cannot be cleared during the serialize tests because the tree used to serialize contains Ustr strings. The performance should have gone down... is it just me or does this _increase_ the benchmark performance???


krakow10:
That makes sense for rojo.  Do you even see any performance difference in the benchmarks?  Maybe this can just be closed if there is little effect and it doesn't represent the use case of the primary library consumer.

krakow10:
Whatever, I'd rather have other changes merged, and you seem to be reviewing them in stack order rather than FIFO.

Issue #350, opened 2023-08-15
Memory leak in SharedString


cassanof:
I ran a memory profiler against my program, as it was leaking memory at the end of the program. I found the problem to be with how SharedString is allocated or deallocated. I think it may be a race condition in the `Drop` implementation: https://github.com/rojo-rbx/rbx-dom/blob/9d2418d834dd5e47dc2927a5860ba339942414ef/rbx_types/src/shared_string.rs#L101
. I'm not entirely sure; this problem only occurs when you parse concurrently.

cassanof:
In my run: In total, `1,343,532` bytes were allocated for the SharedString. At the end of the program, there were `671,760` left.

cassanof:
Looks like this is a known race condition with `Arc::try_unwrap`: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap

cassanof:
Can I make a PR?

Dekkonot:
I would be willing to accept a PR for this, yeah. I'll be happy if it's easily solvable like that.

cassanof:
That would fix it, however there is another unrelated issue i spoted (dangling pointer). I think the lock needs to be hoisted up (in that case, try_unwrap is fine, because there can't be more than 1 thread executing it). Here is the full fix I propose (pardon bad indentation):

```rs
impl Drop for SharedString {
    fn drop(&mut self) {
        // If the reference we're about to drop is the very last reference to
        // the buffer, we'll be able to unwrap it and remove it from the
        // SharedString cache.
            let mut cache = match STRING_CACHE.lock() {
                Ok(v) => v,
                Err(_) => {
                    // If the lock is poisoned, we should just leave it
                    // alone so that we don't accidentally double-panic.
                    return;
                }
            };
        if Arc::try_unwrap(self.data.take().unwrap()).is_ok() {

            cache.remove(&self.hash);
        }
    }
}
```


I know this makes it very inefficient, but it is necessary without making large changes. Here is why: 
1. Thread A is doing some stuff, allocates shared string "AAAA"
2. Thread A is done done doing stuff, runs Drop on the string "AAAA". There are no more references, the current execution context is inside the if statement.
3. At the same exact time, thread B allocates the shared string "AAAA", it gets the lock before thread A. The string is already allocated, so the Arc is given to the data. The lock is now freed.
4. Thread A now gets the lock, and removes the hash from the cache.
5. Thread B now has a dangling Arc

[...]

PR #577, opened 2025-12-03
Refactor SharedString to `Arc<[u8]>`

krakow10:
This has always bothered me, maybe I am good enough at Rust now to tackle this.
- Arc::from(&[u8]) allocates a new copy of the slice, which is exactly what we want.
- ~~SharedString::new takes `impl AsRef<[u8]>` for easy migration of previous callers.~~
- Option was removed.  I can see that it was there for the drop impl, but I hope that my reasoning that it can be removed is sound.
- Prior art for Arc race conditions: #350
- There are additional intermediate allocation optimizations that can be made in conjunction with #530: remove the .to_owned() after .read_binary_string()? when constructing a SharedString or NetAssetRef

Notes:
- The Drop impl for SharedString includes mem::take which calls Arc::default().  I believe this allocates a new arc on the heap, only to drop it moments later, which is pretty wacky.

I wish I knew how to test for memory leaks!

### DIFF
--- a/rbx_types/src/shared_string.rs
+++ b/rbx_types/src/shared_string.rs
@@ -25,8 +25,7 @@ pub struct SharedString {
 
 impl SharedString {
     /// Construct a SharedString from an owned buffer of data.
-    pub fn new(data: impl AsRef<[u8]>) -> SharedString {
-        let data = data.as_ref();
+    pub fn new(data: &[u8]) -> SharedString {
         let hash = blake3::hash(data);
 
         let data = {
@@ -199,7 +198,7 @@ pub(crate) mod serde_impl {
 
         fn visit_str<E: Error>(self, str: &str) -> Result<Self::Value, E> {
             let buffer = base64::decode(str).map_err(E::custom)?;
-            Ok(SharedString::new(buffer))
+            Ok(SharedString::new(&buffer))
         }
     }
 
@@ -211,7 +210,7 @@ pub(crate) mod serde_impl {
                 // For compatibility reasons, we use `Vec<u8>`'s implementation
                 // of deserialize.
                 let buffer = <Vec<u8>>::deserialize(deserializer)?;
-                Ok(SharedString::new(buffer))
+                Ok(SharedString::new(&buffer))
             }
         }
     }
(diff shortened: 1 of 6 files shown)


rbx_types/src/shared_string.rs
```diff
@@ -104,18 +102,25 @@ impl Drop for SharedString {
...
+        if Arc::strong_count(&self.data) != 1 {
+            return;
+        };
+
+        let Ok(mut cache) = STRING_CACHE.lock() else {
+            // If the lock is poisoned, we should just leave it
+            // alone so that we don't accidentally double-panic.
+            return;
+        };
+
```
krakow10: I didn't realize that the string cache was upgrading weak references.  Maybe this double-check band-aid would work and not suffer from the slowness of directly locking every time that would happen if the strong count check was simply moved here.  I don't understand why the Arc methods in the standard library seem so limited.

```suggestion

        if Arc::strong_count(&self.data) != 1 {
            return;
        };

```

krakow10:
I've come up with an entirely new strategy.  After reading the Arc source code for a while, I've come to understand that once the strong count hits 0, Weak references cannot be upgraded anymore and are effectively locked out.  I devised code that drops the arc as early as possible, and then counts on that to plug all the parallelism holes.  I hope my reasoning in the comments is correct.

### Duplicate backing bug
I added new logic to prevent removing the wrong Weak reference from the string cache.  I believe master branch can also suffer from this with a very precise and unlikely order of operations:
https://github.com/rojo-rbx/rbx-dom/blob/7ef7d448ba6d30c4cb217b228d5c22c6f5ac07f3/rbx_types/src/shared_string.rs#L107-L108
1. The if statement evaluates to true, but the string cache lock is not yet acquired.
2. Another thread calls SharedString::new and acquires the string cache lock.  This thread replaces the weak reference with a new downgraded Arc with a distinct allocation backing.
3. The Drop impl acquires the string cache lock and deletes the weak reference that was just inserted by the other thread.
4. The next call to SharedString::new creates a duplicate string because there is no string cache entry.

### Performance
I expect that this has very minimal impact on performance, but I'm terrified of testing it because I've experienced multiple PC crashes, and only while running rbx-dom benchmarks.  Rebooting my computer is extremely fussy, the BIOS often does not post and I can get locked out for hours.  Please test and make sure it doesn't cause a regression.

### Miri
I installed miri and tried `cargo miri test` but it errors out because of some isolation shenanigans.  I will need to learn more about miri if I want to use it for memory profiling.

PR #579, opened 2025-12-04
Refactor Tags to String

krakow10:
A comment in tags.rs says:
https://github.com/rojo-rbx/rbx-dom/blob/7ef7d448ba6d30c4cb217b228d5c22c6f5ac07f3/rbx_types/src/tags.rs#L14-L15

This does exactly that.

### Design Choices
- `Tags::push` can be called many times, whereas `Tags::encode` is only called once, so I've opted to include a '\0' postfixing every tag to move branching on is_empty() from push into encode.

### Changed Behaviour
- '\0's are trimmed from the beginning and end of the buffer, but are no longer trimmed from the middle.  This is to keep the decode function as close to extend_from_slice (push_str) as possible.  If needed, this can be remedied by making decode slower (`TagsIter::new(str).collect()`)
- Empty strings pushed into tags will be skipped when iterating.  The old behaviour was to iterate over all elements of the inner Vec.   If needed, this can be remedied by adding a special case for empty tags to the iterator, since .split('\0') returns an iterator which yields one element for an empty string.

### Changed Interfaces
- `Tags::decode`'s Result::Err is changed from `FromUTF8Error` to `UTF8Error`.  All call sites in rbx-dom throw away this error value.  The old error value can be hobbled together with a duplicated UTF8 check if we don't want to change this public interface.
- `Tags::encode` return value is changed from `Vec<u8>` to `&[u8]`.  It would be a waste to make an intermediate allocation here.

### Performance
I measured no significant change to performance.

### New Tests
- Tags encode/decode behaviour is tested more extensively. Yay!

### DIFF
--- a/rbx_binary/src/serializer/state.rs
+++ b/rbx_binary/src/serializer/state.rs
@@ -920,7 +920,7 @@ impl<'dom, 'db: 'dom, W: Write> SerializerState<'dom, 'db, W> {
                                 }
                                 Variant::Tags(value) => {
                                     let buf = value.encode();
-                                    chunk.write_binary_string(&buf)?;
+                                    chunk.write_binary_string(buf)?;
                                 }
                                 Variant::Attributes(value) => {
                                     let mut buf = Vec::new();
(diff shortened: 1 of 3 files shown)


krakow10:
Preemptively closing this assuming Roblox deduplicates Tags upon deserialization.  In that case, this should be a HashSet.

PR #583, opened 2025-12-05
Improve `CloneContext::rewrite_refs`

krakow10:
Two passes of identical loops, reduced to one, so 2x speed.  There aren't any benches for DOM manipulation so I don't actually know what the real performance looks like.

~~This optimization requires unsafe!  Here is my explanation:~~
<details>
<summary>Show</summary>

```Rust
// We need to hold the lifetime of a mutable reference,
// and at the same time, an immutable reference.
//
// dest *const
// └ instances *const
//
// dest *mut
// └ instances *mut
//   └ instance *mut
//     └ properties *mut
//
// We are holding `&mut dest` for properties and `&dest` for instances,
// thus the borrow checker complains, even though we know in this case
// that it is safe.
//
// Undefined behaviour would only be introduced if we used our
// raw pointer to hold references to instance properties while
// they are being rewritten.
```

</details>

I did include an optimization without unsafe in dbaf4fe6eef218665f756d9cc07387e50bcc356b, however, it may not be faster in every shape of DOM input, especially when there is many instances but few / no rewrites.

Edit:
Latest version is back to no unsafe.


kennethloeffler (commented):
The code as presented does not pass Miri under stacked borrow rules:
```
test dom::test::clone_into_external ... error: Undefined Behavior: trying to retag from <17037435> for SharedReadOnly permission at alloc3240105[0x0], but that tag does not exist in the borrow stack for this location
   --> rbx_dom_weak/src/dom.rs:557:31
    |
557 |                     !unsafe { &*dest_instances }.contains_key(original_ref) {
    |                               ^^^^^^^^^^^^^^^^ this error occurs as part of retag at alloc3240105[0x0..0x40]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <17037435> was created by a SharedReadOnly retag at offsets [0x0..0x40]
   --> rbx_dom_weak/src/dom.rs:540:40
    |
540 |         let dest_instances: *const _ = &dest.instances;
    |                                        ^^^^^^^^^^^^^^^
help: <17037435> was later invalidated at offsets [0x0..0x90] by a Unique function-entry retag inside this call
   --> rbx_dom_weak/src/dom.rs:543:28
    |
543 |               let instance = dest
    |  ____________________________^
544 | |                 .get_by_ref_mut(new_ref)
    | |________________________________________^
    = note: BACKTRACE (of the first span) on thread `dom::test::clone_into_external`:
    = note: inside `dom::CloneContext::rewrite_refs` at rbx_dom_weak/src/dom.rs:557:31: 557:47
note: inside `dom::WeakDom::clone_into_external`
   --> rbx_dom_weak/src/dom.rs:409:9
    |
409 |         ctx.rewrite_refs(dest);
    |         ^^^^^^^^^^^^^^^^^^^^^^
note: inside `dom::test::clone_into_external`
   --> rbx_dom_weak/src/dom.rs:707:27
    |
707 |         let cloned_root = dom.clone_into_external(dom.root_ref, &mut other_dom);
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside closure
   --> rbx_dom_weak/src/dom.rs:689:29
    |
688 |     #[test]
    |     ------- in this attribute macro expansion
689 |     fn clone_into_external() {
    |                             ^
``` 
The problem is that the code creates a reference to `dest` at the `get_by_ref_mut` method call, which invalidates the pointer. Usage of raw pointers must abide by the aliasing rules. You cannot generally [use references and raw pointers alongside each other without violation](https://doc.rust-lang.org/std/ptr/index.html#safety):

> The result of casting a reference to a pointer is valid for as long as the underlying allocation is live and no reference (just raw pointers) is used to access the same memory. That is, reference and pointer accesses cannot be interleaved.

For an optimization like this to work, we'd need to use interior mutability, which will involve significant refactoring. That being said, I wrote a quick benchmark (set up very similarly to rbx_binary's benchmarks):
```rust
use std::time::Duration;

use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
use rbx_dom_weak::{InstanceBuilder, WeakDom};

pub fn clone_into_external(c: &mut Criterion) {
    let mut group = c.benchmark_group("clone_into_external");

    group
        // cranking sample size and measurement time to ensure we get a
        // statistically significant result
        .sample_size(1000)
        .measurement_time(Duration::from_secs(30))
        .bench_function("1000 ref rewrites", |b| {
            b.iter_batched(
                || {
                    let mut dest_tree = WeakDom::new(InstanceBuilder::new("Folder"));
                    dest_tree.reserve(1000);

                    let mut src_tree = WeakDom::new(InstanceBuilder::new("Folder"));
                    let src_instance = InstanceBuilder::new("Folder");
                    let src_referent = src_instance.referent();
                    let mut children = Vec::with_capacity(1000);

                    for _ in 0..1000 {
                        children.push(
                            InstanceBuilder::new("ObjectValue")
                                .with_property("Value", src_referent),
                        );
                    }
                    src_tree.insert(src_tree.root_ref(), src_instance.with_children(children));
                    (src_referent, src_tree, dest_tree)
                },
                |(src_instance, src_tree, mut dest_tree)| {
                    src_tree.clone_into_external(src_instance, &mut dest_tree);
                },
                BatchSize::SmallInput,
            );
        });
}

criterion_group!(bench_suite, clone_into_external,);

criterion_main!(bench_suite);
```
On an M1 MacBook Air, I only saw a few percentage points of performance uplift using the (broken) code, so it's probably not worth it.

Dekkonot:
> I don't understand what retagging means and how that can invalidate a perfectly good pointer.

Retagging is a term that doesn't really exist outside of Miri (yet?). The doc linked in the error [here](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md) and the [intro blog post](https://www.ralfj.de/blog/2018/11/16/stacked-borrows-implementation.html) do a good job at explaining but essentially:

1. Stack borrows are how Miri has decided to implement tracking pointer aliasing, which is important for reasons. I can get into it if you want but it's kind of its own topic.
2. Pointers are tagged in Miri with their origin. A retag is caused by a pointer changing what "type" it is. Think something like `let foo = &mut bar; let foo = &*bar;`.
3. A pointer being retagged *does not* invalidate it, but a tag being incorrect does mean that there is a potential alias violation happening. Assuming Miri is right, which isn't a given. This *is* unstable right now.

I am inclined to agree with Miri here because I don't fully understand why it thinks this is bad. We can likely just ask though, as this seems like a teachable moment either way.

kennethloeffler:
It's UB essentially because it breaks Rust reference invariants:

```rust
        // Taking a raw pointer derived from &mut dest, this is fine
        let dest_instances: *const _ = &dest.instances;

        for &new_ref in self.ref_rewrites.values() {
            // Taking a mutable reference derived from &mut dest, which stays live
            // until we exit the loop. Still fine, but we cannot dereference the
            // pointer anymore without invoking UB
            let instance = dest
                .get_by_ref_mut(new_ref)
                .expect("Cannot rewrite refs on an instance that does not exist");

            for prop_value in instance.properties.values_mut() {
                if let Variant::Ref(original_ref) = prop_value {
                    if let Some(new_ref) = self.ref_rewrites.get(original_ref) {
                        *original_ref = *new_ref;
                    } else if
                    // The mutable reference taken earlier in the parent loop is still live, and
                    // dereferencing + creating the reference breaks exclusivity guarantees
                    !unsafe { &*dest_instances }.contains_key(original_ref) {
                        *original_ref = Ref::none();
                    }
                }
            }
        }
```

The compiler assumes that mutable references are not aliased by any other reference or pointer access. So there's no actual machinery invalidating the pointer, it's just that trying to dereference it there breaks the assumptions of the compiler. It is undefined to have both a mutable and immutable reference to overlapping memory at the same time. This happens to work right now, but it could break when we change (perhaps entirely unrelated!) code or the compiler does more or different optimizations. 

Side note, the pattern `unsafe { &*ptr }` is pretty spooky, because it creates a reference out of thin air that the compiler cannot check. In this case it allowed us to create an immutable reference while a mutable reference to overlapping memory was live, which is clearly UB. It doesn't really matter if it touches exactly the same piece of memory or not, because the compiler doesn't care or know about that, it just assumes that the aliasing rules are obeyed.

krakow10:
@Dekkonot Thanks for the explanation about tagging.  Seems like it's a formalization of borrow checking?  Maybe that was the stacked borrows, and the tagging is just part of the process.

> I am inclined to agree with Miri here because I don't fully understand why it thinks this is bad.

I'm in the same boat, I don't understand it but I am inclined to believe that Miri knows more about UB than what I've learned on my junior unsafe adventures.

> trying to dereference it there breaks the assumptions of the compiler

This is a very clear explanation.  So it's UB not because it's wrong in every case, but because the assumptions that the compiler uses for optimizations are broken.

### The Benchmark
I committed the benchmark to the [clone-bench](https://github.com/krakow10/rbx-dom/tree/clone-bench) branch and changed it to measure cloning of the Miner's Haven place.
My results show a 4% improvement with the safe remove-insert strategy in 63f8f178ee32b0ca0b96197272db34b9e0f3090d.
image
The "comments v3" commit a5616a3472b7a7443b04d4a453e853e591b7ddc6 with UB shows 7% improvement over master.
image

krakow10:
The remove-and-insert strategy could be optimized with methods that don't exist.
1. Get OccupiedEntry of instance `let entry = dest.instances.entry(*rewrite_ref)`
2. Split OccupiedEntry into separate &RawTable and &mut Instance ahead of properties loop
3. Check if table contains key equivalent to `dest.instances.contains_key`
4. Modify instance

No removing and re-inserting the entry this way, but no methods exist to do steps 2 and 3.

Issue #581, opened 2025-12-05
Add Non-Nil Ref Type


krakow10:
If a new type was added, say `NonNilRef(NonZeroU128)`, a lot of strange edge cases jump out such as #517.  The type can also be used within `Ref(Option<NonNilRef>)` as the inner type and for `Content::Object(NonNilRef)` since [the Roblox documentation for Content](https://create.roblox.com/docs/reference/engine/datatypes/Content) states that Content::Object always contains a non-nil ref.

Here is the primary example that shows that this is a more accurate data model:
before:
```Rust
if parent_ref.is_some() {
    let parent = self.instances.get_mut(&parent_ref).unwrap();
```
after:
```Rust
if let Some(parent_non_nil) = parent_ref.into_non_nil() {
    let parent = self.instances.get_mut(&parent_non_nil).unwrap();
```

I have some WIP branches implementing this
- https://github.com/krakow10/rbx-dom/compare/master...non-nil-ref setting up the types for Ref and Content without making any breaking changes
- https://github.com/krakow10/rbx-dom/compare/non-nil-ref...non-nil-dom putting it everywhere that it makes sense, making large sweeping breaking changes in the process (currently failing tests)

Dekkonot:
I am vetoing this because of the breaking change and cognitive load it brings with it. You're not wrong that it better encodes correct behavior, but I don't think that the existing behavior is really all that complicated and the improvements don't justify the cost.

Thank you for the idea though!

krakow10:
Thanks for spending the time to review.

PR #584, opened 2025-12-07
SomeRef v1 Part 1: Introduce SomeRef

krakow10:
Adds a new type `SomeRef(NonZeroU128)` as described in #581.  The new type signature of Ref reads as follows: `OptionalRef(Option<SomeRef>)`. The names are open to debate.  Many Ref values passed as function arguments should never be allowed to be null, but this is not reflected by the data model.  This series implements the new type in 3 stages:
1. Introduce just the new type and its bells and whistles, no code is broken.
2. Implement the new type internally within the library with minimal public interface changes.
3. Implement the new type where applicable in public interfaces.

Some tests are modified to use the public interface of the things being tested, because in part 2 the public interface is not supposed to change, but the internals do change.

Notes:
- Content is given a new function `from_some_ref`, but does not use the new type internally.  That change is in part 2.
- Ref is renamed to OptionalRef to more accurately describe it, and a type alias for `Ref` is created.

### DIFF
--- a/rbx_dom_weak/src/dom.rs
+++ b/rbx_dom_weak/src/dom.rs
@@ -592,7 +592,7 @@ mod test {
         let target = InstanceBuilder::new("Folder")
             .with_name("Target")
             .with_child(InstanceBuilder::new("Part").with_name("Some Child"));
-        let target_ref = target.referent;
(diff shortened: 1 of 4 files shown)


Dekkonot:

PR #585, opened 2025-12-07
SomeRef v1 Part 2: Use SomeRef Internally

### LINKED ISSUE #581: Add Non-Nil Ref Type

krakow10:
- Closes #581.
- Depends on #584.

Changes many types to use `SomeRef` and `Option<SomeRef>` internally.  The goal is to implement the new type as thoroughly as possible with the fewest number of observable changes.

### Highlights
```Rust
pub struct WeakDom {
    instances: AHashMap<SomeRef, Instance>,
    root_ref: Option<SomeRef>,
    ...
}
```
```Rust
pub struct Instance {
    referent: SomeRef,
    children: Vec<SomeRef>,
    parent: Option<SomeRef>,
    ...
}
```
```Rust
pub enum ContentType {
    None,
    Uri(String),
    Object(SomeRef),
}
```
Every new if statement added is a previously unhandled case.  For example, at the bottom of `WeakDom::transfer`:
before:
```Rust
let dest_parent = dest.instances.get_mut(&dest_parent_ref).unwrap_or_else(|| {
    panic!("cannot move an instance into an instance that does not exist")
});
dest_parent.children.push(referent);
```
after:
```Rust
if let Some(dest_some_ref) = dest_parent_ref {
    let dest_parent = dest.instances.get_mut(&dest_some_ref).unwrap_or_else(|| {
        panic!("cannot move an instance into an instance that does not exist")
    });
    dest_parent.children.push(referent);
}
```
This is just like #517, it turns out the transfer function also has the same bug where the destination parent is not allowed to be null, and this is fixed by nature of using a more accurate data model.

### Observable Changes
- `WeakDom::from_raw` and `WeakDom::into_raw` expose the new inner types.  These functions are designed to do exactly that though, which is why no conversion is performed.
- ContentType::Object exposes the new inner type.  Improving the data model accuracy of this object was the inspiration for these changes.
- No tests should change other than tests involving the above observable changes.

### Unsafe
  - `Instance::children(&self) -> &[Ref]` must transmute using unsafe to keep the same return type.  This is removed in part 3.

### Performance
I observed no performance improvements greater than +3%.

### Rojo
Since Rojo uses the WeakDom::into_raw function, it requires 3 lines of code changes: https://github.com/krakow10/rojo/compare/master...ref-some. Rojo also touches the ContentType::Object inner type, but no code changes are necessary there.


krakow10:
Why did it auto request review on this PR but not part 1

Dekkonot:

PR #586, opened 2025-12-07
SomeRef v1 Part 3: Use SomeRef Externally

krakow10:
- Depends on #585.

Changes the public interfaces to use the `SomeRef` type where applicable.  Library consumers will get the privilege and joy of updating their code to use a more accurate data model.  Most function signatures that involve Ref should not allow null Refs, so the number of uses of OptionalRef is dramatically reduced.
### Highlights
  - instance.referent() -> `SomeRef`
  - instance.parent() -> `Option<SomeRef>`
  - instance.children() -> `&[SomeRef]`
  - dom.root_ref() -> `Option<SomeRef>`
  - dom.get_by_ref(`referent: SomeRef`) -> Option<&Instance>
  - dom.insert(`parent_ref: impl Into<OptionalRef>`, root_builder: InstanceBuilder) -> `SomeRef`
    - dom.insert(`None`, InstanceBuilder::new("Folder")) will be all the rage with that trait bound

### Rojo
Attempting to compile Rojo with these changes spits out 66 errors.  Changing `pub type Ref = OptionalRef;` to `pub type Ref = SomeRef;` reduces it to 40 errors, so maybe the SomeRef type really should be renamed to just Ref?  The majority of these errors are due to dom.root_ref() being made optional.


Dekkonot:

PR #587, opened 2025-12-08
SomeRef v2: Replace Ref

### LINKED ISSUE #581: Add Non-Nil Ref Type

krakow10:
- Supersedes #586.
- Supersedes #585.
- Supersedes #584.
- Closes #581.
- The overall difference of this is much smaller than the individual diffs of parts 1-3. 

**All of the descriptions of the superseded pull requests 1-3 apply.**

Removes OptionalRef (old Ref) entirely, and renames SomeRef to Ref. All of the same wonderful breakage and none of the OptionalRef baggage.  I wrote v1 because I didn't think it was possible to remove OptionalRef, but I was mistaken.

Notes:
- The Serialize and Deserialize implementations of `Variant::Ref(Option<Ref>)` may need attention for rojo protocol compatibility reasons
- Should Ref::none() be implemented and return Option::None?  It would only be a migration convenience.


Dekkonot:

PR #588, opened 2025-12-11
Defer Deserialization of Prop Chunks

krakow10:
As mentioned in https://github.com/rojo-rbx/rbx-dom/pull/532#issuecomment-3621583121, procedurally growing instance properties is siphoning performance potential from the deserializer.  This PR moves deserialization of prop chunks until after the number of PROP chunks for each class is known.

### Performance
This does not seem to have the expected significant performance improvement.  I'm seeing a 5% improvement in Miner's Haven Deserialize on my laptop, but the benchmarks are noisy.  Please test and compare.


krakow10:
I'm going to assume this does not have any performance benefit on windows and close in favour of changes yielding tangible gains (#591).

PR #590, opened 2025-12-17
Implement Typestate Deserializer

### LINKED ISSUE #589: Enforce Chunk Order
There is no enforced chunk order, the file will simply fail to decode.  Introduce serializer and deserializer stages using the typestate pattern.  This way relevant state only exists during the relevant stage, and chunk order is also enforced.

krakow10:
- Closes #589

Enforces an absolute chunk order.  Only relevant state is included in each stage of the deserializer, so collections which are not yet needed are not yet created, and values which are no longer used are dropped.

Notes:
- A new InnerError variant `UnexpectedChunk` is introduced.
- I really like trait bounds
- This is a fair bit of code just to enforce a chunk order

### Performance
I saw the 1000 parts deserialize benchmark fluctuate wildly among my test systems (+50% throughput), but that may not be an indication of anything at all because no other benchmarks are affected.


krakow10:
This doesn't make sense given the existence of other chunk types such as SIGN, OSTR, HASH. The appearance of an unexpected chunk would mean this throws an error instead of gracefully ignoring it. However, the deserializer should be written so that the order does not cause silent corruption.

PR #591, opened 2025-12-17
Instance Slices (+29%)

krakow10:
Push instances into a regular ol' Vec, and keep track of the inserted range in TypeInfo.  This means that all instances of a specific class are in a contiguous slice, and are also in the correct iteration order expected by the decode_prop_chunk loops.  No more `instances_by_ref.get_mut(referent).unwrap()`. A HashMap tracks which i32 ref id corresponds to which instance using indices.

### Performance
I observe a 29% improvement in throughput in the Miner's Haven deserialize benchmark.

### Ramblings

There's probably some more optimizations hiding in `DeserializerState::finish`.  ~~The algorithm is currently using Vec::swap_remove to remove the elements from the instances list.~~ (swap_remove removed in 695e88ef0212d6947aeeac4cef499457523d14d1) The `instances_to_construct` machinery can probably be combined with the instances_by_ref and PRNT chunk decode in some clever way to do even less work.

~~I wrote an unsafe version that avoids swap_remove here: 3df205bee0cefcdee65cd398ccbc036b0436b827 but there doesn't seem to be a significant impact.~~ I wrote a safe version using a similar technique. Theoretically it should be faster than the swap_remove method, but I haven't put it to the test and it's not the source of the major performance improvement. 

### Lore
The instances could be owned by type_info directly, I was able to run type_ids in parallel with rayon ([branch](https://github.com/krakow10/rbx-dom/tree/rayon)) using this method.  I got a 1% speedup!!! Wow!!!  Probably because it spent the entire time waiting for 36000 Parts to decode while every other class only has <2000 instances.  The InstanceKey technique would still work, but removing each instance in finish() would be triple indirection (key, type_id, index) instead of double indirection (key, index).

I was using the indexmap crate for the Vec + HashMap abomination several times, but it's better to implement separately because of the separable mutable + immutable aliasing over the Vec + HashMap  parts.

This has three previous incarnations:
- v1 https://github.com/krakow10/rbx-dom/compare/categorize-chunks...categorize-chunks2: Store instances in TypeInfo.  The problem with this is it's based on the nasty categorize-chunks branch.  At least I found the performance I was looking for.
- v2 https://github.com/krakow10/rbx-dom/compare/master...type_info-instances: better, but the design is recollecting instances twice - once into instances_by_ref and then again into the dom.
- v3 https://github.com/krakow10/rbx-dom/compare/typestate-de...instance-slices: A breakthrough design using a Vec to back the instances to avoid recollecting twice.  I thought the typestate pattern would make it easier to implement, but it turns out it was my past failings (v1 and v2) that were calling out for typestate.

### DIFF
--- a/rbx_binary/src/deserializer/state.rs
+++ b/rbx_binary/src/deserializer/state.rs
@@ -1443,7 +1443,7 @@ rbx-dom may require changes to fully support this property. Please open an issue
             if parent_ref == -1 {
                 self.root_instance_refs.push(id);
             } else {
-                let instance_key = self.instance_key_by_ref.get_mut(&parent_ref).unwrap();
+                let instance_key = self.instance_key_by_ref.get(&parent_ref).unwrap();
                 let instance = self.instances.get_mut(instance_key.key).unwrap();
                 instance.children.push(id);
             }
--- a/rbx_binary/src/deserializer/state.rs
+++ b/rbx_binary/src/deserializer/state.rs
@@ -87,7 +87,8 @@ struct InstanceKey {
 /// instance. Incrementally built up by the deserializer as we decode different
 /// chunks.
 struct Instance {
-    /// The ref_id for this instance
+    /// The ref_id for this instance.  This is only used in one place,
+    /// so may be a good place to look for inspiration for a refactor.
     referent: i32,
 
     /// A work-in-progress builder that will be used to construct this instance.
--- a/rbx_binary/src/deserializer/state.rs
+++ b/rbx_binary/src/deserializer/state.rs
@@ -87,15 +87,19 @@ struct InstanceKey {
 /// instance. Incrementally built up by the deserializer as we decode different
 /// chunks.
 struct Instance {
-    /// The ref_id for this instance.  This is only used in one place,
-    /// so may be a good place to look for inspiration for a refactor.
-    referent: i32,
-
     /// A work-in-progress builder that will be used to construct this instance.
     builder: InstanceBuilder,
 
     /// Document-defined IDs for the children of this instance.
     children: Vec<i32>,
+}
+impl Instance {
+    fn empty() -> Self {
+        Self {
+            builder: InstanceBuilder::empty(),
+            children: Vec::new(),
+        }
+    }
 }
 
 /// Properties may be serialized under different names or types than
@@ -316,7 +320,6 @@ impl<'db, R: Read> DeserializerState<'db, R> {
                 },
             );
             self.instances.push(Instance {
-                referent,
                 builder,
                 children: Vec::new(),
             });
@@ -1488,14 +1491,9 @@ rbx-dom may require changes to fully support this property. Please open an issue
         }
 
         while let Some((referent, parent_ref)) = instances_to_construct.pop_front() {
-            // swap_remove swaps the last element into the removed element.
-            // We need to keep `instance_key_by_ref` up to date with the swaps as they happen.
-            let instance_key = self.instance_key_by_ref.get(&referent).unwrap().key;
-            if let Some(last) = self.instances.last() {
-                let last_instance_key = self.instance_key_by_ref.get_mut(&last.referent).unwrap();
-                last_instance_key.key = instance_key;
-            }
-            let instance = self.instances.swap_remove(instance_key);
+            let instance_key = self.instance_key_by_ref.remove(&referent).unwrap().key;
+            // Replace each instance with an impostor!
+            let instance = core::mem::replace(&mut self.instances[instance_key], Instance::empty());
             let id = self.tree.insert(parent_ref, instance.builder);
 
             for referent in instance.children {
--- a/rbx_binary/src/deserializer/state.rs
+++ b/rbx_binary/src/deserializer/state.rs
@@ -93,14 +93,6 @@ struct Instance {
     /// Document-defined IDs for the children of this instance.
     children: Vec<i32>,
 }
-impl Instance {
-    fn empty() -> Self {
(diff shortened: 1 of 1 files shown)


rbx_binary/src/deserializer/state.rs
```diff
@@ -755,20 +753,18 @@ rbx-dom may require changes to fully support this property. Please open an issue
...
                 VariantType::BrickColor => {
-                    let values = chunk.read_interleaved_u32_array(type_info.referents.len())?;
-
-                    for (value, referent) in values.zip(&type_info.referents) {
-                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
-                        let color = value
-                            .try_into()
-                            .ok()
-                            .and_then(BrickColor::from_number)
-                            .ok_or_else(|| InnerError::InvalidPropData {
```
kennethloeffler: ok_or_else is ok for simply converting Option to Result

rbx_binary/src/deserializer/state.rs
```diff
@@ -430,39 +443,35 @@ This may cause unexpected or broken behavior in your final results if you rely o
...
-                    for referent in &type_info.referents {
-                        let instance = self.instances_by_ref.get_mut(referent).unwrap();
+                    for instance in instances {
                         let buffer = chunk.read_binary_string()?;
 
-                        let value = Tags::decode(buffer.as_ref()).map_err(|_| {
-                            InnerError::InvalidPropData {
+                        let Ok(value) = Tags::decode(buffer.as_ref()) else {
+                            return Err(InnerError::InvalidPropData {
                                 type_name: type_info.type_name.to_string(),
```
kennethloeffler: Explain the reasoning for replacing map_err()? and ok_or_else()? with let else. In all of the cases, we are simply mapping Result->Result or Option->Result, so I don't see the need to change to let else.

rbx_binary/src/deserializer/state.rs
```diff
@@ -291,25 +299,31 @@ impl<'db, R: Read> DeserializerState<'db, R> {
...
         // TODO: Check object_format and check for service markers if it's 1?
 
-        for &referent in &referents {
-            self.instances_by_ref.insert(
+        let start = self.instances.len();
+        for (key, referent) in referents.enumerate() {
+            let builder =
+                InstanceBuilder::with_property_capacity(type_name.as_str(), prop_capacity);
+            // TODO: assert / error when the ref already exists.
+            self.instance_key_by_ref.insert(
```
kennethloeffler, line 307: These TODOs are possible using the return value of insert, e.g. (could use combinators too):
```rust
let existing_instance_key = self.instance_key_by_ref.insert(
    referent,
    InstanceKey {
        key: start + key,
        referent: builder.referent(),
    },
);

match existing_instance_key {
    Some(_) => return Err(InnerError::DuplicateRef(referent)),
    None => (),
};
```
Since we're already here, I'd like to see the check performed instead of a TODO comment.

kennethloeffler (commented):
This significantly reduces the amount of time we spend jumping around in memory in the binary deserializer's hot path, nice. I didn't see as drastic of an improvement on my M1 MacBook Air (it's more in the ballpark of a 10% reduction for me), but I expect the effectiveness of this optimization to depend on memory read latency and the performance characteristics of the CPU caches, branch predictor, etc.

A few comments:

rbx_binary/src/deserializer/state.rs
krakow10: I did this because using a closure has borrow checker issues with `type_info`.  It's also possible to use a let binding, I opted to change from a let binding to to let-else in commit 8fcdf07618ceaf6c41c19d35e5059b58050473be.  I changed it because the let binding is only used in the error path.

rbx_binary/src/deserializer/state.rs
krakow10: This is implemented in #593.

kennethloeffler: Alright, let's address these todos once this lands then.

rbx_binary/src/deserializer/state.rs
kennethloeffler: Let's keep these as map_err and ok_or_else. I just tried the previous incantations against the rest of the changes and they compiled clean, my thinking is not to change code if we don't have to. IMO it's also more stylish to use combinators for these cases, mapping between option and result types is why they're there.

krakow10: I've reverted 8fcdf07618ceaf6c41c19d35e5059b58050473be and split TypeInfo at the top.

kennethloeffler: I would argue that we do not strictly need the changes from [fb1b4d1](https://github.com/rojo-rbx/rbx-dom/pull/591/commits/fb1b4d1fb2657fe0376ec5662236c5c590d40560), but I appreciate the effort and the slight readability improvement they bring.

kennethloeffler (approved):
Cool, let's fix up https://github.com/rojo-rbx/rbx-dom/pull/593 next. Thanks for the PR!

krakow10:
I believe that will involve merging the master branch into the branch for that PR, so this would need to be merged first.  Thanks so much for the review :heart:

kennethloeffler:
I think you should have a merge button available to you at this moment. If we approve a PR, and you don't have any further qualms about merging, feel free to press it. If you didn't see it, let me know and we'll sort it out.

krakow10:
No merge button for me
image

kennethloeffler:
Annoying, I'll figure that out tomorrow.

PR #594, opened 2025-12-26
Remove Unwrap in `inner_insert`

krakow10:
Inner insert has exclusive ownership of this object, just modify the unique id before giving away ownership.


Dekkonot (approved):
Low priority, but we'll try to get this merged before we do a new release, whenever that is.

krakow10:
Better late than never

PR #598, opened 2026-02-21
Replace "Fast Path" with Explicit Function Call

krakow10:
In #463, a "fast path" is introduced by making an optional queue argument to avoid a heap allocation.  The heap allocation can instead be avoided by using `VecDeque::new()` and directly calling `insert`, which also means the `if let Some(queue) = queue` in `insert` can be removed. This change effectively unrolls the first loop iteration.


krakow10:
@kennethloeffler this is essentially a simpler version of your pull request #463 without the extra code path. If you could review this that'd be great.

PR #599, opened 2026-03-27
`AvatarJointUpgrade` should be an alias for `AvatarJointUpgrade_SerializedRollout`

badcc:
when I tried to re-generate database, it seemed to re-order everything.. maybe someone can help :)


kennethloeffler:
Hi, thanks for the PR.

The reason why it looks like things have been reordered after regenerating the reflection database is because we alphabetize it, and there have been lots of changes to the API dump since we've last updated. It's expected -- but I'll update the database in a different PR, rebase this one, then regenerate again so we can see your changes cleanly!

kennethloeffler (approved):
Looking good, thank you!

badcc:
thanks for the rebase & approve!

PR #600, opened 2026-04-23
Fix `PropertiesSerialize` and add `PropertyTransitionsSerialize`.

cameronpcampbell:
## Summary
- Fixes #597.
- Adds support for `PropertyTransitionsSerialize`.

## Changes
- Introduces property `Add` patches. Aliases `StyleRule.PropertiesSerialize` to `Properties` and `PropertyTransitionsSerialize` to `PropertyTransitions`.
- Adds a new `DiscriminatedAttributes` data type for patches. Discriminated Attributes have an extra prefix at the start that need to be stripped. This is to support `PropertyTransitionsSerialize`.
- Adds `TweenInfo` to `rbx_types`. This is to support `PropertyTransitionsSerialize`.
- Updates documentation (`docs/attributes.md‎` and ‎`docs/patching-database.md`‎) with information about `TweenInfo` and `Add` patches.

## Example of `PropertyTransitionsSerialize` layout.
| Bytes | Meaning |
|-------|---------|
| `02 00` | Unknown header (not part of standard Attributes format) |
| `01 00 00 00` | Attribute count = 1 |
| `04 00 00 00 Size` | Key = "Size" |
| `24` | Type ID = 0x24 = TweenInfo |
| remaining bytes | TweenInfo fields (time, delay_time, repeat_count, easing_style, easing_direction, reverses) |

## `TweenInfo` layout
**Type ID `0x24`**
| Field Name      | Format | Value                                                                           |
|:----------------|:-------|:--------------------------------------------------------------------------------|
| Time            | `f32`  | The duration of the tween in seconds                                            |
| DelayTime       | `f32`  | The delay before the tween begins in seconds                                    |
| RepeatCount     | `i32`  | The number of times the tween repeats (`-1` for indefinite)                     |
| EasingStyle     | `u32`  | The easing style, corresponding to the `EasingStyle` enum                       |
| EasingDirection | `u32`  | The easing direction, corresponding to the `EasingDirection` enum               |
| Reverses        | `u8`   | Whether the tween reverses; `0x00` for `false`, any non-zero value for `true`   |

## Tests
I have an [open PR](https://github.com/rojo-rbx/rbx-test-files/pull/32) on `rojo-rbx/rbx-test-files` which adds a test for StyleSheet's/StyleRule's.


rbx_types/src/tween_info.rs
```diff
@@ -0,0 +1,172 @@
...
+            _ => return None,
+        })
+    }
+
+    /// Converts this vaue into a number identifier. Values are converted into
+    /// their respective value of the Roblox [`EasingStyle`][EasingStyle-rbx]
+    /// enum. As an example, `EasingStyle::Linear` is converted to `0`.
+    ///
+    /// [EasingStyle-rbx]: https://create.roblox.com/docs/reference/engine/enums/EasingStyle
+    pub fn as_u8(self) -> u8 {
```
krakow10, line 54: The standard function name for `fn(self) -> u8` is `to_u8`. `as_u8` would be for `fn(&self) -> u8`.  I recommend changing the function name to `to_u8` since the value is Copy.

rbx_types/src/tween_info.rs
```diff
@@ -0,0 +1,172 @@
...
+            _ => return None,
+        })
+    }
+
+    /// Converts this vaue into a number identifier. Values are converted into
+    /// their respective value of the Roblox [`EasingDirection`][EasingDirection-rbx]
+    /// enum. As an example, `EasingDirection::In` is converted to `0`.
+    ///
+    /// [EasingDirection-rbx]: https://create.roblox.com/docs/reference/engine/enums/EasingDirection
+    pub fn as_u8(self) -> u8 {
```
krakow10, line 108: to_u8

krakow10 (changes requested):

rbx_types/src/tween_info.rs
cameronpcampbell: I'd agree with you but I was trying to follow established naming conventions, particularly from the font.rs file: https://github.com/rojo-rbx/rbx-dom/blob/master/rbx_types/src/font.rs

krakow10: tragic

phoriah:
@cameronpcampbell How did you determine EasingStyle & EasingDirection to be i32? Shouldn't it be u32 since it's an Enum?

cameronpcampbell:
You're probably right, they should be u32. Small positive numbers have the same representation in i32 and u32. However, u32 seems more correct as it is in line with how enums are serialized elsewhere.

cameronpcampbell:
heres a luau script which parses a `PropertyTransitionsSerialize`:

```luau
local alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"

local function b64decode(input)
	input = input:gsub("[^%w+/=]", "")
	local result = {}
	local i = 1
	while i <= #input do
		local a = alphabet:find(input:sub(i, i)) - 1
		local b = alphabet:find(input:sub(i + 1, i + 1)) - 1
		local c_char = input:sub(i + 2, i + 2)
		local d_char = input:sub(i + 3, i + 3)
		local c = c_char == "=" and 0 or (alphabet:find(c_char) - 1)
		local d = d_char == "=" and 0 or (alphabet:find(d_char) - 1)

		table.insert(result, string.char(bit32.bor(bit32.lshift(a, 2), bit32.rshift(b, 4))))
		if c_char ~= "=" then
			table.insert(result, string.char(bit32.bor(bit32.lshift(bit32.band(b, 0xF), 4), bit32.rshift(c, 2))))
		end
		if d_char ~= "=" then
			table.insert(result, string.char(bit32.bor(bit32.lshift(bit32.band(c, 0x3), 6), d)))
		end
		i += 4
	end
	return table.concat(result)
end

local encoded = "AgABAAAABAAAAFNpemUkzcxMPgAAAAD/////CAAAAAEAAAAA"
local data = b64decode(encoded)
local buf = buffer.fromstring(data)

local off = 0

-- Discriminant header
local disc0 = buffer.readu8(buf, off); off += 1
local disc1 = buffer.readu8(buf, off); off += 1
print(`Discriminant: {string.format("%02x %02x", disc0, disc1)}`)

-- Attributes count
local count = buffer.readu32(buf, off); off += 4
print(`Attribute count: {count}`)

for j = 1, count do
	local nlen = buffer.readu32(buf, off); off += 4
	local name = buffer.readstring(buf, off, nlen); off += nlen
	local tid = buffer.readu8(buf, off); off += 1
	print(`\nKey: '{name}' (type_id=0x{string.format("%02x", tid)})`)

	local time = buffer.readf32(buf, off); off += 4
	local delay_time = buffer.readf32(buf, off); off += 4
	local repeat_count = buffer.readi32(buf, off); off += 4
	local easing_style = buffer.readu32(buf, off); off += 4
	local easing_direction = buffer.readu32(buf, off); off += 4
	local reverses = buffer.readu8(buf, off); off += 1

	print(`  time={time}`)
	print(`  delay_time={delay_time}`)
	print(`  repeat_count={repeat_count}`)
	print(`  easing_style={easing_style}`)
	print(`  easing_direction={easing_direction}`)
	print(`  reverses={reverses ~= 0}`)
```

phoriah:
> You're probably right, they should be u32. Small positive numbers have the same representation in i32 and u32. However, u32 seems more correct as it is in line with how enums are serialized elsewhere.

What about i32 & f32 on other properties of TweenInfo? How can you be sure?

cameronpcampbell:
- RepeatCount is a `i32` as it can be negative.
- Reading the `f32` components as if they were `f64` results in a buffer out of bounds error.

phoriah:
> * RepeatCount is a `i32` as it can be negative.
> 
>     * Reading the `f32` components as if they were `f64` results in a buffer out of bounds error.

I mean f32 could be u32, no?

cameronpcampbell:
reading the `f32` components as `u32` results in the incorrect numbers.

I defined `Time` as 0.2 but reading it as u32 gives `1045220557`.
I defined `DelayTime` as 0.5 but reading it as u32 gives `1056964608`

`AgABAAAABAAAAFNpemUkzcxMPgAAAD8AAAAACAAAAAEAAAAA`

kennethloeffler:
Please split the fix for #597 into a separate PR -- property transitions parsing does not depend on it, and we'll probably want to ship the fix before landing property transitions parsing.

cameronpcampbell:
Made a new PR which solely addresses the `PropertiesSerialize` issue here: https://github.com/rojo-rbx/rbx-dom/pull/602

PR #601, opened 2026-04-23
fix(rbx_types): rename approx_unit_or_zero to unit_or_zero + use exact equality

ZubeySweet:
## What

Change `approx_unit_or_zero` in `rbx_types/src/basic_types.rs` from a one-sided tolerance check to exact equality against `0.0`, `1.0`, and `-1.0`.

## Why

The existing tolerance check is one-sided:

```rust
} else if value.abs() - 1.0 <= f32::EPSILON {
```

evaluates true for any value less than `1.0` (including `0.5`, `0.0`, and small negatives) because `x - 1.0 <= EPSILON` is true whenever `x <= 1.0 + EPSILON`. The zero branch (`value.abs() <= f32::EPSILON`) has the symmetric problem — it admits small-but-nonzero values as "zero."

### Reproduction

Any `CFrame` whose rotation matrix has diagonals slightly off exact `1.0` — e.g. `0.9999987` from float-drift in imported bone transforms — gets passed through `Matrix3::to_basic_rotation_id`. With the one-sided check, those near-identity matrices get misclassified as canonical and snapped to exact `+/-1/0` tokens by the binary serializer (`rbx_binary/src/serializer/state.rs`, the basic-rotation-id token path). The result: CFrame entries silently alter on round-trip.

This surfaces loudest when round-tripping `.rbxl` place files with many near-identity rotations (mesh imports, character rigs). We've seen four-digit numbers of affected CFrames in a single place file.

### Fix

Exact equality is the strictest possible check. Only rotation matrices whose entries are bit-exact `0`, `1`, or `-1` qualify for the basic-rotation-id token path. Everything else gets the full 9-float encoding and the original bits survive byte-exact through the round trip.

### Trade-off

Models with many near-identity rotation matrices (mesh / bone transforms) produce slightly larger `.rbxl` files under this change, since those matrices now emit 9 floats instead of a 1-byte token. The worst-case size delta is O(matrices × 35 bytes) — negligible relative to typical textures / scripts / audio payloads in a place file, and the trade-off favors data fidelity.

## Test plan

- [x] `cargo test --all` passes (60 `rbx_binary` tests, 133 `rbx_xml` tests, plus everything else) — no existing test depended on the one-sided tolerance.
- [x] Verified against the `rojo-rbx/rbx-test-files` submodule — all model/place fixtures round-trip without the snap behavior now.
- [x] Confirmed the specific failure mode: a `CFrame` with `Vector3::new(0.9999987, 0.0, 0.0)` as a rotation column no longer snaps on serialize.

## Alternatives considered

- **Keep tolerance, make it symmetric** (`(value.abs() - 1.0).abs() <= f32::EPSILON`): fixes the one-sided branch but still admits drift. Our earlier iteration used this and it reduced but didn't eliminate the snap. Exact equality is the only check that's byte-exact.
- **Tighten tolerance to `f32::MIN_POSITIVE`**: same category as symmetric tolerance — still inexact. Not meaningfully better than exact equality for this purpose, and exact equality is simpler.

Happy to take feedback on preferred wording for the code comment or the commit message. First-time contributor here; let me know if there's anything else you'd like to see before merge.

### DIFF
--- a/rbx_types/src/basic_types.rs
+++ b/rbx_types/src/basic_types.rs
@@ -98,11 +98,33 @@ pub struct Vector3 {
     pub z: f32,
 }
 
-fn approx_unit_or_zero(value: f32) -> Option<i32> {
-    if value.abs() <= f32::EPSILON {
+fn unit_or_zero(value: f32) -> Option<i32> {
+    // Exact equality check against 0.0 / ±1.0.
+    //
+    // The previous one-sided tolerance `value.abs() - 1.0 <= f32::EPSILON`
+    // evaluated true for ANY value less than 1.0 (including 0.5, 0.0,
+    // small negatives) because `x - 1.0 <= EPSILON` is true whenever
+    // `x <= 1.0 + EPSILON`. That caused `Matrix3::to_basic_rotation_id`
+    // to misclassify near-identity rotation matrices (e.g. diagonals
+    // of 0.9999987 from float-drift in imported bone transforms) as
+    // canonical, snapping them to exact ±1/0 tokens during binary
+    // serialization and silently altering CFrame entries on round-trip.
+    //
+    // Exact equality is the strictest possible check: only rotation
+    // matrices whose entries are bit-exact 0 / ±1 qualify for the
+    // basic-rotation-id token path. Everything else gets the full
+    // 9-float encoding and the original bits survive byte-exact.
+    //
+    // Trade-off: slightly larger `.rbxl` files for models with many
+    // near-identity rotation matrices (mesh / bone transforms in
+    // particular). File-size delta is O(matrices × 35 bytes) in the
+    // worst case — negligible relative to textures / scripts / audio.
+    if value == 0.0 {
         Some(0)
-    } else if value.abs() - 1.0 <= f32::EPSILON {
-        Some(1.0f32.copysign(value) as i32)
+    } else if value == 1.0 {
+        Some(1)
+    } else if value == -1.0 {
+        Some(-1)
     } else {
         None
     }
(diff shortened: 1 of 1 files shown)


krakow10:
You should change the function name if it is no longer approximate.  What is Roblox's behaviour for near-exact rotation matrices?

ZubeySweet:
> You should change the function name if it is no longer approximate. What is Roblox's behaviour for near-exact rotation matrices?

Yeah your right, approx_ stops making sense once the tolerance is gone — maybe unit_or_zero (or exact_unit_or_zero if you want the "no tolerance" fact in the name), idk.

On Roblox's behavior: from round-tripping places with drift-prone rotations (diagonals like 0.9999987 from mesh/FBX imports), Studio doesn't snap them on save — both the .rbxl binary and the in-memory CFrame keep the 9 floats verbatim, and the engine handles them fine in physics/rendering. The basic-rotation-id token path seems reserved for bit-exact identity / 90° / 180° / 270° matrices that Studio produces itself (e.g. integer-degree Orientation values set via the Properties panel).

So the exact-equality check here matches what Studio actually does — the token is a size optimization for matrices already canonical, not a normalization.

ZubeySweet:
Pushed the rename (`approx_unit_or_zero` → `unit_or_zero`) as an amend to the same commit, and updated the title to match. No behavior change beyond what we already discussed. Three call sites in `Vector3::to_normal_id` updated accordingly; `cargo build --all` + `cargo test --all` still green.

kennethloeffler (approved):
Hello, thanks for the PR. 

Agreed that exact comparison is the right call here: I tested Roblox Studio's behavior for both binary and XML formats and that indeed seems to be what it's doing.

One additional drawback is that any consumers using rbx-dom's current behavior (e.g. Rojo's syncback command) may see different output as a result of these changes, but exact roundtripping and adherence to Roblox Studio behavior outweigh any annoyances caused there.

PR #602, opened 2026-04-27
Fix `PropertiesSerialize`

cameronpcampbell:
## Summary
- Fixes #597.

## Changes
- Introduces property `Add` patches. Aliases `StyleRule.PropertiesSerialize` to `Properties`.
- Updates documentation (`docs/patching-database.md`‎) with information about `Add` patches.


rbx_dom_lua/src/customProperties.lua
```diff
@@ -220,9 +220,7 @@ return {
 
 				local existing = instance:GetProperties()
 
-				for itemName, itemValue in pairs(value) do
-					instance:SetProperty(itemName, itemValue)
-				end
+				instance:SetProperties(value)
```
kennethloeffler, line 223: Should we care about the ways that StyleRule:SetProperties can fail? It fails and prints a warning in case of mismatched types, and silently fails in case of invalid property names; [docs here.](https://create.roblox.com/docs/reference/engine/classes/StyleRule#SetProperties)

Detecting and surfacing these failures is probably tricky and it's not immediately obvious to me how it would be useful to consumers (doesn't seem recoverable through any means), so this isn't a blocker, just something to think about

kennethloeffler (approved):
The documentation is accurate, the property add patch implementation looks correct, and the new tests are working well. Nice work, thanks.

To elaborate on the possible breakage I was worried about in the issue: consumers of rbx-dom have exposed PropertiesSerialize in previously released versions, and users may be relying on it. The two cases top of mind for me are:
* rbx-dom's serde format (e.g. PropertiesSerialize defined in Rojo project files)
* Lune's `roblox.Instance.get` and `roblox.Instance.set`, or anything that relies on their outputs

Rojo turns out not to be a problem, because rbx-dom's code has logic to resolve property aliases across the board, and Rojo uses rbx-dom to handle all these details.

However, Lune does not have special handling for property aliases, and property getters/setters is pretty much a direct look into the DOM. Pretty certain that any existing code that accesses PropertiesSerialize in Lune will start erroring, but it's more of a missing feature of Lune (or perhaps rbx-dom) and not a blocker for this fix.

PR #612, opened 2026-05-21
Add one-to-many property migrations, use them for UICorner.CornerRadius

kennethloeffler:
Recently, Roblox changed the serialization behavior of UICorner.CornerRadius. This property now migrates to the following properties:

* UICorner.BottomLeftRadius
* UICorner.BottomRightRadius
* UICorner.TopLeftRadius
* UICorner.TopRightRadius

rbx-dom's property migration system does not yet support this kind of one-to-many migration, where one property expands into multiple. This PR adds that capability, and uses it to add migrations for UICorner.CornerRadius to its new counterparts. 

Because it is no longer easily possible to get Roblox Studio to save out UICorner.CornerRadius, I added synthetic tests for rbx_binary and rbx_xml, creating doms containing the property to test against.

rbx_binary's serializer has grown some serious complexity since I last touched it, and getting this to work was pretty tricky. @krakow10, additional review and sanity checking in this area is greatly appreciated!

Also note that this is a breaking change to rbx_reflection, because it replaces `PropertyMigration.new_property_name` with `PropertyMigration.new_property_names`, makes it private, with an accessor for it.


rbx_binary/src/deserializer/state.rs
```diff
@@ -181,22 +181,29 @@ fn find_canonical_property<'de>(
...
+                        }
+                    }
                 }
+
                 Err(e) => {
-                    log::warn!("Failed to migrate property {old_property_name} to {new_property_name} because: {e}");
+                    log::warn!("Failed to migrate property {old_property_name} because: {e}");
                 }
             };
         }
```
krakow10:
```rust
        match migration.perform(&value) {
            Ok(new_value) => {
                for new_property_name in migration.new_property_names() {
                    if !instance.builder.has_property(new_property_name) {
                        log::trace!(
                            "Attempting to migrate property {old_property_name} to {new_property_name}"
                        );

                        instance
                            .builder
                            .add_property(new_property_name, new_value.clone());
                    }
                }
            }

            Err(e) => {
                log::warn!("Failed to migrate property {old_property_name} because: {e}");
            }
        };
```
I believe the `should_migrate` test is more expensive than the migration itself, using repeated linear scans with `has_property`.  Skip the pre-check and just do the cheap migration and run the scans once after.

rbx_xml/src/deserializer.rs
krakow10:
```rust
                    match migration.perform(&value) {
                        Ok(migrated_value) => {
                            for new_property_name in migration.new_property_names() {
                                let new_property_name = Ustr::from(new_property_name);
                                if !props.contains_key(&new_property_name) {
                                    log::trace!(
                                        "Attempting to migrate property {old_property_name} to {new_property_name}"
                                    );
                                    props.insert(new_property_name, migrated_value.clone());
                                }
                            }
                        }
                        Err(error) => {
                            return Err(reader.error(DecodeErrorKind::MigrationError(error)));
                        }
                    }
```
The cost of the check here compared to the migration is probably closer to a draw, even with the collect() into a Vec.  It's less code this way, and matches the rbx_binary code, so I'd suggest to take both or neither of my code suggestions for the rbx_xml and rbx_binary deserializer migrations.

rbx_binary/src/serializer/state.rs
```diff
@@ -368,95 +386,125 @@ impl<'db> SerializationInfo<'db> {
...
     }
 }
 
+type LogicalPropertyIndexList = SmallVec<[usize; 4]>;
+
+#[derive(Debug, Clone)]
+struct LogicalPropertyIndices {
+    indices: LogicalPropertyIndexList,
+    is_migration_source: bool,
+}
```
krakow10:
```rust
enum LogicalPropertyIndices {
    Migrates(Box<ArrayVec<usize, 4>>),
    Serializes(usize),
    DoesNotSerialize,
}
```
Curses to Roblox for shifting the requrements.  Here is my first thoughts about implementing this.  Using an enum makes everything explicit. Including a DoesNotSerialize variant means that TypeInfo's `properties_visited: UstrMap<Option<LogicalPropertyIndices>>,` becomes `properties_visited: UstrMap<LogicalPropertyIndices>,`.  Migration variant may or may not want to be Boxed, I'm not sure which one would be faster, but Boxing it reduces the size to 16 bytes, down from 56 bytes with smallvec, and matches the old size of `Option<usize>`.

rbx_binary/src/serializer/state.rs
```diff
@@ -166,6 +171,19 @@ impl<'dom> PropInfo<'dom> {
...
             .extend(core::iter::repeat_n(self.default_value, additional));
     }
+
+    /// Add this instance's value, unless this logical property already has one.
+    fn push_value_for_instance(&mut self, desired_len: usize, value: &'dom Variant) {
+        if self.values.len() > desired_len {
+            // This property for this instance has already been assigned a
+            // value, skip it
+            return;
+        }
```
krakow10, line 202: I don't like this.  Pushing a property value that silently fails feels wrong, and I was going to suggest that it be removed in favour of an assertion because I didn't understand what it was needed for.  Please write a comment that explains that this mechanism is used for migrations to multiple logical indices.

krakow10 (changes requested):
First round of review, hopefully this can be iterated upon.  Apparently I don't know how to operate github's suggest code changes thing.

krakow10:
#528 is also a breaking change in rbx_reflection, replacing all instances of Cow with &str.  The hope for that was to merge at the same time as other breaking changes since it provides a minor memory reduction and type simplification with the removal of Cow.

rbx_binary/src/deserializer/state.rs
kennethloeffler: Right, the pre-check is redundant here. We're already checking for preexisting properties inside the loop, no need to check beforehand.

Dekkonot (commented):
This requires a changelog entry if nothing else. It's technically a breaking change to `rbx_reflection` too, which sucks because that does cascade to the various dependencies if we properly follow semver.

I will defer to Krakow for reviewing the changes here. The implementation looks sound, but I'm a bit rusty (no pun intended) at Rust right now and I've missed a few PRs being merged so I'm not 100% confident in my ability to do anything other than go "yep, that'll work".

kennethloeffler:
It is possible to retain the new_property_name field with a custom Deserialize implementation, so we could do that and deprecate the field to avoid breakage, but if we're going to merge #528 anyway then we should probably just break it. I'm not aware of any consumers actually using the field either (no one wants to deal with migrations), so it shouldn't be too painful.

Dekkonot:
Yeah, I'm not afraid of breaking changes (though they are becoming increasingly common... thanks Roblox). It's a trivial change to move to the new version anyway in this case, so I'm not bothered.

rbx_binary/src/serializer/state.rs
kennethloeffler: Went with a boxed SmallVec, but we can use ArrayVec if there's good reason to

rbx_binary/src/serializer/state.rs
kennethloeffler: Added a comment within this method + comments near the call sites hopefully explaining why this is a noop

rbx_binary/src/serializer/state.rs
```diff
@@ -395,12 +401,28 @@ impl<'db> SerializationInfo<'db> {
...
+            LogicalProperty::Migration(indices) => indices.as_slice(),
+            _ => &[],
+        };
+
+        let serialized_index = match self {
+            LogicalProperty::Serializes(index) => Some(*index),
+            _ => None,
+        };
+
+        migration_indices.iter().copied().chain(serialized_index)
```
krakow10: This is an enum, it can't be both.  One match statement is sufficient.

rbx_binary/src/serializer/state.rs
krakow10: `logical_indices` looks like it may work better as the `LogicalProperty` type.

rbx_binary/src/serializer/state.rs
```diff
@@ -631,30 +647,40 @@ impl<'dom, 'db: 'dom, W: Write> SerializerState<'dom, 'db, W> {
...
+                    // deferred_migrations with its indices and value, and wait until
+                    // later to push its instance value, because we don't know yet if the
+                    // instance specifies any of the migration target properties, which
+                    // should take precedence.
+                    deferred_migrations.push((*indices.clone(), prop_value));
+                    continue;
+                }
+                _ => {
+                    // This property does not migrate, no need to fill deferred_migrations
+                    // and we may proceed normally
```
krakow10: Since this is the only branch that proceeds to below, you can extract the singular logical index and collapse the `for logical_index in logical_property.indices()` loops below.

krakow10 (changes requested):

rbx_binary/src/serializer/state.rs
krakow10: I would say it's a personal preference, something about the Unix philosophy.  ArrayVec is purpose-built to be a stack-allocated array with a fixed length, but SmallVec is more of a catch-all optimization trying to do everything at once.  ArrayVec may not even be the right choice, since we know that current migrations will be either 4 or 1 targets.

Something like this would be maximally explicit and wouldn't need a new dependency:
```rust
enum MigrationIndices {
    Four([usize; 4]),
    One(usize),
}
```

rbx_binary/src/serializer/state.rs
```diff
@@ -368,95 +391,131 @@ impl<'db> SerializationInfo<'db> {
...
+type MigrationIndices = SmallVec<[usize; 4]>;
+#[derive(Debug, Clone)]
+enum LogicalProperty {
+    Migration(Box<MigrationIndices>),
+    Serializes(usize),
+    DoesNotSerialize,
+}
+
+impl LogicalProperty {
+    fn indices(&self) -> impl Iterator<Item = usize> + '_ {
```
krakow10: This function is simplified if you return a slice.  I'm thinking the function can be removed entirely if refactoring `logical_indices` as a `LogicalProperty` works out.

rbx_binary/src/serializer/state.rs
kennethloeffler: Using a fixed-size container seems like over specification. I'd rather keep this more generalized since it is unknown whether more one-to-many migrations will be added or how many target properties they will have. We're not in control of it, and we'll end up churning code if there ever are.

rbx_binary/src/serializer/state.rs
kennethloeffler: Can you explain what you mean more? I'm not certain that it would be better, because if I'm not mistaken it means we'll have to destructure any time we want to add indices.

krakow10: I'm trying to look into this further, I was hoping it would work out simply.

Is it possible to push more than 4 indices into `logical_indices`?

i.e. Is this pushing 4 * 4 times?
```rust
        let mut logical_indices = SmallVec::new();
        for ser_info in ser_infos {
                // ...
                    for logical_index in existing_logical_property.indices() {
                        // ...
                        logical_indices.push(*logical_index);
                    }
                // ...
        }
```

rbx_binary/src/serializer/state.rs
```diff
@@ -368,95 +391,131 @@ impl<'db> SerializationInfo<'db> {
...
+            .iter()
+            .any(|ser_info| ser_info.migration.is_some());
+        let mut logical_indices = SmallVec::new();
+
+        for ser_info in ser_infos {
+            let canonical_name = ser_info.canonical_name;
+            if canonical_name != prop_name {
+                // Check if canonical name is already in properties_visited, return it
+                if let Some(existing_logical_property) =
+                    self.properties_visited.get(&canonical_name).cloned()
```
krakow10: `.cloned()` is not needed

rbx_reflection/src/migration.rs
```diff
@@ -35,9 +42,19 @@ pub enum MigrationOperation {
     FontToFontFace,
     BrickColorToColor,
     ContentIdToContent,
+    CornerRadiusToCornerRadii,
 }
 
 impl PropertyMigration {
+    pub fn new_property_names(&self) -> impl Iterator<Item = &str> {
```
krakow10: This is also simpler returning a slice.

rbx_binary/src/serializer/state.rs
```diff
@@ -583,30 +642,51 @@ impl<'dom, 'db: 'dom, W: Write> SerializerState<'dom, 'db, W> {
...
+            match &logical_property {
+                LogicalProperty::DoesNotSerialize => continue,
+                LogicalProperty::Migration(indices) => {
+                    // This property migrates to one or more other properties. Populate
+                    // deferred_migrations with its indices and value, and wait until
+                    // later to push its instance value, because we don't know yet if the
+                    // instance specifies any of the migration target properties, which
+                    // should take precedence.
+                    deferred_migrations.push((*indices.clone(), prop_value));
+                    continue;
```
krakow10: These two continues are not required here any more

rbx_binary/src/serializer/state.rs
```diff
@@ -583,30 +642,51 @@ impl<'dom, 'db: 'dom, W: Write> SerializerState<'dom, 'db, W> {
...
+
+            match &logical_property {
+                LogicalProperty::DoesNotSerialize => continue,
+                LogicalProperty::Migration(indices) => {
+                    // This property migrates to one or more other properties. Populate
+                    // deferred_migrations with its indices and value, and wait until
+                    // later to push its instance value, because we don't know yet if the
+                    // instance specifies any of the migration target properties, which
+                    // should take precedence.
+                    deferred_migrations.push((*indices.clone(), prop_value));
```
krakow10: The `&logical_property` borrow and `.clone()` on `indices` can be removed

krakow10 (changes requested):

rbx_binary/src/serializer/state.rs
kennethloeffler: Yeah, that seems wrong. For the UICorner.CornerRadius case, it looks like we will push into logical_indices 4 * 4 times (possible because SmallVec is growable), but we shouldn't be doing that. Probably need to rethink the interaction between SerializationInfo::new and TypeInfo::get_or_create_logical_property, seems like I mixed up some responsibilities! I'll be be taking a break right now, but I would greatly appreciate it if you continued thinking about this, and feel free to make a PR against this branch if you can come up with a solution

rbx_binary/src/serializer/state.rs
```diff
@@ -368,95 +391,129 @@ impl<'db> SerializationInfo<'db> {
...
+            _ => &[],
+        }
     }
 }
 
 impl<'dom, 'db: 'dom> TypeInfo<'dom, 'db> {
     /// Helper function for `TypeInfo::get_or_create_logical_property`.
     /// Push a new PropInfo into properties and add a list of aliases which point to it.
+    /// Return the index into self.properties where we inserted the new PropInfo.
     fn push_prop_info_with_names(
```
krakow10: This function can be downgraded from accepting an iterator to a singular name value

krakow10:
I have had a nagging feeling for a long time that there may be a more accurate data model for the rbx_reflection property serialization and migration datastructures, but I haven't come up with anything specific. That could be related to the complexity problems we're facing here. All the helper functions `PropertyDescriptors::new`, `find_property_descriptors`, `find_serialized_from_canonical`, suggest there may be a more direct database structure that requires less matching and scanning.

kennethloeffler:
Addressed comments, and did a fairly involved refactor to more clearly separate out the responsibilities of PropInfo bookkeeping/indirection, and resolution of properties to their serialized forms. This made it easier to do the fanout for one-to-many property migrations correctly, so we're in a bit better of a spot, but there's probably still room for improvement. 

I also decided to just use Vec for the target property lists, because boxing a SmallVec is somewhat counterproductive and I wasn't able to observe any performance difference.

Figuring out how messy input should be organized for binary serialization is part of the essential complexity here, and I'm not sure if we can effectively avoid it. Changing the structure of the db to specifically cater to rbx_binary's use case might just push the complexity there, and make it harder to use elsewhere.

rbx_binary/src/serializer/state.rs
```diff
@@ -289,174 +276,241 @@ impl<'dom, 'db> TypeInfos<'dom, 'db> {
...
+                    // to properties on the same class.
+                    // This avoids re-walking the superclasses.
+                    let mut targets = Vec::new();
+                    for new_property_name in prop_migration.new_property_names() {
+                        let new_descriptors = superclass_descriptor
+                            .properties
+                            .get(new_property_name.as_str())
+                            .and_then(|prop| PropertyDescriptors::new(superclass_descriptor, prop));
+
+                        targets.push(SerializedProperty::from_descriptors(new_descriptors?)?);
```
krakow10: This control flow looks strange, is the intent to bail on all property migrations if any property migration does not resolve?  I understand that doesn't happen in practice but perhaps the error-path logic should be hashed out to make the most sense according to the expected invariants, such as "multiple migrations should resolve completely or not at all".  Perhaps log an error message and bail?

krakow10 (commented):
Excellent work untangling the control flow and removing functions that have outlived their usefulness.  👍 to using a Vec; no smallvec dependency is preferred especially when there is no performance benefit.  I have one question on my first overview:

rbx_binary/src/serializer/state.rs
```diff
@@ -583,30 +634,52 @@ impl<'dom, 'db: 'dom, W: Write> SerializerState<'dom, 'db, W> {
...
-            // getting all properties of all instances again in `serialize_properties`.
-            logical_property.values.push(prop_value);
+                    let prop_info = &mut type_info.properties[prop_info_index];
+                    prop_info.push_value_for_instance(desired_len, prop_value);
+                }
+            };
+        }
+
+        for (prop_info_indices, prop_value) in deferred_migrations {
+            for prop_info_index in prop_info_indices.iter() {
```
krakow10: `.iter()` can be removed

rbx_binary/src/serializer/state.rs
kennethloeffler: It was just to get the types to work out, but it is probably a good idea to surface when a migration target property either does not exist in the db, or does not serialize. I'll `expect` on these because these kinds of problems mean we either messed up the database or messed up a migration.

rbx_binary/src/serializer/state.rs
```diff
@@ -289,174 +276,241 @@ impl<'dom, 'db> TypeInfos<'dom, 'db> {
...
+                        type_name,
+                        serialized_property,
+                        Some(migration),
+                    )?;
 
-                return Ok(prop_info);
+                    prop_info_indices.push(prop_info_index);
+                }
+
+                if prop_info_indices.is_empty() {
```
krakow10: `prop_info_indices.is_empty() == true` is impossible.  `SerializationResolution::Migration` is only created at one place in the code and `targets` is never empty since it is sourced from `PropertyMigration`, although that is not enforced at the type level.  This also means the `if matches!(resolved_property, DoesNotSerialize)` below can be removed.

You could also assert `PropertyMigration::Many(strings) => !strings.is_empty()` in rbx_reflector when constructing a `PropertyMigration` to enforce the invariant.  Actually, looking into it, I don't see any construction sites, how would you do that?

krakow10 (changes requested):

rbx_binary/src/serializer/state.rs
kennethloeffler: PropertyMigration is constructed via serde's Deserialize, and we can add some enforcement using a custom impl rather than deriving it

krakow10:
> Figuring out how messy input should be organized for binary serialization is part of the essential complexity here, and I'm not sure if we can effectively avoid it. Changing the structure of the db to specifically cater to rbx_binary's use case might just push the complexity there, and make it harder to use elsewhere.

That makes sense to me.  If the complexity is moved into the database, surely that is only as it pertains to migrations, which I doubt there would be many users for.  Moving the complexity to the database means it lives at compile time / initialization time rather than runtime, which is always preferred all else being equal.  Or at least that's what I would have been thinking about before your excellent refactor that seems to have tackled the problem rather nicely.

krakow10 (approved):
LGTM 👍

Dekkonot (approved):

PR #613, opened 2026-05-21
Support SSTR chunk version 1

max-au:
Version 1 of the SSTR chunk does not have unused MD5 field.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -149,16 +149,16 @@ The `SSTR` chunk has this layout:
 
 | Field Name          | Format                | Value                                        |
 |:--------------------|:----------------------|:---------------------------------------------|
-| Version             | `u32`                 | The version of the `SSTR` chunk (always `0`) |
+| Version             | `u32`                 | The version of the `SSTR` chunk (`0` or `1`) |
 | Shared String Count | `u32`                 | The number of SharedStrings in the chunk     |
 | Strings             | Array(Shared Strings) | The actual shared string entries             |
 
 A shared string entry looks like this:
 
-| Field Name    | Format              | Value                                                                   |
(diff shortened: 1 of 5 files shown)


kennethloeffler:
Hello, thanks for the PR. 

Is there an expected timeline for when this change will be rolled out? Any more information there would help us to know how urgent it is to get this released, and also serve as a reminder to create tests against it once Roblox starts writing the new format.  

Will the Roblox XML format change as well? Right now, the rbx_xml deserializer strictly requires the presence of an md5 attribute in shared string entries.

I'm also considering whether it's important for the serializer(s) to start writing the new format -- my initial impression is no, because Roblox should remain backwards compatible with the old format.

max-au:
Added XML bits. Despite the name, it's not MD5, and not BLAKE3 either.

The timeline is unknown, a few weeks at least, if not months. It's mainly a "heads up" note, to avoid potential breaking changes. 

Serializers should _not_ write a new format, at least not until Roblox does it for some time. I'll open a separate PR to implement that when the time comes.

Dekkonot:
Updating the [xml spec](https://github.com/rojo-rbx/rbx-dom/blob/master/docs/xml.md) would be preferred if possible. I would offer to do it myself but I'm not sure what the change in the XML deserializer actually amounts to.

FWIW we're aware that it's been blake2 for a long time, but nobody was using it for anything so we figured it didn't matter. Has that changed?

Dekkonot (approved):
Ideally we'd get some files to test this (are there flags we can flip in Studio to test this locally?) but this looks good to me for now.

max-au:
Soddy for the delay, please hold off with merging this. The change has not been released on the Roblox side yet, we're measuring if the saved file size justifies the hassle. Maybe this can be converted to draft.

PR #615, opened 2026-05-30
Use Entry to avoid navigating HashMap twice

krakow10:
Uses the entry api in `get_or_create_prop_info` and `resolve_visited_property` to remember which bucket is open when inserting into the hash maps.

### Performance
The performance impact is within margin of error due to the hot path being cache hit, and double-index only happening on cache miss.

### Notes
- `resolve_visited_property` return type is changed to a reference to match vacant_entry.insert return type, but it doesn't really matter and we could clone on the spot instead to keep the return type the same.
- `get_or_create_prop_info` is unassociated with `TypeInfo` due to needing multiple split borrows for `type_info.resolved_properties_by_visited_name` and `type_info.prop_info_indices_by_canonical_name` in `resolve_visited_property`.  This pushes the argument count over clippy's maximum of 8, so a closure is used to generate the default values ahead of calling `get_or_create_prop_info`.  Perhaps there's a cleaner solution, but I don't see it.
- Lint check is failing due to #614

### See also
- Another entry api change #526

### DIFF
--- a/rbx_binary/src/serializer/state.rs
+++ b/rbx_binary/src/serializer/state.rs
@@ -396,16 +396,15 @@ fn get_or_create_prop_info<'dom, 'db: 'dom>(
         hash_map::Entry::Vacant(vacant_entry) => vacant_entry,
     };
 
-    let Some(prop_type) = Type::from_rbx_type(serialized_ty) else {
-        // This is a known value type, but rbx_binary doesn't have a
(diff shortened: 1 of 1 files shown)


Dekkonot (approved):

Dekkonot:
Unsure if we even really want this change since it doesn't actually improve performance at all, but what's a cache hit now could be a cache miss in the future so... 🤷🏽

PR #617, opened 2026-06-01
Change rbx-reflector to use yaml_serde, use it to borrow patch data

kennethloeffler:
This PR changes rbx_reflector's YAML parser to [yaml_serde](https://crates.io/crates/yaml_serde) (a fork of serde_yaml maintained by the YAML org, as serde_yaml is no longer maintained), and uses it to borrow from patch file contents to avoid the code duplication from https://github.com/rojo-rbx/rbx-dom/pull/528#discussion_r3329461528. 

The strategy was first to introduce `PatchSources`, which gives us a way to store patch file contents until we're done processing patches, then introduce the new parser which borrows from these contents, remove the duplicate types, then add borrows and annotations, clone data if necessary, dereference, etc.

yaml_serde's default convention for enum parsing differed from serde_yaml's, so `default_value` and `data_type` fields needed `#[serde(default)]` and `#[serde(with = "yaml_serde::with::singleton_map_recursive")]` attributes for our existing patches to parse correctly.


krakow10 (approved):
serde_yaml -> yaml_serde, incredible.  One question, could it be taken further if Serialization was identical to PropertySerialization?  Is it worth breaking PropertySerialization if it can?

Dekkonot (dismissed):

rbx_reflector/src/patches.rs
```diff
@@ -285,20 +291,23 @@ impl PropertyChange {
...
 }
 
-impl PropertyAdd {
+impl<'a> PropertyAdd<'a> {
     fn kind<'db>(&'db self) -> PropertyKind<'db> {
         match (&self.alias_for, &self.serialization) {
-            (Some(alias), None) => PropertyKind::Alias {
-                alias_for: alias.as_str(),
-            },
+            (Some(alias), None) => PropertyKind::Alias { alias_for: alias },
```
krakow10: You can use alias_for as the variable name to shorten this
```
(Some(alias_for), None) => PropertyKind::Alias { alias_for },
```

Dekkonot (dismissed):

rbx_reflector/src/cli/generate.rs
krakow10, line 89: you can avoid repeatedly matching Option this way

```suggestion
        let patch_sources;
        let patches;
        if let Some(patches_path) = &self.patches {
            patch_sources = PatchSources::load(patches_path)?;
            patches = patch_sources.parse()?;
            patches.apply_pre_default(&mut database)?;
            apply_defaults(&mut database, &defaults_place_path)?;
            patches.apply_post_default(&mut database)?;
        } else {
            apply_defaults(&mut database, &defaults_place_path)?;
        }
```

krakow10:
This should definitely be merged before a new release since it removes the unsightly PropertyMigration constructor before it has time to become a breaking change.

kennethloeffler:
> One question, could it be taken further if Serialization was identical to PropertySerialization? Is it worth breaking PropertySerialization if it can?

Doesn't seem possible without breaking either the rbx_reflector patch or rbx_reflection database serde formats. These data are very similar, but are semantically different:`rbx_reflector::patches::Serialization` represents DB preprocess edits, while `rbx_reflection::database::PropertySerialization` represents some level of ground truth. So imo it's fine for them to remain separate.

krakow10 (approved):

PR #620, opened 2026-06-06
Update rand

krakow10:
Notes:
- range is changed to include i64::MAX, assuming its exclusion was a mistake
- rand is annoying to figure out where the damn functions are


Dekkonot (approved):
I can't remember if there was a reason for it to exclude `0`. I believe it was just a mistake.

PR #622, opened 2026-06-06
Update thiserror

krakow10:
Affected:
- rbx_binary
- rbx_reflection
- rbx_types

Is this breaking? Does thiserror expose some kind of custom trait implementation?

### DIFF
--- a/rbx_binary/Cargo.toml
+++ b/rbx_binary/Cargo.toml
@@ -25,7 +25,7 @@ rbx_reflection_database = { version = "2.0.2", path = "../rbx_reflection_databas
 ahash = "0.8.11"
 log = "0.4.17"
 lz4_flex = "0.11"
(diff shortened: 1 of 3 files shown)


Dekkonot:
> Is this breaking? Does thiserror expose some kind of custom trait implementation?

I have no idea! It might be breaking since we're returning the error type and if someone is currently using `thiserror` 1.X it will be a new type. Hesitantly, let's hold off on this for now. I'd rather we just swap to a proper error implementation if we're going to be making a breaking change anyway.

krakow10:
https://github.com/dtolnay/thiserror/blob/7214e0e8331d76afbea7173d8a14997512ac8713/src/private.rs

Looks like there is all sorts of traits exported.  That must be one of the reasons why they version bumped to 2.0.0.

PR #625, opened 2026-06-09
Move CompressionType from serializer.rs to chunk.rs

krakow10:
CompressionType is weirdly located as well.

### DIFF
--- a/rbx_binary/src/chunk.rs
+++ b/rbx_binary/src/chunk.rs
@@ -4,12 +4,21 @@ use std::{
     str,
 };
 
-use crate::{
-    core::{RbxReadExt, RbxWriteExt},
-    serializer::CompressionType,
-};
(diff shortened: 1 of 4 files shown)


Dekkonot:
This change is technically breaking, I think. The reason it's in `serializer.rs` rather than `chunk.rs` is because it's only used by the serializer. The deserializer doesn't care about it.

krakow10:
> This change is technically breaking

Is it breaking because the exposed type is moved into a different file?  I don't know the rules for what constitutes a breaking change very well.

> The reason it's in `serializer.rs` rather than `chunk.rs` is because it's only used by the serializer.

The most compelling reasoning for this change is that `CompressionType` is an argument to `ChunkBuilder::new`.  rbx_binary_core was a good litmus test for probing module boundaries :grin:

PR #626, opened 2026-06-09
Create rbx_binary_core crate

krakow10:
- Depends on #530
- Depends on #592
- Depends on #573
- Depends on #624
- Depends on #625

Provide rbx_binary core functionality to code outside of rbx_binary.

Useful for:
- #444 needs RbxWriteInterleaved & RbxReadInterleaved traits
- #92 My implementation is not complete, but here is the current branch https://github.com/krakow10/rbx-dom/compare/strong-base...strong4-binary


Dekkonot:
At the moment, I'm not interested in us doing this. If we were going to release a crate that had the reader and writer traits separately from `rbx_binary`, it wouldn't be `rbx_binary_core` (we'd give it a different name). Additionally, I don't think it's a good idea to focus on `rbx_dom_strong` at this time.

krakow10:
Closing due to above. I thought I was about to write the serializer and deserializer for strong types but that didn't happen. I will think of something else if I end up working on it again.

PR #628, opened 2026-06-11
Update database to 725 and add missing Content migrations

Dekkonot:
- Adds a new migration `Int64ToContent` because CharacterMesh properties are weirdly `Int64` values
- Adds patches for a _bunch_ of Content properties
- Updates the database to v725

This is a lot of migrations. I added a commit that re-generates the database after the patches were added so that you can double check them. Here's a list of the properties that should be migrated. It is not in the order they appear in the database unfortunately, but I also had to sort it for my own sake when writing the patches:

```
AdGui: FallbackImage -> FallbackImageContent
Animation: AnimationId -> AnimationContent
Beam: Texture -> TextureContent
FileMesh: MeshId -> MeshContent
FileMesh: TextureId -> TextureContent
ImageHandleAdornment: Image -> ImageContent
PackageLink: PackageContent SERIALIZES AS PackageContentSerialize
PackageLink: PackageId -> PackageContent
ParticleEmitter: Texture -> TextureContent
ScreenshotHud: CameraButtonIcon -> CameraButtonIconContent
Trail: Texture -> TextureContent
```

These ones are the weird Int64 to Content migrations:
```
CharacterMesh: BaseTextureId -> BaseTextureContent
CharacterMesh: MeshId -> MeshContent
```

There's a lot of these:
```
Sky: SunTextureId -> SunTextureContent
Sky: MoonTextureId -> MoonTextureContent
Sky: SkyboxFt -> SkyboxFrontContent
Sky: SkyboxDn -> SkyboxDownContent
Sky: SkyboxLf -> SkyboxLeftContent
Sky: SkyboxBk -> SkyboxBackContent
Sky: SkyboxUp -> SkyboxUpContent
Sky: SkyboxRt -> SkyboxRightContent
```


patches/package-link.yml
```diff
@@ -8,3 +8,15 @@ Change:
...
+    PackageContent:
+      Serialization:
+        Type: SerializesAs
+        As: PackageContentSerialize
+    PackageContentSerialize:
+      AliasFor: PackageContent
+    PackageId:
+      Serialization:
+        Type: Migrate
+        To: PackageContent
```
kennethloeffler, line 21: Is `PackageContent` the property name we want here? My understanding of the migration system is that target property get passed through as-is without considering aliases, so we may want `PackageContentSerialize` instead.

Dekkonot: If you look at the test snapshots it looks fine to me. I might be missing something, but it reads as `PackageContent` and writes as `PackageContentSerialize` which seems correct.

kennethloeffler: Taking a look, binary deserializer migrations are not aware of aliases, so `PackageId` gets migrated to directly to `PackageContent`. Binary serializer migrations do not directly resolve aliases, but they get resolved later during serialization, so any `PackageId` properties on the input dom are ultimately written as `PackageContentSerialize` (and of course, any `PackageContent` properties are also written as `PackageContentSerialize`). All that is well and good and correct.

I'm not so sure about XML. I think the serializer there does need to look up the serialized name for `PackageContent` to make sure it writes it as `PackageContentSerialiaze`, and it doesn't currently. I think (though I may be wrong) that this is the first property migration whose target property serializes under an alias, which explains why we haven't thought about this before. It looks like we don't have any snapshots for rbx-dom-serialized XML, so we can't see what's happening either.

Can we add some tests that snapshot rbx-dom XML output so we can nail down this behavior?

rbx_xml/src/tests/snapshots/rbx_xml__tests__package-link__encoded.snap
```diff
@@ -40,9 +40,9 @@ expression: encoded_text
         <string name="Name">PackageLink</string>
         <BinaryString name="AttributesSerialize"></BinaryString>
         <bool name="AutoUpdate">false</bool>
-        <ContentId name="PackageIdSerialize">
-          <url>rbxassetid://9111823334</url>
-        </ContentId>
+        <Content name="PackageContentSerialize">
```
kennethloeffler, line 43: Correct property name, neat!

kennethloeffler (approved):
Interesting, it looks that it does just work.

PR #634, opened 2026-07-01
Add decoding support for Tags stored in SharedString index

kennethloeffler:
This PR adds support to rbx_binary for decoding Tags blobs that are stored inside the SharedString index.

So far, this change has only been observed for binary files downloaded from the Roblox CDN. For example, from the following endpoints:
* `https://apis.roblox.com/asset-delivery-api/v1/assetId/{placeId}`
* `https://assetdelivery.roblox.com/v1/assetId/{placeId}`

I am not sure if rbx-dom should start encoding Tags this way. It doesn't seem necessary, so I'm erring on the side of not, but I'm interested in arguments for it.

### DIFF
--- a/rbx_xml/src/conversion.rs
+++ b/rbx_xml/src/conversion.rs
@@ -66,7 +66,7 @@ impl ConvertVariant for Variant {
             )),
             (Variant::SharedString(value), VariantType::Tags) => Ok(Cow::Owned(
                 Tags::decode(value.data())
-                    .map_err(|_| "Tags contain invalud UTF-8")?
+                    .map_err(|_| "Tags contain invalid UTF-8")?
                     .into(),
             )),
             (Variant::BinaryString(value), VariantType::Attributes) => {
--- a/rbx_xml/src/conversion.rs
+++ b/rbx_xml/src/conversion.rs
@@ -64,6 +64,11 @@ impl ConvertVariant for Variant {
                     .map_err(|_| "Tags contain invalid UTF-8")?
                     .into(),
             )),
+            (Variant::SharedString(value), VariantType::Tags) => Ok(Cow::Owned(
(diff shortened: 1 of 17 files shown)


Dekkonot:
This is technically something all String/BinaryString properties can support. We can talk to Roblox to see if they recommend we do this, but it's probably just an optimization for space or memory. If we don't do it, it should probably get written down in the [compatibility doc](https://github.com/rojo-rbx/rbx-dom/blob/master/docs/compatibility.md) for future reference though.

kennethloeffler:
Looks like Roblox Studio has now also started saving tags this way (sad)

Dekkonot:
Roblox Studio has begun producing files in Studio with this, so we need to also add support to rbx-xml.

kennethloeffler:
Feel free to make any edits here @Dekkonot, I'm going to be busy until tonight!

rbx_xml/src/conversion.rs
```diff
@@ -64,6 +64,11 @@ impl ConvertVariant for Variant {
                     .map_err(|_| "Tags contain invalid UTF-8")?
                     .into(),
             )),
+            (Variant::SharedString(value), VariantType::Tags) => Ok(Cow::Owned(
+                Tags::decode(value.data())
+                    .map_err(|_| "Tags contain invalud UTF-8")?
```
Corecii:
```suggestion
                    .map_err(|_| "Tags contain invalid UTF-8")?
```

Corecii (changes requested):

Corecii (approved):
Quick review. Looks fine!

Ran this on Adopt Me, didn't see any changes in tags (compared pre-this-change vs round-tripped rbxl) and the game launches fine.

PR #640, opened 2026-07-06
Support arbitrarily ordered CFrame coordinates in XML format

neiljohari:
# Problem

Still investigating which serializers created these places, but I've observed some cases where the ordering for CFrame coordinates isn't exactly what's expected.

I believe these shouldn't break the parser though, as these are named coordinates and shouldn't really be position dependent.

# Solution 

We still want to enforce the right number of tags and that every expected tag is there + only once. The suggested change accomplishes this by tracking which tags its seen and failing on truly incorrect inputs.

### DIFF
--- a/rbx_xml/CHANGELOG.md
+++ b/rbx_xml/CHANGELOG.md
@@ -1,6 +1,7 @@
 # rbx_xml Changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


Dekkonot:
If the files in question aren't private, we'd love to have an example for [`rbx-test-files`](https://github.com/rojo-rbx/rbx-test-files).

I imagine our parser is strict about this for all of the other types that serialize like this too (Vector3, UDim2, etc). Do you think making those more lax is a good idea as well?

PR #643, opened 2026-07-25
rbx_types: add Attributes.extend_from_reader

krakow10:
Useful for migrating InstanceHandle / Ref attributes into the existing Attributes data structure.


krakow10:
Nevermind it's not needed

PR #647, opened 2026-07-26
Implement AttributeReader & AttributeWriter

krakow10:
Preparing for Ref attributes, which needs to remap Ref values to i32 in the middle of reading and writing.  The plan is to implement the body of `Attributes::{from_reader, to_writer}` in rbx_binary and rbx_xml.  I believe this is non-breaking as written, since the only observable change is that it exposes new public types.

Notes:
- Replaces type_id.rs with attributes.rs
- New public types `AttributeReader`, `AttributeWriter`, `Attribute`


krakow10:
Maybe the error type should not be exposed? read_attribute and write_attribute can just return crate::Error

Edit:
I've gone ahead and done this.  AttributeError is once again private to the module, at the expense of adding loc and trouble generating errors in the ref attributes implementation.

PR #663, opened 2026-08-01
Add TweenInfo type and support TweenInfo attributes

krakow10:
TweenInfo uses attribute type_id 36 (0x24).

Given the Roblox Lua invocation:
```
TweenInfo.new(
	1.5, -- time: number?,
	Enum.EasingStyle.Sine, -- easingStyle: Enum.EasingStyle?,
	Enum.EasingDirection.InOut, -- easingDirection: Enum.EasingDirection?,
	4.5, -- repeatCount: number?,
	false, -- reverses: boolean?,
	6.5 -- delayTime: number?
)
```
The following bytes are serialized:
```
0x00, 0x00, 0xc0, 0x3f, // time: f32
0x00, 0x00, 0xd0, 0x40, // delay_time: f32
0x04, 0x00, 0x00, 0x00, // repeat_count: i32
0x01, 0x00, 0x00, 0x00, // easing_style: u32
0x02, 0x00, 0x00, 0x00, // easing_direction: u32
0x00, // reverses: u8
```

### DIFF
--- a/rbx_types/src/tween_info.rs
+++ b/rbx_types/src/tween_info.rs
@@ -1,0 +1,81 @@
+use crate::basic_types::Enum;
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct TweenInfo {
+    time: f32,
+    easing_style: Enum,
+    easing_direction: Enum,
+    repeat_count: i32,
+    reverses: bool,
+    delay_time: f32,
+}
+
+impl TweenInfo {
+    pub const fn new(
+        time: f32,
+        easing_style: Enum,
+        easing_direction: Enum,
+        repeat_count: i32,
+        reverses: bool,
+        delay_time: f32,
+    ) -> Self {
+        Self {
+            time,
+            easing_style,
+            easing_direction,
+            repeat_count,
+            reverses,
+            delay_time,
+        }
+    }
+
+    pub fn time(&self) -> f32 {
+        self.time
+    }
+
+    pub fn set_time(&mut self, value: f32) {
+        self.time = value;
+    }
+
+    pub fn easing_style(&self) -> Enum {
+        self.easing_style
+    }
+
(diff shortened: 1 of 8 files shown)


Dekkonot:
Please update the attributes spec with this info!

krakow10:
Oh hey #660 has competing TweenInfo attribute docs

rbx_types/src/tween_info.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    pub fn time(&self) -> f32 {
+        self.time
+    }
+
+    pub fn set_time(&mut self, value: f32) {
+        self.time = value;
+    }
```
kennethloeffler: This type of getter/setter pattern is very unusual for rbx_types -- I'm inclined to say that we should not have them. TweenInfo is a POD type that does not require any additional logic for getting or setting values, and we should probably prefer just pub fields for consistency's sake if nothing else. Consumers may use struct update syntax for changing a subset of the fields.

Since removing these will make the definition very short, it can then be moved into basic_types.rs.

krakow10: I was thinking the same thing a few days ago.  When I made this, I looked at CustomPhysicalProperties for reference and didn't look at anything else.

kennethloeffler: Probably not great for physical properties either, but we can't change it now without breaking stuff 😅

kennethloeffler:
We should add some files to the [rbx-test-files](https://github.com/rojo-rbx/rbx-test-files) corpus to test this implementation against. 

You'll use Roblox Studio to save out some rbxm and rbxmx files containing TweenInfo attributes, add them to the rbx-test-files repository (following the conventions there), PR that, we'll merge it, then update the rbx-test-files submodule here and add tests for them to rbx_binary/src/tests/models.rs and rbx_xml/src/tests/models.rs.

I don't think the Roblox Studio "add attribute" UI displays TweenInfo yet, so you'll have to use SetAttribute from the command bar to create instances with TweenInfo attributes.

krakow10:
I slopped come code to generate a file f43535048d681538ff057d52e9770f4c2a312380, loaded the file into roblox studio, and saved a rbxm and rbxmx from roblox studio.  I've added the test to the macro for rbx_binary and rbx_xml and generated the snapshots.

